PR #10, opened 2019-10-17
Hotfix/sticky header

evajacobs:
Fix z-index issue


app/styles/app.scss
```diff
@@ -219,12 +219,6 @@ tr.disabled {
   height: 95% !important;
 }
 
-.vl-modal {
```
Joachimzeelmaekers: Removing this results in select-options not being shown. Try opening the modals, for example 'Nieuwe agenda aanmaken'.
We're using Ember-Power-Select for the index here. Either lower this index or bump the z-index of the power-select.

Joachimzeelmaekers: Very good, seems to work perfectly for the creates.
Could you check the upload of the files for example: http://localhost:4200/agenda/5DA5ADD557D6EC0008000007/documenten

Try uploading a file and select a document-type. This one is still broken for me.

evajacobs: Hmm seems to work for me

evajacobs: Screenshot 2019-10-17 at 16 17 19
You mean this right?

Joachimzeelmaekers: Screenshot 2019-10-17 at 16 25 21
If I click it, I don't get any options..

evajacobs: And is there a z-index on this element: .ember-basic-dropdown-content--left.ember-basic-dropdown-content--below.ember-basic-dropdown--transitioned-in

evajacobs: I don't know how to change it because I can't find it in the code and I can't see the problem, could you give me some more direction?

Joachimzeelmaekers: I think I found a fix. I'll just merge the PR and commit the fix with it.
Weird that we can't reproduce it on both machines though.

evajacobs: Okay, thanks!

PR #35, opened 2019-11-08
KAS-1076 fix document linking between subcase and agendaitem & page s…

ValenberghsSven:
Added/changed adapters to have a page-size of 9999 for documentVersions and linkedDocumentVersions

Fixed the linking document-versions from subcases to multiple agendaitems on design agendas.


ValenberghsSven:
Not fixed

PR #38, opened 2019-11-08
Fix an issue which causes Firefox to only print a single page

Wolfr:
Signed-off-by: Wolfr <johan.ronsse@gmail.com>


Wolfr:
I branched from development. Going to close this PR. And branch off of master instead.

PR #144, opened 2020-01-16
Kas 581

woutersf:
- Add functionality that user can move subcase to different case.


app/components/utils/case-search/component.js
```diff
@@ -0,0 +1,126 @@
...
+        event.stopPropagation();
+      }
+      this.onSelect(caseItem);
+    },
+    selectSize(size) {
+      this.set('size', size);
+    },
+    filterCases(filter) {
+      this.set('page', 0);
+      const stoppingSearch = Boolean(!filter.searchText);
```
rafael-mindreau: Throws an exception, `filter` is `undefined` always. Because the location calling this action doesn't supply the actual filter value. So I'm going to plug this action out completely.

PR #149, opened 2020-01-17
KAS-1190: no sortable header for these templates

woutersf:
Disable sorting (Data-tables does not allow it)


woutersf:
> Liefst de comments weg, staan toch in de git-history.
ok

PR #151, opened 2020-01-17
KAS-1170: Extra fixes for new modals

rafael-mindreau:
image

image

* Did a proactive search of the common padding problem and found some more hidden candidates
* Fixed outline from webuniverse which we no longer need/want
* Fixed some issues regarding padding missing in the confirmation dialogs

### DIFF
--- a/app/components/subcases/assign-to-meeting-form/template.hbs
+++ b/app/components/subcases/assign-to-meeting-form/template.hbs
@@ -12,7 +12,7 @@
     {{/if}}
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


Wolfr:
If these are screenshots from the latest version, I would change the header to sans-serif font, here is the reference to the design.

https://www.figma.com/file/8tKG5CYiDDFPWJJ62kPYTj/%F0%9F%92%AF-Kaleidos-Master-Library?node-id=5%3A1013

PR #163, opened 2020-01-21
Kas 1192

RoelGo:
searchMessage is meant to tell the user to type to search
Even geduld gives a mixed signal

### DIFF
--- a/app/mixins/model-selector-mixin.js
+++ b/app/mixins/model-selector-mixin.js
@@ -14,7 +14,6 @@ export default Mixin.create({
 	filter: null,
 	loadingMessage: "Even geduld aub..",
 	noMatchesMessage: "Geen zoekresultaten gevonden",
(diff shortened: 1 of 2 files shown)


app/mixins/model-selector-mixin.js
```diff
@@ -14,14 +14,10 @@ export default Mixin.create({
 	filter: null,
 	loadingMessage: "Even geduld aub..",
 	noMatchesMessage: "Geen zoekresultaten gevonden",
-	searchMessage: "Even geduld aub..",
+	searchMessage: null,
```
Joachimzeelmaekers: Maybe we could create a useful search message here like: "Type je zoekterm".

Joachimzeelmaekers: @RoelGo what do you think?

PR #172, opened 2020-01-24
KAS-1232 Removing OC from the frontend

ValenberghsSven:


app/components/subcases/subcase-description/template.hbs
```diff
@@ -48,62 +44,60 @@
           {{t "number-session"}}
         </span>
         <span class="vl-description-data__value">
-          {{#if (not (await subcase.isOC))}}
+          {{#if (await subcase.case.latestMeeting.number)}}
+            {{await subcase.case.latestMeeting.number}}
+          {{else}}
```
rafael-mindreau: I can't seem to figure out where this comes from?

app/controllers/application.js
```diff
@@ -5,7 +5,6 @@ import { on } from '@ember/object/evented';
 import { inject } from '@ember/service';
 import isAuthenticatedMixin from 'fe-redpencil/mixins/is-authenticated-mixin';
 import { A } from '@ember/array';
-import moment from 'moment';
```
rafael-mindreau: I suppose moment was just simply unused here?

ValenberghsSven: Yes, unused import

app/components/subcases/subcase-description/template.hbs
ValenberghsSven: this span used to be 
```
        <span class="vl-description-data__value">
          {{#if (not (await subcase.isOC))}}
            {{t "no-number-yet"}}
          {{else}}
            {{await subcase.case.OCMeetingNumber}}
          {{/if}}
        </span>
```

-subcase.isOC relation has been removed
-OCMeetingNumber was a computed value that was equal to `subcase.case.latestMeeting.number`
The computed was removed but the number value (if present) should still be shown on the subcase description view

PR #185, opened 2020-02-05
KAS-1251: dubbele mededeling. await store.

woutersf:


ValenberghsSven:
No more PR into development, just merges

PR #188, opened 2020-02-05
KAS_1253: wait for the store to create record

woutersf:


ValenberghsSven:
No more PR into development, just merges

PR #193, opened 2020-02-07
KAS-507 Possible fix for themes not showing in KB

ValenberghsSven:


app/components/news-item/edit-item/component.js
```diff
@@ -26,9 +25,13 @@ export default Component.extend(DocumentsSelectorMixin, RdfaEditorMixin, {
...
 
   themes: computed(`agendaitem.themes`, {
-    get() {
+    async get() {
       const { agendaitem } = this;
-      if (agendaitem) return A([...agendaitem.get('themes').toArray()]);
+      if (agendaitem) {
+        return await agendaitem.get('themes').then((themes) => {
+          return themes.toArray();
+        })
```
rafael-mindreau: I think this can also be written as:
```javascript
  return await agendaitem.get('themes').then(themes => themes.toArray());
```

You can choose to do this, I just know some ESLint configs will complain on this. If you decide to leave it, the upcoming ESLint will nontheless complain about a missing semicolon 🤡

PR #201, opened 2020-02-12
KAS-1236: bugfix uitstel ongedaan maken

BenjaminDendas:


app/components/agenda/agendaitem/agendaitem-controls/component.js
```diff
@@ -137,15 +150,15 @@ export default Component.extend(isAuthenticatedMixin, {
...
-    return subcase;
+      const postponedPhases = await subcase.get("postponedPhases");
+      if (postponedPhases && postponedPhases.length) {
+        await Promise.all(
+          postponedPhases.map(phase => {
+            phase.destroyRecord();
+          })
+        );
+      }
+      return subcase;
```
Joachimzeelmaekers: Check format please.

BenjaminDendas: Seems ok according to the editor.

rafael-mindreau: But it doesn't seem OK to my **EYES** 👀 

Why is there an indentation of 4 spaces here?

app/components/agenda/agendaitem/agendaitem-controls/component.js
```diff
@@ -28,6 +33,14 @@ export default Component.extend(isAuthenticatedMixin, {
     });
   }),
 
+  async checkIfSubcaseExist(){
+      const subcase = await this.agendaitem.get('subcase');
+      if(!subcase) {
+        return;
+      }
+      return subcase;
```
rafael-mindreau: Same here. Your editor seems to go about 3 spaces for these async functions? That's strange!

rafael-mindreau: Can this be rewritten as the following?
```javascript
async checkIfSubcaseExist() {
  return await this.agendaitem.get('subcase');
},
```

rafael-mindreau: You don't have to, the contents just seemed like they don't really do much more than what can be reduced to a single line. If you like the more explicit version, we can keep it.

ValenberghsSven: This can be rewritten even better as:
```isPostPonable: computed("sessionService.agendas.@each","agendaitem.subcase", async function() {
    const subcase = await this.agendaitem.get('subcase');
    if(!subcase) {
      return false;
    }

  ....rest of the function
```

ValenberghsSven: Since there is no reuse of the
`checkIfSubcaseExist()`

Joachimzeelmaekers: checkIfSubcaseExist() is named well for me. It could be reusable and it's perfectly allowed to create a function for this if it's only used once.

app/components/agenda/agendaitem/agendaitem-controls/component.js
Joachimzeelmaekers: Fixed it with my editor.

PR #210, opened 2020-02-20
KAS-1282: toon indienende minister

woutersf:


app/components/web-components/vl-mandatee-link/template.hbs
```diff
@@ -9,7 +9,11 @@
   <div class="vlc-minister-item__title">
     <h6 class="vl-title vl-title--h6">
       <p>
+
         {{await titleToShow}}
+
+        {{#if  submitter }} ({{ t "submitter" }}){{/if}}
```
MikiDi: Een stijl-kleinigheid, maar bevorderlijk voor de leesbaarheid: [Inline form if's](https://api.emberjs.com/ember/3.8/classes/Ember.Templates.helpers/methods/if?anchor=if)

MikiDi: Concreet zou deze `{{if  submitter (t "submitter" )}}` worden.

Joachimzeelmaekers: Fixed the inline if and added another translation so it's not instantly added to the name of the mandatee. Also fixed linting errors because there were 3 new once introduced in other stories.

PR #220, opened 2020-02-26
Kas 1253 bis doc onzichtbaar

ValenberghsSven:


ValenberghsSven:
This branch has a wrong name and ticket number, this should be KAS-1251 part 2.

PR #231, opened 2020-02-28
KAS-1306: add monitoring to settings page

BenjaminDendas:


app/transforms/unix-to-datetime.js
```diff
@@ -0,0 +1,14 @@
+import DS from 'ember-data';
+import moment from 'moment';
+
+
+export default DS.Transform.extend({
+  deserialize(serialized) {
+    console.log('serializing: ', serialized);
```
rafael-mindreau: Not sure, but I don't think you wanted this in production?

app/transforms/unix-to-datetime.js
```diff
@@ -0,0 +1,14 @@
...
+      return moment.unix(serialized).format("HH:MM:SS DD-MM-YYYY");
```
rafael-mindreau: Strannge extra indent

translations/nl-be.json
```diff
@@ -519,5 +521,11 @@
...
   "error-delete-agenda": "Er ging iets mis tijdens het verwijderen van de agenda. Contacteer de administrator of probeer het opnieuw.",
   "no-themes-selected": "Geen thema's geselecteerd",
-  "no-themes-selected-message": "Er zijn nog geen thema's geselecteerd. Bent u zeker dat u het kort bestek wilt opslaan?"
+  "no-themes-selected-message": "Er zijn nog geen thema's geselecteerd. Bent u zeker dat u het kort bestek wilt opslaan?",
+  "monitoring-id": "ID",
+  "monitoring-type": "Type",
+  "monitoring-service": "Service",
+  "monitoring-timestamp": "Timestamp",
+  "monitoring-state": "Status",
+  "monitoring-error": "Error"
```
ValenberghsSven: Why put these things in the nl-be translation file but then keep the english words for timestamp and error.
Don't think the monitoring page needs a translation since all the errors we see will be in english anyway.

app/router.js
```diff
@@ -37,6 +37,8 @@ Router.map(function() {
     this.route('ministers', { path: '/ministers' });
     this.route('overview', { path: '/overzicht' });
     this.route('users', { path:"/gebruikers"});
+    this.route('errors', { path: "/errors"});
+    this.route('logs', { path: "/logs"});
   });
```
ValenberghsSven: This adds 2 routes to the settings overview, but both of them relate to monitoring.
Would it be better to add just the 1 route with subroutes ?

translations/nl-be.json
BenjaminDendas: I would keep them here just incase it might someday be used by other people that for example speak french. If we have a translation system, we better use it to its full extend throughout the application. Better have the translation in there then not having it there and have to redo it in the future.

BenjaminDendas:
> Model error.js and log.js are very similar with only 1 extra property.
> And the type property is basically "Error" or "Info"
> Could we just use 1 object and check the type instead of using 2 models ?

I've designed these models with keeping in mind that in the near future, those can be extended without much effort. When going to functional logging or tracing for example we could append extra information to the `log` domain without carrying load on the `error` domain that wouldn't be neccessary.

I would for sure keep this conceptual on hold until someone of us has talked with @nvdk to discuss what can be generalized between logging for lblod and Kaleidos.

MikiDi:
:information_source: Op 11-03-2019 is de `development`-branch overschreven in het kader van een wijziging in branching-strategie (nu: git-flow). Graag nakijken of je feature branch up to date is met de huidige state van `development` alvorens te mergen :pray:

app/router.js
BenjaminDendas: Fixed by removing the logs path as discussed with Niels

BenjaminDendas:
@rafael-mindreau Ik heb hier de changes op toegepast, dit kan nagekeken worden.

PR #251, opened 2020-03-06
KAS-1294: Printable agenda in subroute

rafael-mindreau:
**WHAT CHANGED**

* Added the printable agenda to a subroute
* This subroute is reached via the action menu
* The subroute has the option to return to the original agenda in the actions, or print the agenda

image

This makes it so that opening an agenda doesn't load all of the agenda items, which was the case in the original implementation of the print functionality.


app/components/item-document/template.hbs
```diff
@@ -1,32 +1,36 @@
...
-          </div>
-      {{await lastDocumentVersionName}}
-        </span>
-  {{/if}}
-</li>
-{{yield}}
\ No newline at end of file
+{{#if (and shortNotation (not (is-pending lastDocumentVersion)))}}
+  <p>({{await lastDocumentVersionName}})</p>
+{{else}}
```
ValenberghsSven: Not sure how this works, shortNotation is only true when coming from printable agenda
Why use the item-document component in this case?

app/components/newsletter/newsletter-header-overview/template.hbs
```diff
@@ -69,16 +69,6 @@
...
-              class="vl-button vl-button--tertiary vl-button--icon-before vl-button--narrow"
-              {{action "print"}}
-            >
-              <i
-                class="vl-button__icon vl-button__icon--before vl-vi vl-vi-printer-view"
-                aria-hidden="true"
-              ></i>
-              {{t "print"}}
-            </button>
             {{#if isEditor}}
```
ValenberghsSven: Is this #if statement empty now? Should this still be here or somewhere else ?

app/components/item-document/template.hbs
rafael-mindreau: I first tried to add this as a new component, but getting the latest document version was quite complex logic that would need to be copied, and extracting it into a common parent or mixin soon became overkill imho. So just implemented the short notation of documents as the name within parentheses.

ValenberghsSven: To be able to use the lastDocumentVersion and lastDocumentVersionName,
You have to extend your component with the MyDocumentVersions mixin

app/components/newsletter/newsletter-header-overview/template.hbs
ValenberghsSven: I tried printing a newsletter as minister and there is no button

rafael-mindreau: You mean there's no button outside of an action menu? That's OK. We moved this into the action menu on both ends. If it's missing from the action menu, that is not OK.

app/components/item-document/template.hbs
rafael-mindreau: I refactored this! I have a better feeling using a separate component for this as you sort of suggested!

ValenberghsSven: Looks good!

MikiDi:
:information_source: Op 11-03-2019 is de `development`-branch overschreven in het kader van een wijziging in branching-strategie (nu: git-flow). Graag nakijken of je feature branch up to date is met de huidige state van `development` alvorens te mergen :pray:

PR #256, opened 2020-03-12
KAS-1315: nowrap

woutersf:


app/styles/custom-utilities/_text.scss
```diff
@@ -9,3 +9,7 @@
 .vlc-u-text-decoration-underline {
   text-decoration: underline;
 }
+
+.vl-u-text--nowrap {
```
rafael-mindreau: For our own custom stuff we use the `vlc` namespace instead of `vl` 😬

rafael-mindreau (changes requested):
Almost! 😂

Wolfr:
And I see that the utility above that does not have `!important`.

PR #259, opened 2020-03-12
KAS-1313: hide tine on top of data-table

woutersf:


app/styles/govflanders/components/data-table/_actions.scss
```diff
@@ -1,4 +1,10 @@
 .vl-data-table {
+
+  // override the default data-table line.
+  .data-table__actions--top{
```
rafael-mindreau: Would add space before cruly brace
```css
.data-table__actions--top {
```

app/styles/govflanders/components/data-table/_actions.scss
```diff
@@ -1,4 +1,10 @@
...
+    border-bottom:0px solid transparent !important;
```
rafael-mindreau: This would be better. We no units when 0, and some spacing. But when unsetting a border it's better to just declare it as `none`.
```css
border-bottom: none !important;
```

rafael-mindreau (changes requested):
I feel sooo annoying requesting these changes 😅

Wolfr:
No tables need this so why is this even an override?

app/styles/govflanders/components/data-table/_actions.scss
woutersf: bedankt, zo leer ik ook weer iets bij

PR #266, opened 2020-03-17
KAS-1362 fix not showing the correct titles

ValenberghsSven:
We should show the title when the agendaitem is connected to a subcase or if the agendaitem is an announcement.
The only situation where we don't want to show the long title is for the approval (point 1)
So changed the if statement to check for the "isApproval" attribuut, which is only true for that type of agendaitem.


woutersf (approved):

woutersf:
PR niet meer nodig. Zit op alle branches

PR #267, opened 2020-03-19
KAS-1356: Remove jQuery from app

rafael-mindreau:
image

# jQuery removal

## ~~⚠️ Code review only~~
~~Hold off merging this. Currently requires a lot of testing, preferably local first. But you may comment on the code that you see here~~

> Update 24/03/2020: It's not to push this to `development` ✅ 

## 🔧 New util included in this box
I've created a light-weight wrapper around the native fetch API that mimics how jQuery requests are done. This way, only minimal changes were needed where the jQuery AJAX calls were performed. The util is called `ajax` (big wow), and will handle blobs and json with the typical error handling middleware you would expect with the fetch API. It will also accept classic jQuery `error` handlers passed along, but not the `success` handler that you would normally also use. I've documented it well, and would like you to point out if there are things unclear about it.

## 🌡 `HOT CHANGE`
If something's wrong with this refactor, immediately that means requests won't be getting through, or in the best of cases, feedback isn't shown from those requests. That means stuff is very broken very quickly. We have to make sure we test all cases this impacts, and there are some. Luckily it's easy to sum them up just from the changes you can see in this PR. For some of these you need the correct services running, I'm not sure if that's all doable on a local machine.

## Word on add-ons
We've split up the entire jQuery clean-up into two parts. This first part just covers **our code** and that is why the jquery-integration is still enabled in Ember. No worries. We fear that the second part will be a lot tougher though.


woutersf (approved):
Mooiste PR in tijden!

ValenberghsSven (approved):
Looks good.
I added the changes in cypress tests, some calls made with this ajax call are not picked up by cypress.

PR #272, opened 2020-03-25
KAS-1316: Add escape button functionality to modal

BenjaminDendas:
You can now hit the escape key in the modal to delete a user. The popup will dissapear and no actions are taken.

You can find the modal by going to:
- Instellingen
- Gebruikersbeheer
- hit the trash icon
- The modal appears
- Hit escape
- The modal should disapear


woutersf (approved):
Klein maar fijn
Gaan met die banaan

PR #287, opened 2020-04-02
Kas 1378 - system alerts

MikiDi:
Primarily aims to clean up and fix the logic for fetching system alerts through a new service. Also bugfixes in related components ([KAS-1376](https://kanselarij.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=KAS&view=planning&selectedIssue=KAS-1376), [KAS-1377](https://kanselarij.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=KAS&view=planning&selectedIssue=KAS-1377))


app/components/utils/manage-alerts/component.js
```diff
@@ -15,8 +21,8 @@ export default Component.extend({
 
   title: getCachedProperty('title'),
   message: getCachedProperty('message'),
-  beginDate: getCachedProperty('beginDate'),
-  endDate: getCachedProperty('endDate'),
+  beginDate: getCachedProperty('beginDate') || new Date(), // eslint-disable-line ember/avoid-leaking-state-in-ember-objects
+  endDate: getCachedProperty('endDate') || new Date(), // eslint-disable-line ember/avoid-leaking-state-in-ember-objects
```
rafael-mindreau: Can't we use moment here instead?

app/components/utils/system-alert/component.js
```diff
@@ -0,0 +1,14 @@
...
+    const t = this.args.type;
+    if (t === 'Waarschuwing') {
+      return 'warning';
+    } else if (t === 'Dringend') {
+      return 'error';
+    } else {
+      return null;
+    }
+  }
+}
```
rafael-mindreau, line 14: I cannot quite see where this new component is used from? What's the plan here?

app/components/web-components/vl-datepicker/component.js
```diff
@@ -21,7 +21,11 @@ export default Component.extend({
   selectedDate: computed('date', function () {
     const date = this.get('date');
     if (date) {
-      return this.formatter.formatDate(date.get('firstObject'));
+      if (date instanceof Date) {
+        return date;
+      } else {
```
rafael-mindreau: Oh boy, I can't say I like this bit. We're bridging a gap here that shouldn't have been. The date here - which is eventually passed down to formatters - is expected to be a `formatted date string`. I'd keep it like that. What went wrong with the original code that made you add this?

app/components/utils/manage-alerts/component.js
rafael-mindreau: A bit further up I found out more about this. I would use moment and format it to a date string. The Date had me really confused. I'm curious to find out the motivation/reason behind this.

app/services/system-alert.js
```diff
@@ -0,0 +1,68 @@
...
+import { action } from '@ember/object';
+
+export default class SystemAlertService extends Service {
+  @service store;
+
+  @tracked alerts = A([]);
+  updateInterval = 60 * 1000;
+
+  get currentAlerts () {
+    const now = new Date();
```
rafael-mindreau, line 15: So this is the root reason why there's Date used in the above components?

app/services/system-alert.js
```diff
@@ -0,0 +1,68 @@
...
+  confirmAlert (alert) {
+    alert.set('confirmed', true);
+  }
+
+  async updateAlerts() {
+    /*
+     * Below query fetches all alerts with end date greater then start of the day.
+     * This makes for only 1 unique request per day, which is good for request caching.
+     */
+    const today = moment().startOf('day').format();
```
rafael-mindreau, line 52: Here we use moment again? Hurray, but why is this mixed. I'm dying to find out! 😄

rafael-mindreau (changes requested):
Love the service, but I had some time-related questions >:)

app/components/web-components/vl-datepicker/component.js
MikiDi: I was very surprised to find out that our **date**picker component didn't support native Javascript **date** objects (flatpicker [does](https://shipshapecode.github.io/ember-flatpickr/docs/api/components/ember-flatpickr#date)).  
A "formatted date string" is a serialized format. Imo it is a bad practice to serialize stuff as long as you don't really need to (a native object representation is available). I tried to find a reference clearly explaining this, but couldn't find something nice at first glance ... there are many disadvantages to this, but the baseline is that it just makes things more complicated then need be. Passing this formatted date string is exactly the same as if you would pass `JSON.stringify(person)` instead of the `person`-object to an imaginary `person`-component. We don't do that either?
The choice for `Date` vs `Moment` is more subtle. In most of the app, the "interfacing-type" for sharing dates to components, ember-data, ... is a native Javascript Date object. That could be a `moment` two, but from a "implicit documentation" or portability perspective, a `Date` object is the most logical choice in Javascript-land. This doesn't mean you shouldn't use moment ... moment has a whole lot of functionality that the native date object doesn't have, but the information they represent is the same, so for interfacing you'd try to use what "the one on the other end" expects the most ... does all of this make some sense?

app/services/system-alert.js
MikiDi: As said above, entirely no reason why you couldn't use moment (getting the `startOf` done in vanilla JS would be much less readable).
Looking at it, the `format` doesn't look very consequent with what I'm preaching though :wink: ... that should be `toDate` :ok_hand:

app/components/utils/system-alert/component.js
MikiDi: It gets used [here](https://github.com/kanselarij-vlaanderen/kaleidos-frontend/pull/287/files#diff-f63965ba48df687541db46e0e1b2c2ecR5-R9). It's not much more than a wrapper to abstract away the type that the alert uses (map it to the type of the standard alert component)

MikiDi: If you see a cleaner way of doing this, I'm all ears :ear: :wink:

rafael-mindreau: I'm not sure, but you linked to this PR's changes? I just checked out your branch and globally searched for `system-alert` to find out where the component is used. The searched yielded no results. This is why I asked a question about it.

app/components/web-components/vl-datepicker/component.js
rafael-mindreau: It certainly makes sense, and in the sense of serialisation, I do get your point. But the benefits of using a generic standard for easier-to-read code overrule the negative side-effects of it imho. This won't slow us down or put us knee-deep in some kind of trouble or something. The native JS Date API also doesn't really play very nice. This is the big reason many developers have grown to really like libraries like moment.js (people like myself). Right now we're using the two together, which makes it hard to understand how code you've never seen works, especially if you're not sure what kind of object you're dealing with. I generally assumed it would be string timestamps until I saw the `|| new Date();`. The original code worked out fine, and this seems to patch up to support the off-chance of native Date objects. It really makes this code feel unnecessarily complex for some reason. I'm all in for readable straightforward code in the first place. I had to go back and forth for this to really sink in. Those are just my thoughts here.

app/components/utils/system-alert/component.js
MikiDi: mm ... the html-anchor doesn't seem to work ... click the link and search for `Utils::SystemAlert` :-)

app/components/web-components/vl-datepicker/component.js
woutersf: Personally I must say that I don't mind the idea of the date data being transferred in a very basic format (string). This has been proven to work in a lot of systems (I guess my cobol background).  It will also keep working when Date or Moment api's change (isn't it an iso standard?).
Even though I have a love hate relationship with moment (undefined returns a valid date), I am a fan of using a library consistently when available. 

If we are going to keep the different formats around: a suggestion:
Maybe it is a good idea to name your date variables so there is no doubt which type it is?  
This might be a bit extreme: https://en.wikipedia.org/wiki/Hungarian_notation but the idea can make it easier to see what kind of variables you are looking at? 
This could give you: nativeDateStartDate or momentStartDate. 
If you don't like this either we should eliminate the need for it.

app/components/utils/system-alert/component.js
rafael-mindreau: Ohhhh, of course, the new angle bracket syntax! OK no, this is fine! My bad!

app/services/system-alert.js
rafael-mindreau: Honestly I'd always use `moment` for operations on Date. Doing that vanilla is just a pain really. I think we can leave this out of the discussion as this shouldn't change imo. Except for `toDate`. The thing is that I really would not prefer a reliance on the Date object as a default.

woutersf:
Is da misschien een goeie boyscout rule? Daar waar mogelijk zelfde datetime library gebruiken (moment in ons geval) ?

MikiDi:
Ja, kan zeker ... dat wordt ook een aanzienlijke cleanup-story, maar is goed te doen.  

Het serializatie-verhaal is de laatste dagen wat aan 't malen geweest hier, maar weet nog steeds niet goed wat er van te denken :thinking:  ... dit gaat over de essentie van hoe je (componenten hun) interfaces definieert. Da's iets dat we veel doen (componenten maken), dus wel best belangrijk denk ik ... to be continued :twisted_rightwards_arrows: :wink:

woutersf:
Morgen sprintplanning/ @rafael-mindreau  @MikiDi  kunnen jullie hier tot een boy-scout-visie komen of plan ik hier iets apart voor in?

MikiDi:
Wanted to make changes requested, but noticed the changed `cached` function didn't behave as before, which re-introduced the bug where a system notification didn't receive default dates. I decided to refactor the component (and add related routes) to get rid of this `cached` construct and all the state management within the component.  
I now use moment as much as possible, and added logic for serializing it in the ember-data transforms. The deserialization-part is harder, because the whole of our codebase depends on it. This is why I still use vanilla dates to compare to in the service.  
Also reverted my changes to the datepicker.

woutersf (changes requested):
Ik mis nog een ding: een cypress test 😬

woutersf:
Verder vind ik dit scherm een beetje uit de lijn van de dingen die we doen precies:
Screenshot 2020-04-21 at 14 00 21

Een tabelleke 
| Titel | bericht | type | actions |
|-------|---------|------|---------|
|   Alles kapot    |   Alles is kapot     | warning |  edit/delete          | 

Had misschien meer aangevoeld als Kaleidos (en had je iets meer overzicht gegeven).

Het werkt.
Nu nog een test?

rafael-mindreau (approved):

woutersf (approved):
Nice

PR #289, opened 2020-04-02
KAS-1439: fileUpload was broken becasue of the a element

woutersf:
Updated the file_upload dependency.
File upload was breaking on the A element.


BenjaminDendas (approved):
Works like a charm :D

PR #299, opened 2020-04-07
KAS-1355: Mixin Refactor - Is Modified

rafael-mindreau:
# 🔧 Mixins Refactor: Is Modified

In this refactor we changed a **mixin** to become a **util**. The refactor itself was relatively straightforward.

###  🗑 What was completely removed
Originally the mixin had `changeFormallyOkPropertyIfNotSetOnTrue(subcase)`, which simply changed the `formallyOk` property. It was only used in two locations, so we removed that entirely! 🗑 

### 🔧 Motivation behind moving to a util
Upon inspection, the function that remained to be refactored: `updateModifiedProperty(model)` was not depending of any local states kept within the mixin. It was a **pure function**. So that seemed a good candidate for a util to me. It was used in enough places for it to still make sense, although, because of the small size of the function, I was tempted to remove it altogether.

### 💭 Some thoughts
@ValenberghsSven mentioned that the `updateModifiedProperty(model)` function was only used to set whether it was modified or not. But we have a `model-with-modifier` these days (we didn't have that back when the original mixin was written). So Sven wondered: `Do we still need this util?`. We're thinking we still need this for Yggdrasil to do its job properly, but we're not sure yet. So we're probably best clearing that out, and if that requires more time, make a spike or separate story to completely phase this one out. But for now this refactor is good to go, and we're one mixin less. A step forward!


woutersf (approved):
@rafael-mindreau  ik heb zo het gevoel (hoop) dat we die utils voor een paar mixins gaan kunnen gebruken. 
Dit is in ieder geval wel de goeie richting denk ik.

PR #304, opened 2020-04-10
Fix irrelevant search result

MikiDi:
This bug was probably a pagination/offsetting thing in the old ember-service for mu-search, which switched between the search api and the regular api.  
We didn't try to find this bug. Instead, we used the new mu-search util (already used for searching agenda-items), which uses the search api directly. This makes it more efficient two.


rafael-mindreau (approved):
Looks good to me!

PR #309, opened 2020-04-16
KAS-1171: Tooltip fixes

rafael-mindreau:
# Tooltip Fixes

image

### 📣What changed?

We just simply made text-only tooltips to look like the dark one's from Figma. (Like the one's you can see in the screenshot above). For this we also had to use the `attach-tooltip` which was `attach-popover` for some reason.

### 💡Good to know

In order to make the arrows work on the tooltip, you should do the following:

* Add `arrow="true"` to the attacher.
* Change the animation to something that supports arrows (like: `animation="shift"`). ⚠️

### DIFF
--- a/app/components/utils/text-template/template.hbs
+++ b/app/components/utils/text-template/template.hbs
@@ -15,14 +15,5 @@
   as |model|
   }}
     {{model.label}}
-    {{!-- Disabled for the time being, because this makes for horrible UX (decision Johan Ronsse KAS-1171)
-    {{#attach-popover
-      placement="top"
-      class="ember-attacher-popper ember-attacher-popper-wide"
-    }}
-      <p>
-        {{sanitize-html raw=true value=model.description}}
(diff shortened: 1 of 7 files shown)


app/components/utils/text-template/template.hbs
```diff
@@ -15,13 +15,14 @@
   as |model|
   }}
     {{model.label}}
+    {{!-- Disabled for the time being, because this makes for horrible UX (decision Johan Ronsse KAS-1171)
```
woutersf: Voor mij part mag dit weg.
We gaan dit nooit terugzetten op deze manier. (attach-popover > attach-tooltip).

woutersf (changes requested):
Testen gerunned?
Nieuwe code is proper. 
Oude boel mag er wat mij betreft uit.

woutersf (approved):

PR #312, opened 2020-04-17
KAS-1245: Fix Look/Feel of Side Navigation in Agenda List

rafael-mindreau:
# 🔧 Fix padding to the side

image

### What's fixed?

We used to have this case for agenda's with more than 100 items, that the number would stick to the title. I've given it a fixed width so that the alignment is never broken, and text never gets to stick into anything.


BenjaminDendas (approved):
Looking good to me

PR #322, opened 2020-04-27
KAS-1456 improve document-link component

ValenberghsSven:
**Initial issue:**
- Can't add new document-version to decision.

**After research:**
- Document version was added but the modal did not close, closing the modal would delete the newly added version.
- Using page refresh instead of closing modal would work to add the document version.

_Why did this happen_
The component used for adding a document-version is not generic enough.
We use this component to add document versions (documents) to the existing document (document-container) but with specifics for attaching versions to both 'agendaitem' and 'subcase'.
This component is also used for meeting, 'decision' and 'meetingRecord' models.

Agendaitem / subcase / meeting have '_documentVersions_'. 
We pass in this model and add new documents to the model
MeetingRecord and Decision have a '_signedDocument_', which is a document-container.
This signedDocument does not have '_documentVersions_'. but '_documents_' (same thing but correct name after document refactor).
We do not pass in this model but we pass in the signedDocument and add a version to the document-container
`hasMany('documentVersions').reload()` would fail in those cases since the relationship does not exist. This is the error in the ticket.

**Changes made**
_Document-link component_
- Made component more generic, moving most of the agendaitem or subcase specifics to _agenda-item-util.js_ or using the methods in this class.
- Extracted code from _saveDocuments_ to new functions for better readability
- Changed the variable and method names used (documentVersion => document) and (document => documentContainer) cfr. document refactor

_Document-link template_
- Spacing

_agendaitem-minutes component/template_
- Applied same fix as for decision in KAS-1440
 
_agenda-item-util.js_
 - Moved function from document-link into this util class


woutersf (approved):
@ValenberghsSven  Bekijk je nog de merge conflicts met development?
Lokaal getest 👍

BenjaminDendas (approved):
Getest en lijkt me in orde

PR #325, opened 2020-04-27
KAS-1174: Redesign Login

BenjaminDendas:
# KAS-1174: Redesign login

## ✏️  What has changed:
De login is aangepast volgens de mockup https://www.figma.com/file/wtICRLhGTAMN2sxl50Eixf/%F0%9F%92%AFKaleidos-Mockups-Q2-2020?node-id=31%3A3678

## 🖼 Design in de applicatie

image

## 🕵🏻‍♂️ Review
Dit is kleine story, waarvoor er volgens mij niet echt 2 reviewers nodig zijn.. het lijkt echter wel aangewezen dat @rafael-mindreau de CSS bekijkt of er eventuele fouten gemaakt zijn.


app/styles/custom-components/_vlc-login.scss
```diff
@@ -0,0 +1,12 @@
+.vlc-login {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+}
```
rafael-mindreau: This only makes it so that it is a tiny bit lower (header height to be exact)|, but I would just get rid of this. Remember, you can see these styles appear in Figma quite frequently, and you should disregard them in most of the cases.

app/styles/custom-components/_vlc-login.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.vlc-login-content {
+  margin: auto auto;
```
rafael-mindreau: This can be just `margin: auto`. `auto auto`, or any double-value assignment to these spacing properties allow you to set different spacing for the x and y spacings, but here you make them both `auto`, which equals to `margin: auto`.

app/styles/custom-components/_vlc-login.scss
```diff
@@ -0,0 +1,12 @@
...
+  max-width: 630px
```
rafael-mindreau: Oops, those are pixels! 🤡  Better use `63rem` instead.

rafael-mindreau (changes requested):
Only very minor changes requested

app/styles/custom-components/_vlc-login.scss
BenjaminDendas: Ok, noted

app/styles/custom-components/_vlc-login.scss
BenjaminDendas: I saw, it indeed only changed a little bit.. I need to check in which css class it is centering the component now though as this is unclear to me, as this snippet would center the component.

rafael-mindreau (approved):
👍

PR #333, opened 2020-05-06
KAS-1506: Icon-font Generator

rafael-mindreau:
# Icon Font Automation 🤩 

image

### 🆎 What the font?
We've chosen to go with icon-fonts to get the icons right in the app. We will thus be transitioning **from the flanders iconset** to our own **kaleidos iconset**. Which is kind of cool! Because this means we get to just throw in the icons we see in designs, without having to worry about them being in the set we were using. The automation of the font generation is what this PR is all about.

### ⛑️ Testing this
From now on, all you'll have to do is:

```
make icon-font
```

That should generate:

* The fonts in the `public/fonts/` folder
* The appropriate CSS in the `app/styles/` folder
* The overview page as an Ember `.hbs` template

There's a new route in the application: `http://localhost:4200/icons`. When you visit it, you'll see all the currently loaded icons.

### ➕ Adding new icons
Simply download the SVG files, or copy the markup, and add them to the folder: `icon-font-svg-files`. When you generate the font again, it will now include the one's you've added. The name you give to the file, will also be used in the app including the prefix.

### 📝 Note on prefixing
We've decided to go with `ki` prefix. This means that all icon classes are prepended with this. If you add an `add.svg` plus-icon to the folder, it will generate to become `ki-add` in the app.


woutersf (approved):
Ik kan hier echt niets op aanmerken.
Dit werkt proper

BenjaminDendas (approved):
Looking good!

PR #335, opened 2020-05-07
Bugfix/kas 1492 self concurrency

ValenberghsSven:
The associated ticket had some vague issues about self-concurrency throughout the app.

What this PR changes:
### - components/agenda/agenda-list/component.js:
- Fixed self-concurrency when reassigning priorities of agendaitems (moving the agendaitems in the agenda by drag and drop) by awaiting the async save of the model
### - components/agenda/agenda-list/list-item/component.js:
- Fixed self-concurrency when setting formal ok after making edits by awaiting the async save of the model
### components/agenda/agendaitem/agendaitem-decision-edit/component.js
- Fixed self-concurrency when editing a decision by removing the setting of _modified_ on agendaitem
Note: there is work to be done in this component since the saveChanges has a lot of dead code. To be picked up in a different ticket.
### components/agenda/agendaitem/agendaitem-decision/component.js
-  Fixed self-concurrency when deleting a decision by removing the setting of _modified_ on agendaitem
- Fixed the verifying modal not closing when deleting a decision
- Fixed that the deleted decision was still visible until after a refresh
### components/agenda/agendaitem/agendaitem-decision/template.hbs 
- Added test data tags
### components/agendaitem-press-edit/component.js:
- Fixed self-concurrency when editing a press-item by removing the setting of _modified_ on agendaitem
### controllers/cases/overview.js
- Fixed the self-concurrency when archiving or unarchiving a case by awaiting the async save on subcase
note: un/archiving cases is not working as intended because of the search results from elasticsearch
Archiving a case should remove it from the normal search, but this is not the current situation.
also note: The action to unarchive the case, does not seem possible at the moment because of the above.
### models/agendaitem.js
- Added an @each to the computed to pick up changes in decisions

CYPRESS:
###  cypress/integration/unit/decision.spec.js 
- Updated test to include deleting the decision
- Use more selectors
 ### cypress/selectors/agenda.selectors.js
- Added selectors
 ### cypress/selectors/modal.selectors.js
- Removed unused/old selectors
 ### cypress/support/commands/document-commands.js
- Fixed incorrect selector


app/components/agenda/agendaitem/agendaitem-decision/component.js
```diff
@@ -52,15 +52,18 @@ export default Component.extend(isAuthenticatedMixin, {
...
       if (this.isTableRow) {
-        const subcase = await this.agendaitem.get('subcase');
-        (await subcase.get('decisions')).reload();
         agendaitemToUpdate = await this.agendaitem.content;
       } else {
         agendaitemToUpdate = await this.get('agendaitem');
       }
-      agendaitemToUpdate.set('modified', moment().utc().toDate());
       await agendaitemToUpdate.save();
+      if(!this.isDestroyed) {
```
rafael-mindreau: Can I be annoying again? Small change request: `if (` 😬

rafael-mindreau (changes requested):
Couldn't find a lot, except for one little request which makes me a complete code style nazi 😬. But hey 😉

app/components/agenda/agendaitem/agendaitem-decision/component.js
ValenberghsSven: Fixed it

rafael-mindreau (approved):

BenjaminDendas (approved):
Looks fine to me

PR #349, opened 2020-05-15
KAS-1395: Mixin Refactor - Light Table

rafael-mindreau:
# ⚙️ Mixin Refactor: Light-table

✅ Ran some relevant agenda specs, and they seem to pass

## What was changed?

* Just moved the mixin code to the only component using it 🤷 
* Changed the component over to Octane syntax
* In doing so I found it didn't work any more
* I re-wrote part of the code to make it work again with computed properties watching the model
* I added some documentation

## How to test?
Contrary to what you might think, this is not the table on agenda detail. This is everything in the newsletter route. Go visit KB and try things out. I ran all newsletter tests as well.


woutersf (approved):
Makes sense

PR #353, opened 2020-05-18
Bugfix/kas 1551 lock agenda

ValenberghsSven:
Changelog for this PR:

- Changed the logic to determine if the meeting can be locked:
  We should be able to lock the agenda if there is at least 1 approved agenda.
  We should not be able to lock the agenda if there is only a design agenda.

- Added cypress tests


woutersf (approved):
Seht ja gut aus

BenjaminDendas (approved):

PR #354, opened 2020-05-18
KAS-1398: Mixin Refactor - Document Selector

rafael-mindreau:
# ⚙️ Mixin Refactor - Document Selector

✅ Ran the decision spec successfully

## What changed
We brought the code from the mixin under the two components that made use of it. I also refactored some dead code annoted by @ValenberghsSven that we had better dealt with sooner or later.


woutersf (approved):
Looks good to me.

PR #355, opened 2020-05-20
Kas 1373: Verbetering look document overzicht

brechtPhilips:
**Implementation of panels in agenda overview**

Missing:
-new document icons( icon generator linux)


rafael-mindreau:
There's really something off with the borders:
image

This is the design:
image

The panel should only have a bottom border.

rafael-mindreau:
The panels have a bottom margin of `0.4rem` but it should be `1.5rem` according to the designs.

image

image

app/components/agenda/agenda-list/template.hbs
```diff
@@ -86,7 +86,9 @@
           <Agenda::AgendaList::ListItem
             @agendaitem={{agendaitem}}
             @isEditingOverview={{isEditingOverview}}
-            @selectAgendaItem={{action "selectAgendaItemAction"}} />
+            @selectAgendaItem={{action "selectAgendaItemAction"}}
+            @useNewIcons={{true}}
```
rafael-mindreau: Wait, what did this do exactly? Is it for newly added agendaitems or just new icons?

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,52 @@ $i: 0;
...
   color: $napier-green;
 }
+
+.vlc-agenda-items__body {
+  align-items: center;
+  padding: 1.2rem 1.8rem;
+
+  .vlc-agenda-items__body-padding {
+    padding-bottom: 1.5rem;
+  }
```
rafael-mindreau: You're using BEM to specify a child. We do that to prevent nesting. Nesting will add higher specificity scores than desired. To fix this, you can place this block of code alongside the other selectors instead of nesting it.

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,52 @@ $i: 0;
...
+}
+
+.vlc-agenda-items__icons {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  .vlc-agenda-items__icons::before {
+    color: #FFF;
+  }
```
rafael-mindreau: Does this even work? Shouldn't you use `&` when using the nested approach? You would write this if you put this block next to the one it's in. Here it looks like you're selecting a `vlc-agenda-items__icons` element within another `vlc-agenda-items__icon` container, and selecting its `::before`. 🤔

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,52 @@ $i: 0;
...
+}
+
+.vlc-agenda-items__confidentialityButton span {
```
rafael-mindreau: Preferably use `kebab-casing` instead of `camelCasing` for class names

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,52 @@ $i: 0;
...
+  display: flex;
+  flex-direction: row;
+
+  .vlc-agenda-items__body--number {
+    justify-content: flex-start;
+  }
+
+  .vlc-agenda-items__body--text {
+    width: 100%;
+  }
```
rafael-mindreau: These modifiers should not be nested in another modifier.

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,52 @@ $i: 0;
...
+}
+
+.vlc-agenda-items__confidentialityButton span {
+  color: #FFFFFF;
+  margin-right: 0.1rem;
+  align-items: center;
+}
+
+.vlc-agenda-items__body--flex {
```
rafael-mindreau: It's strange to use a modifier for the flex. A modifier should be used as a modification of something that already exists on its own. Like a button, can have a modifier `--bigger` or `--red`. But using flex feels more like a utility here, but I could be wrong?

app/styles/custom-components/_vlc-document.scss
```diff
@@ -0,0 +1,3 @@
+.vlc-document__icon::before {
+  color: green;
```
rafael-mindreau: I think the Figma makes this green specifically `#8BAE00`.

rafael-mindreau (changes requested):
Small request 😅

app/styles/custom-components/_vlc-agenda-items.scss
```diff
@@ -199,3 +192,51 @@ $i: 0;
   opacity: 1;
   color: $napier-green;
 }
+
+.vlc-agenda-items__body {
+  align-items: center;
+  padding: 1.2rem 1.8rem;
```
rafael-mindreau: This is best to become `2rem` instead. Here's the design:
image

Here's the current app:
image

rafael-mindreau (approved):

app/styles/custom-components/_vlc-document.scss
BenjaminDendas: Dit lijkt nog niet in orde?

BenjaminDendas (changes requested):
Just one little thing left to fix

BenjaminDendas (approved):
In orde nu :D

PR #356, opened 2020-05-20
KAS-1553: Fix Access Level

rafael-mindreau:
# 🐞 Bugfix: Access Level Saving Inconsistencies

image

## What changed?

* Added some protection against saving an empty state `null`
* Disabled useless search in the dropdown (like designs)
* Added a protection against trying to change the loading state when `null`, by moving to a normal class based property for the state


woutersf (approved):
Ziet er ok uit.
Bollen de testen hiermee?

PR #359, opened 2020-05-22
KAS-1543: Fix icon font issues

rafael-mindreau:
# 🐞 Fix icon-font script to support Linux

## What changed, and why?

The `sed` commands behave differently per OS, and we would like to keep the script instead of going full node. Changed the script interpreter, as well as added some conditional blocks per OS.


woutersf (approved):
Nice

PR #366, opened 2020-05-27
Feature/kas 1423 case refactor

ValenberghsSven:
- Changed the type of a case
- Changed the relation between case and subcase
- Removed several unused relationships

### In detail: 
1. components/cases/forms/metadata-form:
This was used for the short title when creating a new case.
A lot of code in the component was unused.
- Deleted metadata/form (only 1 occurence)
- Moved the relevant code and html elements to case/new-case

2. components/cases/new-case: 
- Cleanup the creation of case to have only the current model properties
- Cleanup unused actions

3. components/cases/new-subcase:
- Due to a cleanup in vl-radio-button-list, we have to query the data in this component (query used to be in the radio-button-list)

4. components/subcases/subcase-description:
- Changed the selection for case-type to a radio button (Nota of mededeling) instead of a dropdown list (model-selector)
- The above change requires us to query in the current component (query used to be in model-selector)
- Removed `await caze.getNameForNextSubcase(subcase, type);` This function always returned `type.get('label');` because the method only has 1 parameter and we pass 2. 
- Removed the setting of "remarkType". This is a computed property so setting this value is lost when refreshing.

5. components/web-components/vl-radio-button-list:
- Cleanup the component; removed specific query, removed logic to select a value.
- Changed the template to use a parameter for selectedModel, default selection when this is empty is the first option.

6. models/case-type:
- Removed unused relation (Case-type is set on subcase, but instead of making a relation to the type, we set "isRemark" boolean to true ('Mededeling') or false ('Nota') )

7. models/case:
- Removed unused relations and methods
- Removed `getNameForNextSubcase`, this method was never correctly called, so we always ended up in the first if block.

8. models/meeting:
- Removed unused relation

9. models/subcase:
- Removed unused method

10. utils/config.js
- Removed the unused entries as a result of deleting  `getNameForNextSubcase`


BenjaminDendas (approved):
Ziet er goed uit voor mij. Heb ook de testen gerunned die de subcases gebruiken en deze slaagden.
Important: er is een branch op Kaleidos-project hiervoor https://github.com/kanselarij-vlaanderen/kaleidos-project/pull/90/files

app/components/subcases/subcase-description/component.js
```diff
@@ -53,14 +62,13 @@ export default Component.extend({
     async selectType(type) {
       const subcase = this.get('subcase');
       const caze = await subcase.get('case');
-      const subcaseName = await caze.getNameForNextSubcase(subcase, type);
```
woutersf, line 56: caze wordt hier niet meer gebruikt, dus mag mss ook weg?

woutersf (approved):
Looks ok
Maybe interesting to take this one along in the next release?

app/components/subcases/subcase-description/component.js
BenjaminDendas: solved.

PR #374, opened 2020-05-29
Remove wide buttons. Every button is narrow anyway.

Wolfr:
We discussed this change on Thursday.

We do not use the wide buttons as they are in the original WebUniversum, so these can be removed. This makes the code simpler.

This PR removes the `vl-button--narrow` class on every instance where it is used (including in tests, where I would really recommend not ever testing based on the existence of a CSS modifier) and makes buttons default to the narrow padding instead.


woutersf (approved):

BenjaminDendas (approved):
In orde, testen gerunned en werken ook nog.

BenjaminDendas:
Heb development nog in deze branch getrokken en het merge conflict en linter issues opgelost (div was niet gesloten)

PR #399, opened 2020-06-11
Fixes for tests

woutersf:


woutersf:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/Development/16/console

PR #413, opened 2020-06-19
KAS-1479: Production Environment Pill

rafael-mindreau:
image

# 👀  See the prod environment

## 🔧  What has been added

I've added an option for you to see the flashing prod pill as a developer, **only if you have been added to the whitelist**. So this is a very individual feature. If you want to be warned as well, I can add you to the whitelist. Best comment that on this PR.

It has been written in such a way that it's impossible to see the pill if you're not part of the list, even if there's an error or something similar.

## 📝  Sign me up

If you want to be added to the whitelist, you need to give me your ID. You can find this ID in the Ember Data inspector. There's only one user object, and the ID there is a uuid. I need to add this uuid in order for this to work.


woutersf:
TESTEN ✅

woutersf (approved):

BenjaminDendas (approved):
Ziet er in orde uit voor mij!

woutersf:
All specs passed

PR #428, opened 2020-07-02
KAS-1638: Theme Selector Observer Refactor

rafael-mindreau:
# 👁️  Observer Refactor: Themes Selector

## 🔧  What happened?

* Removed Observer
* Octane Refactor as a best effort

I was able to get the same behaviour without the need for an observer, as the code seemed to be using an observer to watch changes that would not likely take place.


woutersf (approved):
✅ jenkins likes it
✅ I like it

rafael-mindreau:
Is now opened in another PR (different branch)

PR #544, opened 2020-10-20
KAS-1786: Remove old css that seems not to be used

BenjaminDendas:
# 🧹 KAS-1786: Remove old css that seems not to be used

## ✏️ What has changed
- removed CSS that seems like not to be used

## 📝 Note:
Ik heb geen changes gezien in de applicatie waar deze css gebruikt werd.. bij deze eruit gehaald.


woutersf:
Out with the old!

woutersf (approved):

PR #555, opened 2020-11-02
KAS-1952: check campaign is recent before send.

woutersf:
Screenshot 2020-11-02 at 14 57 34


app/errors/sending-old-campaign-error.js
```diff
@@ -0,0 +1,16 @@
...
+
+  this.errors = errors || [
+    {
+      title: 'You tried to send an older mailchimp campaign.',
+      detail: message,
+    }
+  ];
+};
+
+SendingOldCampaignError.prototype = Object.create(EmberError.prototype);
```
MikiDi, line 13: [ES6 class syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) plz :pray:

PR #557, opened 2020-11-03
Feature/linter on jenkins

woutersf:


woutersf:
Screenshot 2020-11-03 at 07 18 49

PR #623, opened 2020-12-16
Enable octane

MikiDi:
Adds configuration that specifies octane. This makes that you can use [ember-cli](https://cli.emberjs.com/release/basic-use/cli-commands/#generatemorefiles) in the same way you would use it in your average ember project instead of having to copy-paste, strip, ... stuff for generation.
The generated `component.js` -files within this PR are for keeping non-octane components working the way they did without refactoring them to octane.


BenjaminDendas (changes requested):
Eslint errors fixen, dan zal jenkins opnieuw laten builden.

PR #785, opened 2021-03-31
KAS-2407: column selector modal

MikiDi:
See [KAS-2407](https://kanselarij.atlassian.net/browse/KAS-2407). Refactors the publications overview table column-toggle modal. Different columns are now sourced from a config file. Slightly out of scope of this story, also the initialization of the "show-column-flags" as well as the rendering of the column headers is now sourced from this same config-file.  
Sourcing the headers from the config file also involved some changes to the sorting logic. The sorting now works as before, but please note that the sorting for some columns is still buggy. The error thrown when sorting by "uiterste publicatie" for example is out of scope of this story.


ValenberghsSven (changes requested):
@MikiDi 
Deze zal nog moeten aangepast worden naar aanleiding van een recente change: "derived publication type" => "regulation type"

MikiDi:
Done @ValenberghsSven

erikap (approved):

PR #806, opened 2021-04-01
Don't hide pub search result list when clicking element inside

MikiDi:
Closing the publication search-result-list on `focusout`-event introduced a side-effect which closed the search-result-list before the click reached the link. Clicking a certain publication as a result didn't navigate to the desired publication, but just closed the result list. This PR doesn't hide the result-list when clicking something inside it.


ValenberghsSven (changes requested):
image

resultListDomElement kan null zijn.
die error komt als je op het input veld klikt

PR #884, opened 2021-05-15
KAS-2378 Upgrade ember simple auth

erikap:
Remove usages of ember-simple-auth mixins:
- Extend session service to replace ApplicationRouteMixin
- Use session.requireAuthentication instead of AuthenticatedRouteMixin
- Use session.prohibitAuthentication on login routes

Octanify and cleanup current-session service


app/components/m-header.js
```diff
@@ -26,8 +26,7 @@ export default class MHeader extends Component {
       this.environmentClass = 'vlc-environment-pill--test';
     }
 
-    if (window.location.href.indexOf('https://kaleidos.vlaanderen.be') === 0
-      && this.currentSession.checkIsDeveloper()) {
+    if (window.location.href.indexOf('https://kaleidos.vlaanderen.be') === 0) {
```
MikiDi: Deze willen we waarschijnlijk in z'n geheel weg ... de eindgebruikers hebben weinig boodschap aan de pill `PROD` (?). +
- if -> else if
- window.location.hostname met straight `===` ipv `indexOf` ...

app/pods/application/route.js
```diff
@@ -1,30 +1,50 @@
...
 
-export default Route.extend(ApplicationRouteMixin, {
-  moment: service(),
-  intl: service(),
-  currentSession: service(),
-  fileService: service(),
-  routeAfterAuthentication: 'agendas',
-  router: service(),
-  metrics: service(),
+function isSupportedBrowser() {
```
MikiDi: We gebruiken [elders](https://github.com/kanselarij-vlaanderen/kaleidos-frontend/blob/45b396bc483ce05cc3d45fe5d0cd0ec35a0a68e4/app/components/web-components/vl-rdfa-editor.js#L17) al https://github.com/willviles/ember-useragent ... misschien de moeite om mee te nemen?

app/components/m-header.js
erikap: Correct. 
Ik heb de logica nog wat verder opgekuist om op hostname te checken en met config-map te werken ipv `if-else`.

app/services/session.js
```diff
@@ -0,0 +1,25 @@
...
+      await this.currentSession.load();
+    } catch (error) { // eslint-disable-line no-unused-vars
+      this.invalidate();
+    }
+    super.handleAuthentication('agendas');
+  }
+
+  handleInvalidation() {
+    const logoutUrl = ENV.torii.providers['acmidm-oauth2'].logoutUrl;
+    if (logoutUrl.startsWith('http')) {
```
MikiDi: I this is a validation, it isn't super self-documenting. a `try ... catch` with [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) might be better suited?

app/services/session.js
```diff
@@ -0,0 +1,25 @@
...
+      super.handleInvalidation(logoutUrl);
+    } else {
+      super.handleInvalidation('aanmelden');
```
MikiDi: Isn't an ember route (`login`) name expected here?

app/services/current-session.js
```diff
@@ -1,171 +1,71 @@
...
-  checkIsDeveloper() {
-    // The keys in this whitelist are only indicative
-    // They are not used, but use them to describe who the id belongs to
-    // Get these from https://kaleidos.vlaanderen.be, and then check the user id via the ember data tab
-    const whitelist = {
-      frederik: CONFIG.frederik,
-      ben: CONFIG.ben,
-      rafael: CONFIG.rafael,
-    };
+  get isOvrb() {
```
MikiDi, line 52: Future work, but I find these to be misleading. role-like naming would be better here imo, eg `canEditAgenda`, `canViewPublications`, `canEditPublications`, `canEditNewsItems`, ....

app/services/session.js
erikap: Indeed. Copied that snippet from lblod/loket :see_no_evil: 
Will change this. `try ... catch` is a better approach and will also handle the case if `logoutUrl` is not configured.

app/services/session.js
erikap: That's what I tried first, but didn't work. Will double check to be sure and add a comment.

app/services/current-session.js
erikap: Agree, but future work indeed.

app/services/session.js
erikap: Looking at the default implementation in ember-simple-auth (https://github.com/simplabs/ember-simple-auth/blob/80b9c152ff3e13e9d03b309a7a4195400a7533b2/packages/ember-simple-auth/addon/-internals/routing.js#L52-L60) it seems to expect a route name when FastBoot is enabled and a URL or path otherwise. I will refactor so we don't use `super.handleInvalidation` and just implement our own logic.

PR #953, opened 2021-06-14
Bugfix/kas 2506 reopen agenda warning for documents

ValenberghsSven:
## What has changed

agenda-header.hbs:
- Added part to confirmation pop-up body for showing the pieces we need to delete
- Changed footer to disable button when calculating pieces, different text depending on whether or not we have pieces to delete

agenda.header.js:
- Added task to fetch all pieces that are new on the designagenda by looping agendaitems and set to tracked variable
- Added part to reopenPreviousAgenda action where we delete all the fetched pieces

agenda.js:
- Made it so agenda.nextVersion and agenda.previousVersion don't save from frontend. Set in backend service and should be read only

agenda-service.js:
- Method was inconsistently throwing errors on queryRecord getting multiple ember objects or something (I suddenly had 8 pieces in my list when I was only expecting 2) . Changed to the queryOne merthod and syntax.

nl-be.json:
- Added translations for new message and button


MikiDi (approved):

MikiDi:
>     Method was inconsistently throwing errors on queryRecord getting multiple ember objects or something (I suddenly had 8 pieces in my list when I was only expecting 2) . Changed to the queryOne merthod and syntax.

Probably (my) missing awaits were what was going wrong ...

PR #975, opened 2021-06-29
Bugfix/kas 2601 pieces go missing

ValenberghsSven:
## What has changed:

model-with-modifier.js:
- Changed setting of modified to a function call
- Added said function call to 3 of the 4 options of dirtyType. (fix)
- Commented why we need dirtytype = undefined to save modified data

agendaitem/documents/controller.js:
- Changed checking concurrency action to a task
- reuse task
- add concurrency checking task to right before we do important actions (failsafe)

agendaitem/documents/template.hbs:
- async method was not awaited, now task is performed and concurrency correctly triggers (fix)

Thoughts:
- I contemplated adding a reload for pieces before saving and decided that it will help in certain situations
- The concurrency check during save, does this need to happen earlier, like before saving the pieces ? If we fail there, the pieces will be created but not linked to anything (except container and files). Orphaned data will exist.
- The saving of agendaitem here is only to change the formal ok status. A follow-up story should be made to extract that functionality to an API call to a service to avoid ember PATCHing everything, potentially overwriting new data with old data (like treatments -> known bug)


erikap (approved):
Ok als bugfix. Ik ben het eens met wat Michael op de chat zei rond die `dirtyType`, maar uitspitten hoe we bij het dirty checking ook eventueel relaties in rekening kunnen brengen (of een andere structurele oplossing) lijkt me ver buiten de scope van deze bugfix.

ValenberghsSven:
DEZE PR niet mergen om te vermijden dat we extra dev zaken meepakken (wou deze in accept hebben)

Ik herhaal deze code changes op ACC, test lokaal tegen ACC backend en maak versie bump als alles ok is

ValenberghsSven:
Code opnieuw gedaan in ACC branch
https://github.com/kanselarij-vlaanderen/kaleidos-frontend/commit/ff5a9e6669ca8a4907ea8086984943c9fc8ecafa

PR #979, opened 2021-06-30
Bugfix/kas 2601 pieces go missing overview

ValenberghsSven:
# :warning: STARTED FROM ACC BRANCH 

## What has changed:

agenda-header.js:
- removed commented loader in existing method
- added the reloading of agendaitem relations to existing method (only for items not "formeel ok")
- try catch in approvallAgendaitems method, could get stuck with infinite overlay modal when errors were thrown

nl-be.json:
- correct business translation


ValenberghsSven:
Testen zijn groen, ga deze mergen

PR #998, opened 2021-07-08
KAS-2713 fixed problem by checking if there's a subcase

TaneeG:


ValenberghsSven (approved):
Jenkins groen :heavy_check_mark:

PR #1009, opened 2021-07-14
Cypress/kas 2695 improve existing tests part 4

ValenberghsSven:
Done (by Mathias):
    - removed within, should exist, should be visible
    - replaced css with selectors
    - removed unnecessary assertions, waits
    - removed duplicate tests

Done (By Sven):
    - cleanup final todo's in specs (excluding commands and publication.spec)
    - rename remaining todo's in specs to have a topic for later listing (excluding commands and publication.spec)
    - cleanup resetCache usage + test (kept resetCache command, can still be used locally for faster reset of data)
    - removed all occurences of `cy.contains('text')`, use a selector first to make sure we know what we are testing (excluding commands and publication.spec)
    - moved system-alert spec to unit folder, removed the existing folder for this spec
    - when using `datatable` selectors, made sure we select the rows in the body, rather then just any children (which include the pagination and the header and could fail on length asserts)
    - added selectors where needed to accomplish the above


ValenberghsSven:
Search spec had 1 test dat was disabled (TIKA dat pdf verwerkt en dan kunnen op die inhoud zoeken)
Die test werkte niet omdat cypress file upload lege PDF's upload.
Heb dat kunnen fixen en nu werkt dat wel :tada: 

jenkins is groen
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/FeatureBranches/view/change-requests/job/PR-1009/2/

PR #1018, opened 2021-07-24
Reduce number of included resources in publications index model hook

erikap:
Adding included resources reduces the number of requests sent to the
backend but also has a performance penalty: the executed query takes
longer and cache entry gets invalidated quickly due to the agressive
cache clearing.

This PR only includes resources that are rather 'stable' to lower
chances of cache invalidation. Other resources are loaded in additional requests.

Loading additional data in separate requests also has a performance
benefit for user that display only a subset of the columns in the
publications data table. If a column is not shown the additional data
will not be loaded, while in the current implementation data is always loaded.


MikiDi:
I don't have a dataset at hand which is suitable for testing this. Something like 50 publication-flows on 1 page would be a good test for this. I'm still slightly concerned with "jumping" while loading, although this effect should be far less noticeable on the horizontal axis as on the vertical one (as the agenda overview used to have).
I'm going to merge this and adapt later if necessary.

PR #1031, opened 2021-07-30
KAS-2655 dropdown-menu

MikiDi:
Replaced all direct uses of `EmberAttacher` which were standard use-cases by a new custom wrapping component `DropdownMenu`, which makes for a more straightforward component-api. Covers label-icon as well as icon-only buttons.

@brenner-company To me it looks like the entire contents of https://github.com/kanselarij-vlaanderen/kaleidos-frontend/blob/development/app/styles/custom-application/ember-attacher.scss can be deleted (each selector for its own reasons). Please verify if so and add to this PR.

To be noted: some older components have links within menu-items, which were left untouched by this PR. These links make for menu-items which contain nested links when translated to html. Also related:  https://github.com/kanselarij-vlaanderen/kaleidos-frontend/blob/69ebe7f4d4e541d44b143cb1c44d33d1184807fb/app/components/auk/menu/item.hbs#L11 . This doesn't have any urgent consequences apart from some bolder text, but should be tackled in the future for consistency.

To be discussed: I created the `DropdownMenu`-component in the root component namespace in the reasoning that it is an app-specific integration of the generic `Auk::Menu` component, but if we consider this integrating component generic enough, it can easily be moved to the Auk-namespace.


brenner-company:
@MikiDi Momenteel ben ik terug nog even op verlof (tot volgende week dinsdag). Als het tot dan kan wachten dan neem ik het direct op als ik terug aan de slag ga, anders heb heb ik wel vertrouwen in zowel jou als Erika haar review capaciteiten 😉

Ook even kort meegeven dat er binnen PR #1030 een aantal kleine aanpassingen zitten mbt de dropdown/menu, maar op het eerste lijkt dat zonder probleem merge-baar.

erikap (approved):

PR #1119, opened 2021-11-10
KAS-2954 removing confidentiality from case

TaneeG:
jira : https://kanselarij.atlassian.net/browse/KAS-2954
backend: https://github.com/kanselarij-vlaanderen/kaleidos-project/pull/240
jenkins : http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/134/


ValenberghsSven:
Ik heb de vraag gesteld op 22/11/2021 of we bij overname vorige procedurestap ook de vertrouwelijkheid moeten overnemen

app/components/cases/new-subcase.js
```diff
@@ -80,7 +80,7 @@ export default class CasesNewSubcase extends Component {
       type: this.type,
       shortTitle: trimText(this.shortTitle),
       title: trimText(this.title),
-      confidential: this.args.case.confidential,
+      confidential: false,  // Set default to false as it isn't derived from case anymore
```
MikiDi: `it isn't derived from case anymore` suggests that the property still exists on case, but isn't taken in to account here. I would suggest to remove this comment.

MikiDi (approved):

app/components/cases/new-subcase.js
ValenberghsSven: Strange that you still saw that comment, as I removed it in a previous commit
https://github.com/kanselarij-vlaanderen/kaleidos-frontend/pull/1119/commits/deccd6edf2c76ea607eb21035f0ac7a9e4fc060f

ValenberghsSven: But the code will change again, since there will be a toggle added to the modal so you can create a confidential subcase instead of having to change it right after creating a normal one

PR #1127, opened 2021-11-22
Feature/kas 2449 session service current session refactor

ValenberghsSven:
This PR combines 3 tickets in 1 PR.

KAS-2448 (currentAgenda removed)
KAS-2449 (currentSession and sessionService removed)
KAS-2991 (bug found & fixed during KAS-2449)

Because KAS-2448 was only very little work and touches the same files as KAS-2449.


app/pods/agenda/route.js
```diff
@@ -20,7 +19,7 @@ export default class AgendaRoute extends Route {
     const agenda = await this.store.findRecord('agenda', agendaId, {
       reload: 'true',
     });
-    const reverseSortedAgendas = await this.store.query('agenda', {
+    const reversedAgendas = await this.store.query('agenda', {
```
MikiDi: `reverseSortedAgendas` vs `reversedAgendas`. I must say I found the original name clearer

app/components/agenda/compare-agenda-list.js
```diff
@@ -7,8 +7,9 @@ import { tracked } from '@glimmer/tracking';
 export default class CompareAgendaList extends Component {
   /**
    * INFO arguments from parent.
-   * @reversedAgendas all agendas from the meeting reverse sorted on serial number
-   * @isShowingChanges flag to show all agendaitems or just the ones with changes
+   * @argument meeting the current meeting
+   * @argument reversedAgendas all agendas from the meeting reverse sorted on serial number
```
MikiDi: The need for the reverse sorting is specific to this component (-functionality), right? Would it be better to move the sorting responsibility inside the component?

ValenberghsSven: It may not be clear from this PR, but reversedAgendas is used in components.
- Agenda-actions
- side-nav
- agendaitem-controls
- compare-agenda-list

ValenberghsSven: It may not be clear from this PR, but reversedAgendas from the top route is used in 4 components.
- agenda-actions
- side-nav
- agendaitem-controls
- compare-agenda-list

app/pods/agenda/route.js
ValenberghsSven: Initially the full name was more clear.
When I started using it on multiple components, the name seemed too long and knowing the contents it seemed to make sense to shorten it.
For clarity, will rename all references (36) to the full name

PR #1151, opened 2021-12-20
KAS-3031 use different property from router

ValenberghsSven:
The issue:
we can get stuck in the /loading route when transitioning too fast when changing the params of the route.

Affected area: 
in the sidebar of agenda overview, agendaitems have a linkTo to switch agendaitems while keeping the current route.
Switching between agendaitems fast causes this issue because we automatically transition to a loading state multiple times.

The root cause:
Right now we are using `currentRouteName` from router in that linkTo, but there is also a `currentRoute.name`.
The first one actively changes to `.x.x.loading`.
The second one doesn't seem to yield the loading route and always shows the current "real" route.
So when switching agendaitems the first tim, we get the loading route from the `currentRouteName`, recalculating all LinkTo's to have this route, a subsequent click on a sidebar item then opens the loading route, getting us "stuck" in this route. 

The solution:
Since the `currentRoute` can be undefined, we only want the value if it exists and fallback to the `currentRouteName`
So when the route is loading for the first time, `currentRouteName` will always point to an existing route (unless address had /loading) while `currentRoute` is undefined
The first time we enter the route via the interface the `currentRoute.name` will yield a real existing route (index, documents, ..)
All Linkto's will point to that route (with different models).
When we start clicking fast to switch agendaitem, the first click will be:
currentRouteName = x.x.x.index
currentRoute.name = x.x.x.index
The second click will be:
**currentRouteName = x.x.x.loading**
currentRoute.name = x.x.x.index
The second click will be:
**currentRouteName = x.x.x.loading**
currentRoute.name = x.x.x.index
etc.. 
Both properties have different values and we want the correct index.
from manual tests, currentRoute.name never seems to have ".loading" so we grab that vlue to build to LinkTo's


ValenberghsSven:
`currentRouter` is basically the previous route it seems like.
When coming from search, `currentRouter.name` = 'search.agendaitems'.
So first we check if the currentRouter matches our base route "agenda.agendaitems.agendaitem." or fall back to `currentRouterName`

MikiDi:
TODO @MikiDi code-comments verduidelijken, daarna mergen.

ValenberghsSven:
Kunnen we dat omdraaien, eerst mergen daarna comments verduidelijken op development?

PR #1152, opened 2021-12-22
KAS-3053 bring back options but grey them out

ValenberghsSven:


ValenberghsSven:
Not merging, buttons not needed

PR #1199, opened 2022-02-14
KAS-3126 used clickOutside with conditional block

ValenberghsSven:
This is a strange one, let me explain.

For the config modal we are using a `AttachPopover` and expect somewhat `Auk::Modal` behavior, which is not what it was designed for.

AttachPopover is meant to show info or actions in a popup that could be clicked. After the click you can decide if the popover should close or stay open with properties (like a dropdown menu). In most cases it is not meant to stay open.

**current**
`@showOn="click"`: when you click the button attached to the popover, the popover toggles open.
We added an "on click" event for the button to change a property and `@isShown`, but the popover will open regardless of the local variable.
`@hideOn="click"` : when you click the attached button again, the popover will close (+ "on click" gets triggered).
This is somewhat broken when first opening the page, the first click opens the popover (local var = true), a second click does not close the popover for some reason (local var = false), the third click closes the popover (local var = true).
From this moment, the `@isShown` is no longer working correctly, the value is true when closed, false when opened..
 `@isShown`: Should show the popover based on a value

**new**
`@showOn=""`: only our "on click" event will open the popover based on the variable tied to `@isShown`
`@hideOn=""` : only our "on click" event will close the popover based on the variable tied to `@isShown`
Added the `ClickOutside` component. Any clicks done outside the inner component (our config modal) will trigger the action.
note: This also means that clicking anywhere on the page when the config modal was closed would open it.
Added the `{{#if}}` block to not render the `ClickOutside` with our config modal when the local var was false.
The button in the config modal still closes the popover (do we still want this button since it's basically useless with click outside or keep it as designed?)

**tested, doesn't work**
`@hideOn="clickout"` : when you click **anywhere outside the attached button** the popover will close.
Since our entire config modal is outside our attached button, the modal would close by ticking a box or clicking anywhere on the page or config modal.
also, the outside clicking does not trigger the "on click" event, so our local variable would be wrong
`@onChange={{this.toggleColumnsDisplayConfigPanel}}`: this throws errors because multiple sources are trying to change the local variable at the same time.
`@hideOn="escapekey"` pressing the escape key works, but doesn't change the local variable, not usable with new solution

Working purely with popover functionality (no `@onChange`, `@isShown` or local var):
`@showOn="click"`
`@hideOn="click escapekey clickout"`
The opening and closing with the button: ok
Pressing escape closes the popover: ok
Clicking outside the model closes the popover: ok
Clicking on a checkbox closes the popover: nok
the close button on top of the config no longer works (because we no longer use the variable): nok

using `@hideOn="clickout escapekey"` `@showOn="click"` but with `@renderInPlace={{true}}`:
Here, the clickoutside seems to work as intended. Click inside the config is fine, click outside the config or escapekey = closes.
However, clicking the button again does not work, this would require  `@hideOn="click clickout escapekey"`. But this also causes the issue that a simple click will close the popover unless we use the `@isShown` with variable but the variable will not get updated when clicking outside or pressing escape.
 `@renderInPlace={{true}}` or  `@renderInPlace={{false}}` changes the behavior of `@hideOn="clickout"`

**Conclusion**
The popover does not see the config modal as "a part of the popover" in regards to clicking outside (seems to depend on value of `@renderInPlace`). Clicking in the inner component of the popover should not be considered "clickout". If "clickout" would work, we may not need the extra `#if` and component and could just work on popover alone.
The mixing of both `@showOn`/`@hideOn` and `@isShown` does not seem to work properly, opt for only 1 of the two (the latter in this case)
The problem lies partially with popover, some options just don't work well together. Perhaps we are using it for the wrong purpose.

The current changes make this work like a charm but may not seem like the cleanest solution.
Clicking the open button or the close(X) button will change the local variable and close/hide the popover
Clicking outside will change the local variable and close/hide the popover
Clicking inside will not close the popover.
Pressing escapekey wasn't requested in ticket, so I didn't add it. But should we? an "on keydown" event or maybe 
https://github.com/adopted-ember-addons/ember-keyboard

https://dtang.dev/listening-to-the-escape-key-with-ember-keyboard/


erikap (approved):
Thanks for explaining the reasoning behind this PR :+1:

MikiDi:
~Ik ga deze mergen~ (gebeurd terwijl ik typte). Technische issues terzijde zou ik wel nog graag weten hoe dit in een ideale wereld hoort te werken. In de huidige setting creeert de settings-modal geen overlay. Intuitief zou ik verwachten dat een eerste keer buiten de modal klikken geen enkel ander gevolg heeft dan de modal die verdwijnt. Als je nu naast de modal klikt op een rij, navigeer je daar ook meteen naartoe. Hoe hoort zoiets in theorie @brenner-company ?

ValenberghsSven:
> ~Ik ga deze mergen~ (gebeurd terwijl ik typte). Technische issues terzijde zou ik wel nog graag weten hoe dit in een ideale wereld hoort te werken. In de huidige setting creeert de settings-modal geen overlay. Intuitief zou ik verwachten dat een eerste keer buiten de modal klikken geen enkel ander gevolg heeft dan de modal die verdwijnt. Als je nu naast de modal klikt op een rij, navigeer je daar ook meteen naartoe. Hoe hoort zoiets in theorie @brenner-company ?

Idd net gemerged na updaten met dev/ drone build ok
Door het ontbreken van een overlay kan je inderdaad op een rij klikken met de modal open.
Dat vind ikzelf niet meteen slecht, anders ga je altijd 2 keer moeten klikken met de modal open, 1 keer om de modal te sluiten en nogmaals om te doen wat je wou (navigeren naar andere route, openen publicatie, aantal per pagina wijzigen).

brenner-company:
Dit is inderdaad wel een grensgeval moet ik toegeven. In de meeste gevallen is modal quasi altijd de logische optie, maar aangezien je hier eigenlijk direct feedback ziet (tabel die kolommen bij krijgt of verliest) en de content binnen de floater (modal, popover, tooltip, etc.) eerder 'beperkt' is, is een popover misschien beter geschikt. Nu, welke je ook kiest, de al dan niet overlay aanwezigheid en klik gedrag moet dan wel gevolgd worden. Zaken combineren lijkt met een verwarrend pad (zowel voor devs als gebruikers) om te volgen. 

Zoals ik het nu zie op [kaleidos-dev.vlaanderen.be](url) is het ok voor mij. Buiten dan iets kleins dat mis gaat qua padding binnen de button zelf (maar dat noteer ik voor mezelf om op te nemen):

CleanShot 2022-02-18 at 11 01 49@2x

PR #1202, opened 2022-02-18
KAS-3121 shared task in new service with max concurrency

ValenberghsSven:
Extracted the query used in 2 classes (1 controller, 1 component) and put them in a service.
Tried limiting the queries in the classes themselves, but this doesn't work for component since there are multiple instances of that component (1 per agendaitem)
Only limited the query portion of the tasks, the processing of the queried data doesn't have to be limited and is different for each class.

Not sure about the chosen name for the service.
I wasn't able to figure out how to make a util out of task (like util/documents.js)

Adding a big timeout to test (5s), really shows that only a maximum of 3 `pieces?filter` calls are done concurrently while initially loading the route (controller task) and when loading the individual components


erikap:
> Not sure about the chosen name for the service.

Something with `throttled` to put focus on that feature of the service?  `throttled-loading-service`?

> I wasn't able to figure out how to make a util out of task (like util/documents.js)

Given you need to inject the `store` service, it's not that simple. There are ways to do that in Ember, using initializers, but that doesn't make it cleaner/easier to understand.

erikap (approved):

ValenberghsSven:
Merging to dev. Only known flaky tests are failing
Was considering making a separate build with this, but if it doesn't work as expected or slows the agenda loading down to much, we can just increase the max concurrency or remove it completely and just keep the shared service call.

PR #1208, opened 2022-02-23
KAS-3162 add a modal to edit the endDate of translationActivity

ValenberghsSven:
- Added modal to edit the endDate of a translationActivity.
- Added DDAU action in controller to handle the save
- Deleted unused key from nl-be.json

The story + rocket-chat do not give a lot of clarity besides adding a modal that can alter the date.
https://kanselarij.atlassian.net/browse/KAS-3162

My questions:
- When uploading the first time, the date is used to fill in `piece.receivedDate`, is checked against `publicationSubcase` dates and updates the publication status, is none of this needed on edit? I would assume so, but not requested so not done.
- First I was unsure if it should have been possible to clear the date but that would cause the `TranslationReceivedPanel` to disappear because there's a check on it here:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/b821681d8dc916d94af48792ec55fe47f80c9154/app/pods/publications/publication/translations/index/template.hbs#L45
So disabled the save button when the date is invalid to counter this.
Not sure why that check on endDate is even needed, since it seems impossible to upload a translation without it. 
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/publications/publication/translations/translation-upload-modal.hbs#L47


ValenberghsSven:
compare with `proof-edit-modal` from https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1207.
Determine what the best solution is and I will update this PR accordingly
- use local modal or new component?
- local save in component or DDAU to controller?
- toggle getter vs open/close getter?
- use validators?

erikap:
> * When uploading the first time, the date is used to fill in `piece.receivedDate`, is checked against `publicationSubcase` dates and updates the publication status, is none of this needed on edit? I would assume so, but not requested so not done.

It's correct to update `piece.receivedDate` as well. The date the translation documents are received should be the same as the end-date of the translation-activity. 

For the subcase, I hope you meant `translationSubcase` instead of `publicationSubcase`. Receiving a translation document puts an end-date on the translation-subcase. Not on the publication-subcase. And that one should also be the same as the `piece.received-date` and `translation-activity.end-date`.

I must admit that most of these dates originate from the previous UI where all those dates we're relevant, visible to the user and used to derive the publication-flow status from. That's not the case anymore: not all dates are shown anymore and status is managed explicitely by the user now. So I'm thinking we're making things maybe a bit too complex now trying to keep everything in sync in all cases in the frontend. Maybe it's easier to just capture a single date and make a reactive backend service that ensures everything stays in sync. But that should become a future story; not something to solve here. @MikiDi This discussion is related to the story of the consistent data on status updates.

> * First I was unsure if it should have been possible to clear the date but that would cause the `TranslationReceivedPanel` to disappear because there's a check on it here:
>   https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/b821681d8dc916d94af48792ec55fe47f80c9154/app/pods/publications/publication/translations/index/template.hbs#L45
>   
>   So disabled the save button when the date is invalid to counter this.
>   Not sure why that check on endDate is even needed, since it seems impossible to upload a translation without it.
>   https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/publications/publication/translations/translation-upload-modal.hbs#L47

The check is required because the translation-activity is already created, with only a start-date, on creation of the request-activity. But the panel should only be shown in the timeline if a translation document is received (and thus the end-date of the translation-activity is set). Disabling the save button looks like a good solution to me.

erikap:
> compare with `proof-edit-modal` from #1207. Determine what the best solution is and I will update this PR accordingly
> 
> * use local modal or new component?

Since it's a very simple modal, only used in 1 place, I'm fine with a local inline model.

> * local save in component or DDAU to controller?

Handling in the controller seems to be inline with `@onDeleteRequest` of the request-panel.

> * toggle getter vs open/close getter?

I'm more in favor of open/close since that cleary shows the intention of the action and ensures you're always in a consistent state. With `toggle` you always implicitely assumes you now what the previous state is. That sometimes yields unexpected behaviour if for example toggle is called twice (which is rather unlikely in a simple  component as this one, but still).

> * use validators?

As already mentioned a couple of times, we should limit user input as little as possible. Only if it breaks the app/data otherwise (like allowing an empty date in the example above). Using the validators for these cases is fine for me, but nevertheless I think in the long run we should start looking at community solutions for validations instead of reiventing the wheel ourselves.

ValenberghsSven:
"The check is required because the translation-activity is already created, with only a start-date, on creation of the request-activity"
Ah now I see. Missed that. Check is ok then.

ValenberghsSven:
Updated PR:
- Made it more consistent to proof code.
- Expanded the saving to include pieces update and translationSubcase update (check needs revising later)
- Opted for open/close vs toggle
- linter/prettier (also on some unrelated files)

ValenberghsSven:
Going to merge this one. Anything discussed on the updating of pieces.receivedDate or the translationSubcase date check on edit can happen later.

PR #1209, opened 2022-02-23
Feature/kas 2308 newsletter

Annelies-P:
Contains the changes from KAS-2308_Refactor_newsletter_service


app/services/newsletter-service.js
```diff
@@ -1,84 +1,121 @@
...
+      body: JSON.stringify(body),
+    });
+    const result = await response.json();
+    if (!response.ok) {
       this.toaster.error(
         this.intl.t('error-create-newsletter'),
         this.intl.t('warning-title')
       );
-      return null;
+      throw new Error('An exception ocurred: ' + response.error);
```
ValenberghsSven: Any specific error we throw from the backend should come out of this, so the user has some feedback of what happened.
But the error we throw here with a message is never used again so we lose that extra bit of information of what went wrong, just a vague message "something went wrong while ...."

ValenberghsSven: Also `response.error` is not correct.
`result` gets the json object with a list of errors
image
while `response` gets the fetch call or something and doesn't contain the proper info
image

ValenberghsSven (changes requested):
Not done with review yet but wanted to close off the day with found issues.
Error handling is not really what I expected yet. 
note: I am comparing it to what I did in `agenda-actions.js`. Where I included the server message in the toaster message. It might not be the best approach so open for debate.

app/services/newsletter-service.js
Annelies-P: I see your point. The problem here is that it is possible the error in the backend is linked to mailchimp. Do we want to bother the user with this? Or is showing a more detailed message like 'Something went wrong creating a mailcampaign' sufficient?

ValenberghsSven: My idea about this is that we receive custom error messages from the backend service in order to identify the exact problem. The user may not really care why something failed but it makes the follow-up communication easier if they can tell us what error they got. 
Vague messages causes speculation and the users tend to get a wrong idea of why something went wrong.
But this is just an opinion, we could start by just logging the correct errors in console (using the errors array from `response.json`).

Errors from this service should be few or none at all.
Besides connection issues with mailchimp and belga, the only other errors can come from user mistakes:
- trying to send a mailcampaign with no Nota's selected (f.e. EP agenda with only 1 announcement, happened in the past)
- trying to send a mailcampaign when there are no themes selected on Nota's

We could translate the errors in the specific cases but maybe not worth the effort.

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -60,36 +60,42 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+
+      if (this.mailCampaign?.isSent) {
+        this.toaster.error(this.intl.t('error-already-sent-newsletter'));
+      } else {
+        if (yield this.validateMailCampaign()) {
+          try {
+            yield this.newsletterService.sendMailCampaign(this.mailCampaign.campaignId);
+            this.mailCampaign.sentAt = new Date();
+            yield this.mailCampaign.save();
+            yield this.args.meeting.belongsTo('mailCampaign').reload(); // TODO Why?
```
ValenberghsSven: Regarding this TODO.
This was probably a fix in some cases, hard to tell what the exact reason/cause was.
I vaguely remember issues where `meeting` did not get updated locally with the `mail-campaign` and in some cases a `meeting.save()` happened later, deleting the relation between mail-campaign and meeting.
Will look into this in next review.

ValenberghsSven (approved):
Errors are logged in console :+1: 

Restored an error message change that was lost in a merge.
Mailchimp fails was showing Belga errors.

Remove the TODO and the reload.
Didn't seem needed in this case.
The fix I mentioned that I once did was the `reload:true` on `findRecord('meeting'`.

PR #1216, opened 2022-03-03
Feature/KAS-3207 Refactor van eenvoudige ember-data models naar Octane

sergiofenoll:
Jira: https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-3207&assignee=621cf63114cd24006906e31c

Backend PR: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/255

Refactored some data models to Octane

Refactored:
- access-level
- agendastatus
- account-group
- account
- agenda-item-treatment
- alert-type
- decision-result-code
- theme
- user
- document-type
- file

Removed:
- confidentiality
- formally-ok
- mandatee-state
- meeting-record
- responsibility
- submitter

Used following regex to double check that removed models aren't used anywhere in the codebase: `(store\..*|belongsTo|hasMany)\(['"]NAME HERE['"]`


app/models/document-type.js
```diff
@@ -1,23 +1,11 @@
...
 
-// TODO: octane-refactor
-// eslint-disable-next-line ember/no-classic-classes
-export default Model.extend({
-  label: attr('string'),
-  scopeNote: attr('string'),
-  priority: attr('number'),
-  altLabel: attr('string'),
+export default class DocumentType extends Model {
+  @attr('string') labe;
```
MikiDi: spotted a typo here :eyes:

app/models/account-group.js
```diff
@@ -1,9 +1,8 @@
...
 
-// TODO: octane-refactor
-// eslint-disable-next-line ember/no-classic-classes
-export default Model.extend({
-  uri: attr(),
-  name: attr(),
-  users: hasMany('user'),
-});
+export default class AccountGroup extends Model {
+  @attr() uri;
```
MikiDi: minor. When using the decorator notation without specifying a transformer, you can drop the parens.

app/models/user.js
```diff
@@ -1,41 +1,22 @@
...
-      return value;
-    },
-  }),
+  @belongsTo('account') account;
+  @belongsTo('account-group', { inverse: null }) group;
+  @belongsTo('organization') organization;
 
-  fullName: computed('firstName', 'lastName', function () {
+  get email() {
+    return this.emailLink && this.emailLink.replace(/^mailto:/, '');
```
MikiDi: Not in scope of this PR, but for future work, you might want to learn about https://github.com/mu-semtech/ember-mu-transform-helpers/ (if you haven't already ;-) )

app/models/document-type.js
sergiofenoll: Gah :facepalm:
Thanks for catching that :relieved: !

app/models/user.js
sergiofenoll: I checked and the transform-helpers package was already a dependency, so I added the transforms here. Left the `email` and `phone` getters as aliases (less work :stuck_out_tongue:)

app/models/account.js
```diff
@@ -1,10 +1,11 @@
...
-  user: belongsTo('person'),
-  gebruiker: alias('user'),
-  voId: attr('string'),
-});
+export default class Account extends Model {
+  @attr('string') voId;
+
+  @belongsTo('person') user;
+
+  get gebruiker() {
```
erikap, line 13: Can you verify whether this property is actually used? If so, can it be replaced with `user`?

app/models/agenda-item-treatment.js
```diff
@@ -1,27 +1,18 @@
...
-    return this.intl.t('signed-document-decision', {
-      // eslint-disable-next-line ember/no-get
-      name: this.get('report.lastPiece.name'),
-    });
-  }),
-  publicationFlows: hasMany('publication-flow'),
-  signFlows: hasMany('sign-flow'),
-});
+export default class AgendaItemTreatment extends Model {
+  @service intl;
```
erikap: Service is not used anymore. Can be removed.

app/models/agendastatus.js
```diff
@@ -1,19 +1,17 @@
...
-  isDesignAgenda: computed('uri', function() {
+export default class AgendaStatus extends Model {
+  @attr('string') uri;
+  @attr('string') label;
+
+  get isDesignAgenda() {
     return this.uri === CONSTANTS.AGENDA_STATUSSES.DESIGN;
-  }),
-  isFinal: computed('uri', function() {
+  }
```
erikap, line 10: Nitpicking, add newlines between get-functions

app/models/user.js
```diff
@@ -1,41 +1,22 @@
...
+  @belongsTo('account-group', { inverse: null }) group;
+  @belongsTo('organization') organization;
 
-  fullName: computed('firstName', 'lastName', function () {
+  get email() {
+    return this.emailLink;
+  }
+  get phone() {
+    return this.phoneLink;
+  }
```
erikap: Can you just rename the `{email,phone}Link` properties to `{email,phone}` in `mu-cl-resources` to get rid of these aliases?

erikap (changes requested):

app/models/account.js
sergiofenoll: Yeah, it's used in the `@lblod/ember-mock-login` package here: https://github.com/lblod/ember-mock-login/blob/3aafb710617f63783343ac32eff29ddaf19b67a0/addon/components/login/each-account.hbs#L4

That's the only place though, perhaps I can add a comment and when/if that gets replaced in the package we can remove this getter?

erikap: Comment is probably a good idea :+1:

erikap:
Triggered a custom frontend/backend run on Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/163

erikap:
> Triggered a custom frontend/backend run on Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/163

Tests in `all-flaky-tests/subcase.spec.js` fail due to an error on subcase updates. Mu-cl-resources seems to return a 500: `Ember Data Request PATCH /subcases/62214C8AB0A4E8576A920206 returned a 500`. @sergiofenoll can you have a look?

app/models/account.js
sergiofenoll: Will add the comment. Looking at this, shouldn't user be a relationship with `user` instead of `person`? Looking at the resource definition I think that's the case. Only noticed because I was locally trying the upgrade to 3.28 and at that version ember complains about the mismatch.

erikap: Yes, should indeed be `user` looking at mu-cl-resources config :+1:  
I was a bit confused how the mock-login used to work then? :confused: I assume Ember is currently casting the `user` resource it gets from the backend in a `person` ember-data record? And since both have `firstName` and `lastName` properties we didn't notice the mismatch until now...

sergiofenoll:
> > Triggered a custom frontend/backend run on Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/163
> 
> Tests in `all-flaky-tests/subcase.spec.js` fail due to an error on subcase updates. Mu-cl-resources seems to return a 500: `Ember Data Request PATCH /subcases/62214C8AB0A4E8576A920206 returned a 500`. @sergiofenoll can you have a look?

I missed some subcase <-> access-level accessors, I am guessing that's why mu-cl-resources was complaining. Does the Jenkins build get retriggered automatically or do you/someone need to re-trigger it?

erikap:
> 

Jenkins is automatically triggered for 'regular' builds where you want to run your frontend feature branch from the PR against the development branch in the backend. If you need a feature branch in the backend as well, it needs to be manually triggered. 

I just retriggerd one now for this PR: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/165

erikap:
Custom Jenkins build succeeded. So once account-user relation has been fixed and comment provided for `gebruiker` property, this one can be merged.

ValenberghsSven (commented):
`accessLevel` of subcase is used in frontend
You can see it when editing the titles of a subcase:
image

However this has been broken for a while (selecting does nothing) and no support issues have been logged. So it seems users are not using it. We also have no yggdrasil rules in place for these levels, so the level wouldn't change anything even if it worked.

Should we remove the option when editing titles then?
A quick query on the relation `ext:toegangsniveauVoorProcedurestap` yielded 0 results on PROD

Relevant code here:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/cafd474d99d82fc8c0c48559dfb5c7b613a1eeeb/app/components/subcase/subcase-case/subcase-titles-edit.js#L37

erikap:
> `accessLevel` of subcase is used in frontend You can see it when editing the titles of a subcase: image
> 
> However this has been broken for a while (selecting does nothing) and no support issues have been logged. So it seems users are not using it. We also have no yggdrasil rules in place for these levels, so the level wouldn't change anything even if it worked.
> 
> Should we remove the option when editing titles then? A quick query on the relation `ext:toegangsniveauVoorProcedurestap` yielded 0 results on PROD
> 
> Relevant code here:
> 
> https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/cafd474d99d82fc8c0c48559dfb5c7b613a1eeeb/app/components/subcase/subcase-case/subcase-titles-edit.js#L37

I would indeed remove it from frontend then. @sergiofenoll Can you take care of this?

ValenberghsSven:
@sergiofenoll Besides the subcase-titles-edit component/template, this line can also be removed:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/cafd474d99d82fc8c0c48559dfb5c7b613a1eeeb/app/utils/agendaitem-utils.js#L18

sergiofenoll:
> @sergiofenoll Besides the subcase-titles-edit component/template, this line can also be removed:
> 
> https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/cafd474d99d82fc8c0c48559dfb5c7b613a1eeeb/app/utils/agendaitem-utils.js#L18

@ValenberghsSven thanks, that change should already be in there :D
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1216/files#diff-48fa728cfa4b2c2d07b5054c970c29deb0032dac524f5b854a9565f74dbe87c8L18

ValenberghsSven:
> > @sergiofenoll Besides the subcase-titles-edit component/template, this line can also be removed:
> > https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/cafd474d99d82fc8c0c48559dfb5c7b613a1eeeb/app/utils/agendaitem-utils.js#L18
> 
> @ValenberghsSven thanks, that change should already be in there :D https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1216/files#diff-48fa728cfa4b2c2d07b5054c970c29deb0032dac524f5b854a9565f74dbe87c8L18

ah my bad, glossed over that one.

erikap (approved):

app/models/theme.js
```diff
@@ -1,15 +1,10 @@
...
-    inverse: null,
-  }),
-  deprecated: attr('boolean'),
-});
+export default class Theme extends Model {
+  @attr('string') label;
+  @attr('string') scopeNote;
+  @attr('boolean') deprecated;
+
+  @hasMany('subcase', { inverse: null }) subcase;
```
ValenberghsSven: subcase doesn't have themes, neither does agendaitem (themes used to be on here, maybe some legacy data remains in db)
Newsletter-info does, but there is no point to mention this relation here since we should never ask "what newsletters have this theme"  by doing `theme.newsletters`. We would use a query in that scenario.

app/models/user.js
```diff
@@ -1,41 +1,16 @@
...
-  email: computed('emailLink', {
-    get() {
-      return this.emailLink && this.emailLink.replace(/^mailto:/, '');
-    },
-    // eslint-disable-next-line no-unused-vars
-    set(key, value) {
-      this.set('emailLink', `mailto:${value}`);
-      return value;
-    },
-  }),
```
ValenberghsSven, line 26: Getting `user.email` changed in this file, the adding and removing of `mailto:` no longer happens
To my knowledge we never set the email or allowed it to be edited so the set portion is not needed.

Looking at DEV, this `mailto:` is actually in the data (first off, not sure how Hannes made a user account in DEV, acmidm shouldn't work on this one)
image

quick check on PROD and the `mailto:` seems present in user data.

user.mail is used in these places:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3207-refactor-van-eenvoudige-ember-data-models-naar-octane/app/pods/settings/users/index/template.hbs#L83

https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3207-refactor-van-eenvoudige-ember-data-models-naar-octane/app/pods/settings/users/user/template.hbs#L26

app/models/user.js
```diff
@@ -1,41 +1,16 @@
...
-  phone: computed('phoneLink', {
-    get() {
-      return this.phoneLink && this.phoneLink.replace(/^tel:/, '');
-    },
-    // eslint-disable-next-line no-unused-vars
-    set(key, value) {
-      this.set('phoneLink', `tel:${value.replace(/[/ .]/, '')}`);
-      return value;
-    },
-  }),
```
ValenberghsSven, line 36: same with `user.phone`, the removing of `tel:` no longer happens so the raw string will just be displayed
This is prod data:
image

ValenberghsSven (changes requested):
the changes in user will mean a visual change in the user settings (booting up stack to verify) where we will now display the raw strings with `mailTo:` and `tel:` and longer trim off these bits

Edit: not as easy to test locally since importing users doesn't add emails, we get those from acmidm logins

Edit: the transforms from the addon https://github.com/mu-semtech/ember-mu-transform-helpers/  take care of cleaning the email and phone uri's for display.

app/models/user.js
sergiofenoll: I will verify in a minute, but we're using the transforms `email` and `phone` for the `.email` and `.phone` attributes. These transforms come from this addon: https://github.com/mu-semtech/ember-mu-transform-helpers/ and they handle the stripping of `mailto:` and `tel:`

ValenberghsSven: Ah, I was not aware of this addon tbh.
Not completely sure if I like the approach of this addon, since you can't seem to choose which transforms you want.

it's hardly visible that those transforms get added to your code unless you know about them.

Discussion about the addon is not really PR related, so perhaps we pick up on this later

ValenberghsSven: @sergiofenoll What you say is correct, the transforms are doing their job
the data is showing the prefixes
image

The displayed data does not have the prefixes
image

app/models/user.js
ValenberghsSven: this is done by a transform, resolving

erikap:
Can merge (together with https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/255) @ValenberghsSven ?

ValenberghsSven:
> Can merge (together with [kanselarij-vlaanderen/app-kaleidos#255](https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/255)) @ValenberghsSven ?

 1 test is failing, but I'm pretty sure this is because of the backend changes for user 
(saving of user is failing `PATCH /users/75118ece-c201-4808-b746-696b921c63f1 returned a 406`)
 So to verify, we could run a custom jenkins build.
  
  edit: also check user/person conversation
  https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1216#discussion_r824501408

ValenberghsSven:
custom jenkins build

http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/167/

app/models/account.js
ValenberghsSven: this model was either wrong, or we don't know the reason for this approach. If it was just for mock-login then maybe it's fine with the other changes.
I'm not fully convinced that everything will work perfectly, but I couldn't tell you why it was this way if there was any other reason.

So move fast and break things (after custom jenkins build). I will keep an eye out for any strange behaviour after this change.

ValenberghsSven:
custom jenkins build failed because the frontend branch is up to date with dev, but the backend branch isn't.
Piece model was changed. All failed test are related to saving piece

Willing to just merge, we'll fix any issues on dev (not expecting any)

PR #1217, opened 2022-03-03
Feature/kas 3206 refactor van modal om nieuwe zitting aan te maken naar octane

sergiofenoll:
Jira: https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-3206&assignee=621cf63114cd24006906e31c

Renames the component & refactors for Octane syntax. Replaces as many Vl-components to Auk-components as possible (Datepicker still TODO because `Auk::Datepicker` does not support selecting a time).

Also fixes a bug in the `generateNewNumber()` method where it would break if there were too many future agendas.

Note: didn't run tests (yet), but adapted the selectors to deal with the change from `Webcomponent::VlModalFooter` to `Auk::Modal::Footer`.


app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+
+    assert(
+      `'didSave' argument is required and must be a function`,
+      this.args.didSave !== undefined && typeof this.args.didSave === 'function'
+    );
+    assert(
+      `'onCancel' is required and must be a function`,
+      this.args.onCancel !== undefined &&
+        typeof this.args.onCancel === 'function'
+    );
```
erikap: Currently we don't have these kind of checks in components, so I would leave them out for now. What we try do to more consistently is adding some documentation about the expected arguments. E.g. https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/publications/publication/request-activity-panel.js#L6-L11  You can provide some additional information in that comment block in case the argument name isn't obvious.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      },
+      sort: '-number',
+    });
+
+    if (meetings.length) {
+      const meetingsFromThisYear = meetings.filter(
+        (meeting) =>
+          meeting.plannedStart &&
+          meeting.plannedStart.getFullYear() === this.currentYear
+      );
```
erikap: Can't we add this as an additional filter on the backend query?  Something like `':lte:planned-start': last-day-of-the-year`.

If so, given we sort on `-number`, we only need 1 result. You can use [our custom story extension `this.queryOne`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/store.js#L7) for that.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+    );
+
+    this.generateNewNumber();
+  }
+
+  get isAnnexMeeting() {
+    return isAnnexMeetingKind(this.selectedKindUri);
+  }
+
+  async generateNewNumber() {
```
erikap: Can you convert this to a concurrency-task and disable the meeting number input-field as long as this task is running.
Maybe `initializeMeetingNumber` is a more clear name for the method?

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      const meetingsFromThisYear = meetings.filter(
+        (meeting) =>
+          meeting.plannedStart &&
+          meeting.plannedStart.getFullYear() === this.currentYear
+      );
+      const meetingIds = meetingsFromThisYear
+        .map((meeting) => meeting.number)
+        .filter((meetingId) => meetingId !== undefined);
+      let id = 0;
+      // FIX voor de eerste agenda van het jaar -> Anders math.max infinity
```
erikap: You can remove this comment. It confuses me more than just reading the code :smiley:

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+import CONFIG from 'frontend-kaleidos/utils/config';
+import { isAnnexMeetingKind } from 'frontend-kaleidos/utils/meeting-utils';
+
+export default class MeetingNewMeetingModal extends Component {
+  @service store;
+  @service agendaService;
+  @service newsletterService;
+  @service toaster;
+  @service formatter;
```
erikap: Can you replace usages of this service with inline date formatting? We should get rid of this service in the long run, since it doesn't have an added value, but just adds complexity to the code.
The case `this.formatter.formatDate(null)` can even just be replaced with `new Date()`.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+    const fallBackDate = this.formatter.formatDate(null);
+    const agenda = this.store.createRecord('agenda', {
+      serialnumber: 'A',
+      title: `Agenda A voor zitting ${moment(meeting.plannedStart).format(
+        'D-M-YYYY'
+      )}`,
+      createdFor: meeting,
+      status,
+      created: date || fallBackDate,
+      modified: date || fallBackDate,
```
erikap: As far as I see, `date` is always present in the current code flow. Fallback can be removed.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+    });
+    const savedAgenda = await agenda.save();
+    return savedAgenda;
```
erikap: To be consistent with most of the codebase, I would replace with
```javascript
await agenda.save();
return agenda;
```

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  // new meeting parameter prevents extra request of agenda.createdFor
+  async createAgendaitemToApproveMinutes(agenda, newMeeting, closestMeeting) {
+    const now = new Date();
+
+    // load code-list item
```
erikap: Remove comment. Code is clear imo.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+    const decisionResultCode = await this.store.findRecordByUri(
+      'decision-result-code',
+      CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD
+    );
+
+    // Treatment of agenda-item / decision activity
```
erikap: Remove comment. Code is clear imo.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+        closestMeeting.plannedstart
+      ).format('dddd DD-MM-YYYY')}.`,
+      formallyOk: CONSTANTS.ACCEPTANCE_STATUSSES.NOT_YET_OK,
+      mandatees: [],
+      pieces: [],
+      approvals: [],
+      isApproval: true,
+      treatments: A([agendaItemTreatment]),
+    });
+    return await agendaitem.save();
```
erikap: To be consistent with most of the codebase, I would replace with
```javascript
await agendaitem.save();
return agendaitem;
```

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      created: now,
+      agenda,
+      number: 1,
+      shortTitle: `Goedkeuring van het verslag van de vergadering van ${moment(
+        closestMeeting.plannedstart
+      ).format('dddd DD-MM-YYYY')}.`,
+      formallyOk: CONSTANTS.ACCEPTANCE_STATUSSES.NOT_YET_OK,
+      mandatees: [],
+      pieces: [],
+      approvals: [],
```
erikap: No need to initialize those empty relations

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      isApproval: true,
+      treatments: A([agendaItemTreatment]),
+    });
+    return await agendaitem.save();
+  }
+
+  @task({ drop: true })
```
erikap: Fyi, there is also `@dropTask` as an alias

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+  *createNewSession() {
```
erikap: Rename to `createMeeting`

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      plannedStart: startDate,
+      created: date,
+      kind: kindUriToAdd,
+      mainMeeting: this.selectedMainMeeting,
+      number: meetingNumber,
+      numberRepresentation: formattedMeetingIdentifier,
+    });
+
+    const closestMeeting =
+      yield this.agendaService.getClosestMeetingAndAgendaId(startDate);
```
erikap: This is the only place where `getClosestMeetingAndAgendaId(startDate)` is used in the codebase. It just does a custom call to the backend and is stateless. Can you convert it to a util (or maybe add to existing `meeting-utils.js`) and import that one instead of using the `agendaService`.  I also have a preference to name it `fetch...` instead of `get...` when it includes a call to the backend.

Same reasoning holds for `agendaService.assignNewSessionNumbers`.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      isDigital,
+      extraInfo,
+      selectedKindUri,
+      meetingNumber,
+      formattedMeetingIdentifier,
+    } = this;
+    const kindUriToAdd = selectedKindUri || CONFIG.MINISTERRAAD_TYPES.DEFAULT;
+    const date = this.formatter.formatDate(null);
+    const startDate = this.startDate || date;
+    const newMeeting = this.store.createRecord('meeting', {
```
erikap: Just `meeting` instead of `newMeeting`?

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+  @task({ drop: true })
+  *createNewSession() {
+    const {
+      isDigital,
+      extraInfo,
+      selectedKindUri,
+      meetingNumber,
+      formattedMeetingIdentifier,
+    } = this;
+    const kindUriToAdd = selectedKindUri || CONFIG.MINISTERRAAD_TYPES.DEFAULT;
```
erikap: Fallback not needed, since dropdown to select kind always has a value (and there is no option to clear)

app/components/meeting/new-meeting-modal.js
erikap: Can you rearrange the order of the methods such that `createMeeting` is above the 'helper' methods `createAgenda` and `createAgendaitemToApproveMinutes`?

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+  setKind(kind) {
+    this.selectedKindUri = kind;
+    if (!this.isAnnexMeeting) {
+      this.selectedMainMeeting = null;
+      this.generateNewNumber();
+    }
+  }
+
+  @action
+  meetingNumberChangedAction(meetingNumber) {
```
erikap: Use an active action names instead of passive one. E.g. `updateMeetingIdentifier`

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+    this.meetingNumber = meetingNumber;
+    this.formattedMeetingIdentifier = `VR PV ${this.currentYear}/${meetingNumber}`;
```
erikap: This formatted string is constructed at several locations in this component. It may be better to extract it to a separate method which sets the formatted identifier (a.o. based on `this.isAnnexMeeting`)

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  @action
+  editFormattedMeetingIdentifier() {
+    this.isEditingFormattedMeetingIdentifier = true;
+  }
+
+  @action
+  saveAction() {
```
erikap: Name is too generic. It's not clear what the action does. Action can maybe even be merged with the method from the previous comment. If a value is given as argument, it is assigned to `this.formattedMeetingIdentifier`; otherwise the formatted identifier is automatically constructed.

app/components/meeting/new-meeting-modal.js
```diff
@@ -0,0 +1,226 @@
...
+      isApproval: true,
+      treatments: A([agendaItemTreatment]),
+    });
+    return await agendaitem.save();
+  }
+
+  @task({ drop: true })
+  *createNewSession() {
+    const {
+      isDigital,
```
erikap: Doesn't seem to be used and is not defined on the `meeting` model. Can be removed.

erikap (changes requested):

erikap:
> Renames the component & refactors for Octane syntax. Replaces as many Vl-components to Auk-components as possible (Datepicker still TODO because `Auk::Datepicker` does not support selecting a time).

@brenner-company Staat het al op je roadmap om `Auk::Datepicker` uit te breiden met support voor time selection?

brenner-company:
@erikap Dit staat inderdaad al op de roadmap en zou volgende sprint kunnen opgenomen worden 👍

Momenteel ben ik binnen https://kanselarij.atlassian.net/browse/KAS-3182 bezig met het syncen van het design (Mono) met de datepickers. Ben daar quasi klaar mee (enkel time input heeft nog beetje werk nodig):

CleanShot 2022-03-04 at 10 02 42@2x

erikap:
:+1: Dan laat ik deze PR niet blokkeren op het gebruik van `WebComponents::VlDatepicker`. Er staat een TODO bij om in de toekomst te replacen met `Auk::Datepicker`.

sergiofenoll:
> This formatted string is constructed at several locations in this component. It may be better to extract it to a separate method which sets the formatted identifier (a.o. based on this.isAnnexMeeting)

> Name is too generic. It's not clear what the action does. Action can maybe even be merged with the method from the previous comment. If a value is given as argument, it is assigned to this.formattedMeetingIdentifier; otherwise the formatted identifier is automatically constructed.

@erikap resolved these two by wrapping `meetingNumber` and `formattedMeetingIdentifier` in getters & setters. I realize now that it may be a little confusing because it's hiding behaviour behind assignments, so if that's undesired let me know and I'll refactor it into a regular method and add calls where necessary.

erikap (changes requested):
The current refactor of `KindSelector` component is complex with the get/set construction and breaks the preselected value in the `EditSession` component. This can imo only be cleanly fixed by refactoring the `EditSession` component as well.
There is also a future story to refactor the `KindSelector` component to retrieve the codelist from the backend. So, for this PR, I would revert the `KindSelector` component changes such that this PR can pass.

erikap (approved):

PR #1230, opened 2022-03-08
KAS-3238 Make relation read-only on case side, reload case.subcase af…

ValenberghsSven:
# HOTFIX

## :warning: Branch started from PROD

- Made relation one-way in case by defining `serialize:false`
- Reload the list of subcases on case after adding a new one

Mainly the change in case.js wil prevent further bugs, the reloading is just to keep the data in sync.

More improvements in follow up PR

### DIFF
--- a/app/components/cases/new-subcase.js
+++ b/app/components/cases/new-subcase.js
@@ -130,6 +130,8 @@ export default class CasesNewSubcase extends Component {
     }
     // We save here in order to set the belongsTo relation between submission-activity and subcase
     await subcase.save();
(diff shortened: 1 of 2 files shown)


ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/166/
2 testen falen, maar die waren toen flaky (iets te maken met lagere ember versie + rdfa editor waardoor cypress daar niet mee kon werken)

PR #1234, opened 2022-03-10
Feature/kas 3201 upgrade to ember and ember data v3.28

sergiofenoll:
Upgrades dependencies to be compatible with v3.28.8 of Ember

Bumped versions based on the version generated by a brand new Ember 3.28 project, because the upgrade path using `ember-cli-update --to 3.28.8` didn't seem to actually change anything.

Removed typescript dependency which didn't seem necessary & caused issues while installing using `npm@8.3.0`, but this can be reverted if still necessary.

Removed `ember-ajax` dependency & disabled ajax support because I couldn't find usages anywhere (other than the fetch wrapper called ajax, but that does not use the dependency), but perhaps I'm wrong :stuck_out_tongue_closed_eyes:.

Tag of `ember-docker` is kept to 3.26.1 because newer versions use `npm@8.3.0` and we have mismatching peer dependencies between `@appuniversum/ember-appuniversum@^1.0.1` and `@lblod/ember-rdfa-editor@0.50.0-beta.4` and with newer npm version the install can't proceed. FWIW, bumping `@lblod/ember-rdfa-editor` to version 0.50.0 fixes the peer dependencies issue.

Fix for `account` model re-implemented here because v3.28 no longer works with `person` model. Presumably <v3.28 would cast the `user` model that we received from the backend into a `person` model (because we only access the name properties), but that no longer works in this version.

Fix for `mandatee-assinging.spec.js` test, where in one of the tests adding mandatees to a subcase (deeldossier), the new mandatees would not be saved. Fixed it by removing a call to `await model.get(key)` in the `setNewPropertiesToModel` function.
My guess as to why it breaks: instead of using a for-loop, the code originally used `forEach`, which means that the `await` keyword isn't actually respected. The call to `model.set(key, propertiesToSet[key])` would occur before the promise for `model.get(key)` had settled. Further guess, in previous versions of Ember Data, the promise of `model.get(key)` resolving would not overwrite the value contained in the model, but with v3.28 it does. Meaning that, the model we're saving has the old relationships. Explaining why it was impossible to add mandatees to a subcase.
This is just speculation though, I did not manage to find actual proof.

Fix for `subcase.spec.js`: replace a call to `.destroyRecord()` with a call to `.deleteRecord()` followed by `.save()`. I don't even have a guess as to why this one broke :pensive:.

There are probably more broken tests, so PR is still a draft (mostly so they get run on Jenkins instead of on my local PC :smile:)


package.json
```diff
@@ -65,36 +65,35 @@
...
-    "ember-cli-terser": "^4.0.1",
+    "ember-cli-terser": "^4.0.2",
     "ember-cli-update": "^1.0.0",
     "ember-click-outside": "^2.0.0",
     "ember-code-snippet": "^3.0.0",
     "ember-composable-helpers": "^4.4.1",
     "ember-concurrency": "^2.0.2",
     "ember-concurrency-decorators": "^2.0.3",
-    "ember-data": "~3.25.0",
+    "ember-data": "~3.28.6",
```
MikiDi: I think we really want to make sure we have 3.28.8 or up because of this https://github.com/emberjs/data/pull/7805#issuecomment-1028568379

ValenberghsSven:
"Fix for subcase.spec.js: replace a call to .destroyRecord() with a call to .deleteRecord() followed by .save(). I don't even have a guess as to why this one broke pensive."

Kinda want to see that one, might boot the stack locally

app/models/account.js
```diff
@@ -4,7 +4,7 @@ import { alias } from '@ember/object/computed';
 // TODO: octane-refactor
 // eslint-disable-next-line ember/no-classic-classes
 export default Model.extend({
-  user: belongsTo('person'),
```
ValenberghsSven: person model is a `person:Person`, while user is a `foaf:Person`.
person has a different first-name predicate `persoon:gebruikteVoornaam`.
Need to investigate what the consequences are from this small change, with all the mixed uses of user or person I can't judge it without looking deeper into it.

erikap: @ValenberghsSven No impact, since the relation is correctly defined in the backend models. It's just the frontend where the relationship is wrong. See also https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1216#discussion_r819454975

ValenberghsSven: I think the reason we did this is because the email and phone come from acmidm,  but if we use the `user` model, you technically override that data with every save..

ValenberghsSven: I completely missed this change/conversation in PR1216
Will be responding in that thread instead

sergiofenoll:
Seems that all tests pass, so I think this can be fully reviewed

PR #1237, opened 2022-03-11
Feature/kas 3243 refactor case subcases uses

ValenberghsSven:
Moved data loading of subcases from `cases/case/subcases/overview` to parent route `cases/case/subcases`
Was expecting some tests to break since the model of the parent route gets refreshed on subcase creation

You know when you think "I'm going to fix/change this little thing" and it turns out you refactored several components/routes to get it working again.
That. This PR. Anyway..

 ## What has changed

COMPONENTS
subcase-header.js:
- did not octane refactor (out of scope, should be done in KAS-3256)
- Added a method from args to call when the deletion or moving of a subcase is done, in order to refresh the parent route `cases.case.subcases` (see model change there)

subcase-item.js:`
- Removed some awaits from template (not all, too much side-tracking) by making `@subcase` sync
- Added `@case` to args, to avoid loading the relation here.
- Removed computed `nameToShow` from subcase.js model and used local code.
- Removed computed `onAgendaInfo`  from subcase.js model and used local code.
- changed `{{yield}}`, subcase was not needed here.

agendaitem-case/all-subcases => renamed to subcases/subcases-history:
- Renamed because logic, this component was used agendaitems a long time ago but not anymore. Also it could technically go anywhere as a generic componetn so did not put it under `subcase-case` folder for that reason.
- removed the awaits because of parent model changes
- added `@case` to args of `subcaseItem´

subcases-process:
- removed the awaits because of parent model changes
- added `@case` to args of `subcaseItem´
- `case.type` has been refactored a long time ago, nobody noticed this stopped working, fixed it to show the type of subcase by using the `showAsRemark` boolean from subcase.js model.

models/subcase.js:
- removed the inverse null from case relation, ember should now properly update the inverse relation (the inverse is read-only since hotfix). In some cases that means we can remove `case.hasMany(subcases).reload();` because ember will already know of the updates. But the hasMany reload also helps reload the relation with concurrency in mind so opted to keep it in.
- computed cleanup  note:I checked the uses of each coputed before and again in this PR 
- removed `nameToShow` and used local code
- removed `sortedMandatees`, did not seem used
- removed alias `latestMeeting`, removed other computeds that used it or just used `requestedForMeeting`
- removed `latestAgenda`, did not seem used
- removed `latestAgendaitem`, did not seem used
- removed `onAgendaInfo` and used local code
- removed `subcasesFromCase` and used proper route/subroute data loading

PODS:

cases/case/subcases/overview:
- renamed the refreshes (still confusing, because the flow is template > controller > route > parentRoute)
- moved the model of subcase to parent route `cases/case/subcases`, seemed logical
- added `@case` to args of `subcasesProcess`

cases/case/subcases:
- created route.js (only template existed)
- Added model from `cases/case/subcases/overview:` + comments from the computed `subcasesFromCase` from subcase.js

cases/case/subcases/subcase:
- octane refactor
- cleanup unused code regarding mandatees
- Added the methods for `cases/case/subcases` model refresh, started from `onMoveSubcase` on `SubcaseHeader`

cases/case/subcases/subcase/overview:
- Used data from route instead of computed `subcasesFromCase`  from subcase.js
- Renamed `allSubcases` because we had all subcases minus our own model. Named it `siblingSubcases`
- added `@case` to args of `subcasesHistory´
- `onCreateSubcase` in args of `SubcasesOverviewHeader` changed from refreshing local model to refreshing parent route `cases/case/subcases` model (or the new subcase wouldn't show)

nl-be.json:
- replace now unused translation by new one to replace harcoded text (`nameToShow` where there are no titles, this is possible to do since we don't enforce a short or long title to be present)


app/components/subcases/subcase-item.js
```diff
@@ -42,6 +65,11 @@ export default class SubcaseItemSubcasesComponent extends Component {
     this.hasDocumentsToShow = doc !== null;
   }
 
+  @task
+  *loadRelatedMeeting() {
+    this.requestedForMeeting = yield this.args.subcase.requestedForMeeting;
```
erikap: Waarom deze aparte load-task/tracked property en niet gewoon in de template laten resolven met `@subcase.requestedForMeeting`? Dit wordt toch maar op 1 plaats gebruikt?

erikap:
> PODS:
> 
> cases/case/subcases/overview:
> 
> * renamed the refreshes (still confusing, because the flow is template > controller > route > parentRoute)

As of Ember 4.0 the [router service provides a `refresh()`](https://api.emberjs.com/ember/4.0/classes/RouterService/methods/refresh?anchor=refresh) method to simplify these cases (see also [RFC631](https://emberjs.github.io/rfcs/0631-refresh-method-for-router-service.html)) and avoid usages of the magic `send()` method. There is a [polyfill](https://github.com/Windvis/ember-router-service-refresh-polyfill) available if we would like  start using it already, but it's probably best to keep that out of this PR.

erikap (commented):
Love this PR :star_struck: 

Just one question about the need for tracked property `requestedForMeeting`, but other than that, this PR is approved by me :+1:

app/components/subcases/subcase-item.js
ValenberghsSven: hmmm, true.
Ik heb daar een awaited computed vervangen en in die mindset heb ik een tracked gebruikt. Die tracked is op zich niet nodig, gewoon data loading doen bij constructie en dan in de template aanspreken.

ValenberghsSven: fixed it

ValenberghsSven:
> > PODS:
> > cases/case/subcases/overview:
> > 
> > * renamed the refreshes (still confusing, because the flow is template > controller > route > parentRoute)
> 
> As of Ember 4.0 the [router service provides a `refresh()`](https://api.emberjs.com/ember/4.0/classes/RouterService/methods/refresh?anchor=refresh) method to simplify these cases (see also [RFC631](https://emberjs.github.io/rfcs/0631-refresh-method-for-router-service.html)) and avoid usages of the magic `send()` method. There is a [polyfill](https://github.com/Windvis/ember-router-service-refresh-polyfill) available if we would like start using it already, but it's probably best to keep that out of this PR.

4.0 offers lots of goodies.
Not sure if we want to polyfill, it does remove a lot of magic and interaction between template/controller/route.
This does mean that you can refresh routes from anywhere (component/controller/route).

My first thought was "Is this against Seperation of Concerns?" with a mindset that `route.js` should be responsible for refreshing the route/bubbling up to parent route. But i'm not really engaged in the ember community so my ideas about responsibility are probably incorrect. It does make reloading parents a lot easier.

PR #1250, opened 2022-03-18
Kas 3257 treatments relation avoid many to many

ValenberghsSven:
Full story in https://kanselarij.atlassian.net/browse/KAS-3257

Short recap:
many-to-many in data is one-to-many in frontend meaning we have to thread carefully when loading relations and saving.
Specially when both sides can be saved/serialized.
In this case we would prefer to only serialize 1 side of this relation, just like between `agendaitem` and `pieces`.
Since we don't do that between `agendaitem` and `treatments` (maybe refactor later) we need to work around this issue.

## What has changed:
publication-case-info-panel & decisions-info-panel:
- Get the models needed for the linkTo via new method that takes the `treatment` as params.
- use the models in the `ButtonLink`
- Add the tracked property to the `#if` to avoid errors from `@models` was null while loading
- Limited the saving of `agendaItemTreatment`, we shouldn't save `via MR` in this case since nothing can be changed. Also to avoid the deletion of relations in case they are not loaded correctly as a safety.

agenda-item-treatment.js:
- added comment about incorrect relation in frontend and workaround.

publication-service:
- New method to query the agendaitem on the latest agenda that is connected to the treatment (op pubflow).
- Commented why it's needed.
- also load the relation to agenda and meeting so we can return the id's instead of the actual models.


erikap (approved):
Not super nice, but probably the best we can do at the moment without a lot of impact. The reasons are well documented in the code and in this PR :+1:  I will try to prioritize the story to fix the hasOne relationship to hasMany in the near future.

PR #1253, opened 2022-03-21
KAS-3267 use action instead of task, use a less complex way to set th…

ValenberghsSven:
Quick write-up, will go into details later.

Assuming ember upgrade did something different when working with deleted models.
Also assuming the reason it still works for agendaitem pieces restore is that it's an `@action` and not a `@task` because we do something very similar there and it still works (for now?)

New way of preparing to restore the previous version on delete is to load the previous report.
When actually deleting, we no longer have to use the deleted piece to fetch relations to previousPiece or documentContainer.
We use the previous version that we fetched earlier.

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision.js
+++ b/app/components/agenda/agendaitem/agendaitem-decision.js
@@ -84,6 +84,10 @@ export default class AgendaitemDecisionComponent extends Component {
     await piece.save();
     this.args.treatment.report = piece;
     await this.args.treatment.save();
+    // This reload is a workaround for file-service "deleteDocumentContainer" having a stale list of pieces
+    // when deleting the full container right after adding a new report version without the version history open.
+    const documentContainer = await piece.documentContainer;
+    await documentContainer.hasMany('pieces').reload();
     await this.loadReport.perform();
   }
 
--- a/app/components/agenda/agendaitem/agendaitem-decision.js
+++ b/app/components/agenda/agendaitem/agendaitem-decision.js
@@ -81,7 +81,6 @@ export default class AgendaitemDecisionComponent extends Component {
 
   @action
   async attachNewReportVersion(piece) {
-    await piece.previousPiece;
     await piece.save();
     this.args.treatment.set('report', piece);
     await this.args.treatment.save();
--- a/app/components/agenda/agendaitem/agendaitem-decision.js
+++ b/app/components/agenda/agendaitem/agendaitem-decision.js
@@ -3,7 +3,6 @@ import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency-decorators';
-import { sortPieces } from 'frontend-kaleidos/utils/documents';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 
 export default class AgendaitemDecisionComponent extends Component {
@@ -11,6 +10,7 @@ export default class AgendaitemDecisionComponent extends Component {
   @service store;
 
   @tracked report;
+  @tracked previousReport;
 
   @tracked isEditing = false;
   @tracked isVerifyingDelete = null;
@@ -56,6 +56,7 @@ export default class AgendaitemDecisionComponent extends Component {
   @task
   *loadReport() {
     this.report = yield this.args.treatment.report;
+    this.previousReport = yield this.report?.previousPiece;
   }
 
   @action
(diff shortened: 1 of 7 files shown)


app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -73,31 +74,34 @@ export default class AgendaitemDecisionComponent extends Component {
...
     this.isAddingReport = false;
     await this.loadReport.perform();
   }
 
-  @task
-  *attachNewReportVersion(piece) {
-    yield piece.save();
+  @action
+  async attachNewReportVersion(piece) {
+    await piece.previousPiece;
```
erikap: I don't understand why we have to fetch this relation here.

ValenberghsSven: I added this somewhere during testing, the `loadReport` task was not loading the `previousPiece` when it should have.
Changed some more things later that probably fixed it. This await is no longer needed, just tested it again without.

ValenberghsSven:
While analyzing this bug I had some weird situations.
While uploading a doc with ember inspector open I sometimes hit other errors, when looking them up it seemed to be caused by having ember inspector open (but not consistent)

After everything worked I tried to turn the 2 actions back into tasks but it just doesn't work the same way. Not sure why and didn't want to spend too much time on it.

ValenberghsSven:
This bug led to some weird situations.
While uploading a doc with ember inspector open I sometimes hit errors, when looking them up it seemed to be caused by having ember inspector open (but not consistent)

After everything worked, I tried to turn the 2 actions back into tasks but it just doesn't work the same way. Not sure why and didn't want to spend too much time on it.

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -73,31 +74,33 @@ export default class AgendaitemDecisionComponent extends Component {
...
+  // eslint-disable-next-line no-unused-vars
+  async attachPreviousReportVersion(piece) {
+    /*
+      the deleted piece comes in from params, but not used here because we tracked the previous version.
+      Note: this used to be a task but something kept going wrong.
+      Trying to get piece.documentContainer after piece was destroyed kept failing after an ember upgrade:
+      "Cannot create a new tag for '<(unknown):ember890>' after it has been destroyed"
+    */
+    if (this.previousReport) {
+      this.args.treatment.set('report', this.previousReport);
```
erikap: Since agenda-item-treatment data model has been converted to a native ES6 class, there is no need to use `set` anymore. Occurs a couple of times in this file. Not blocking for this PR.

erikap (approved):

app/components/agenda/agendaitem/agendaitem-decision.js
ValenberghsSven: Kinda glossed over that since I wasn't refactoring, getting rid of the `.set()` is something we should do when we find it.

ValenberghsSven:
Wait with merging, something is going wrong when the entire container is deleted.

There seems to be a difference when deleting a container with versionHistory open and versionHistory never opened.
In the second one, the `documentContainer.get('pieces')` in the `file-service` has a wrong count (1 piece instead of 2) when there is a BIS added immediatly without refreshing after it.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/bugfix/KAS-3267-decision-restore-previous-report/app/services/file-service.js#L47

Preferably I would like to improve the file-service later, so right now I have will have to add a `container.hasMany('pieces').reload()` when adding a new version of the report to counter this.
The use case of this happening on PROD is small, but we should address it.

ValenberghsSven:
> Wait with merging, something is going wrong when the entire container is deleted.
> 
> There seems to be a difference when deleting a container with versionHistory open and versionHistory never opened. In the second one, the `documentContainer.get('pieces')` in the `file-service` has a wrong count (1 piece instead of 2) when there is a BIS added immediatly without refreshing after it. https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/bugfix/KAS-3267-decision-restore-previous-report/app/services/file-service.js#L47
> 
> Preferably I would like to improve the file-service later, so right now I have will have to add a `container.hasMany('pieces').reload()` when adding a new version of the report to counter this. The use case of this happening on PROD is small, but we should address it.

Dug a little deeper.
The reason this was suddenly happening was because of the change in "loadReport". namely the line `this.previousReport = yield this.report?.previousPiece;`
The network stack makes it more clear:
On a agendaitem decision: upload document, upload BIS, delete document fully (never open history)
 - without the line (+ commented out the fix):
A GET is done to documentContainer/id/pieces, yielded 2 results.
 - with the line (+ commented out the fix):
This GET no longer happens, resulting in that relation not being reloaded properly.

Assuming the GET is something **ember** does automatically to keep up since we don't manually call that GET.
We do manually reload the pieces when opening history, that's why there is a difference with history closed/open
The reason that ember is not up to date is because we have `inverse: null` on the relation in piece.js:
```
@belongsTo('document-container', {
    inverse: null
  }) documentContainer;
```

When uploading a new piece in document-card, we only `createRecord('piece'` and pass on the documentContainer.
The inverse null means the documentContainer does not know a piece was added.
Unsure why ember was reloading the relation in some cases, but this is some "magic" we should control better since it seems easy to lose that magic.

ValenberghsSven:
We could try removing the inverse:null in a different ticket. This is sure to cause issues so not doing it here.
It is generally known that we have too many inverse:null in our code, we need to look at these after publication goes live

PR #1262, opened 2022-03-25
Cancelling publication documents upload was not working properly

ValenberghsSven:
Discovered in PR#1257 that the deletion of documents(file/container/piece) was not happening correctly in publication flow views.

One of the reasons was the approach in the separate components.
the `this.uploadedPieces.removeObject(piece);` was happening too fast, the list we were mapping changed during operation.
f.e. when you uploaded 5 documents and cancelled, we only hit the `deleteUploadedPiece` task twice instead of 5 times.
Fixed that by moving the `removeObject` after the actual deletes.

Even with this fix, sometimes large amounts of documents would fail on cancelling.
Deletes to piece were failing.
Didn't dug too deep into it. Suspecting that the random order of deletes due to the promise.all was causing issues.
From experience in this project, the correct order to delete the models is:
1. file
2. document-container
3. piece

Manually tested every publication modal that I changed with 10+ documents, deleting single ones with the trash icon in the modal and cancelling. No more errors occurred.
I did see that there is a time when all documents are uploaded but not loaded in the view. So you could technically hit save with only 3 out of 10 properly saved. Tested hitting cancel once and not every document got deleted. 
out of scope since this is partially caused by user behavior, not really a bug.

### DIFF
--- a/app/services/publication-service.js
+++ b/app/services/publication-service.js
@@ -380,10 +380,8 @@ export default class PublicationService extends Service {
   async deletePiece(piece) {
     const file = await piece.file;
     const documentContainer = await piece.documentContainer;
-    await Promise.all([
-      piece.destroyRecord(),
-      file.destroyRecord(),
-      documentContainer.destroyRecord(),
-    ]);
+    await file.destroyRecord();
+    await documentContainer.destroyRecord();
+    await piece.destroyRecord();
   }
 }
--- a/app/components/publications/publication/proofs/proof-request-modal.js
+++ b/app/components/publications/publication/proofs/proof-request-modal.js
@@ -148,8 +148,8 @@ export default class PublicationsPublicationProofsProofRequestModalComponent ext
 
   @task
   *deleteUploadedPiece(piece) {
-    this.uploadedPieces.removeObject(piece);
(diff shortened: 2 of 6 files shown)


app/services/publication-service.js
```diff
@@ -380,10 +380,8 @@ export default class PublicationService extends Service {
...
     const file = await piece.file;
     const documentContainer = await piece.documentContainer;
-    await Promise.all([
-      piece.destroyRecord(),
-      file.destroyRecord(),
-      documentContainer.destroyRecord(),
-    ]);
+    await file.destroyRecord();
+    await documentContainer.destroyRecord();
+    await piece.destroyRecord();
```
erikap, line 385: I assume `file` and `documentContainer` can still be executed in parallel since they don't have any direct connection to each other. But since you tested it with this flow and we're sure that one works, it's ok for me to merge the PR like this. It will only be a marginal gain.

erikap (approved):

app/services/publication-service.js
ValenberghsSven: Small gains maybe, but I do assume that cancelling an upload of many documents is not going to be done often in day to day work.

PR #1263, opened 2022-03-25
KAS-3286 Opladen van extra bestanden onder 'Besluiten' tab voor publicaties via MR

sergiofenoll:
Allow uploads of extra pieces to publications via MR

Only enable delete of pieces via MR if they don't have an access-level
relation.


erikap (approved):

ValenberghsSven (approved):
Works, not sure if any tests will fail on this so wait for those.

PR #1268, opened 2022-03-28
KAS-3291 status-change read-only, persist the inverse belongsTo

ValenberghsSven:
[Jira ticket](https://kanselarij.atlassian.net/browse/KAS-3291)

- Set the relation between publicatie-flow and publication-status-change only one-way while still allowing read access
- When replacing the current status change we need to reload the relation, to ensure we are trying to delete an existing status-change since the current one could have gone stale.

### DIFF
--- a/app/components/publications/editable-publication-status-pill.js
+++ b/app/components/publications/editable-publication-status-pill.js
@@ -131,20 +131,20 @@ export default class PublicationStatusPill extends Component {
       this.decision = undefined;
     }
 
-    // update status-change activity
-    const oldChangeActivity = yield this.args.publicationFlow
-      .publicationStatusChange;
-    if (oldChangeActivity) {
-      yield oldChangeActivity.destroyRecord();
-    }
-    const newChangeActivity = this.store.createRecord(
+    // update publication-status-change
+    // reload the relation for possible concurrency
+    const currentStatusChange = yield this.args.publicationFlow
+      .belongsTo('publicationStatusChange')
+      .reload();
+    yield currentStatusChange?.destroyRecord();
+    const newStatusChange = this.store.createRecord(
       'publication-status-change',
       {
         startedAt: date,
         publication: this.args.publicationFlow,
       }
     );
-    yield newChangeActivity.save();
+    yield newStatusChange.save();
 
     yield this.args.publicationFlow.save();
     this.loadStatus.perform();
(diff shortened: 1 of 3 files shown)


erikap:
I understand the motivation of this PR - you can reproduce this issue consistently - but wonder where we draw the line then. If the existing status-change-activity is outdated, that basically means the status is outdated as well. Should we reload that resource too? If you extend this reasoning consistently, we will land on a server-side rendered app.

My assumption is that this problem is not going to occur in practice for OVRB. For publications, the same people are rarely working on the same publication at the same time. Unlike the secretarial staff who do spend an entire week working on the same agenda. 

I think, as @MikiDi already mentioned a couple of times, we should start looking at long-term solutions instead of keep adding bandaid: 
- only serialize dirty attributes/relations
- get rid of `inverse: null` flags on relations
- maybe (re)evaluate the usage of async ember-data relationships
- push updates from backend to frontend

ValenberghsSven:
_" If the existing status-change-activity is outdated, that basically means the status is outdated as well. Should we reload that resource too? If you extend this reasoning consistently, we will land on a server-side rendered app"_

Yes, this and any other relation that was updated is classed under "use-case 2" of my analysis.
When 2 people work on the same publication-flow, we are going to overwrite data.
In most cases, this is probably fine and may result in some data loss depending who saved what first.
The reasoning for this PR is that **when** it happens, there is no error halfway through our process because of it. 
That would lead to more issues than just overwritten data.
The reason I don't refresh status is because we don't really need to in this PR, overwriting stale status will work since we are not attempting to destroy models that no longer exist like we do with the status-change.

_"My assumption is that this problem is not going to occur in practice for OVRB"._
It can, and therefore it will at some point. But I also think it will be rare.
That's why "use-case 2" is lower prior right now and we have to inform the users to not work on a browser tab that has been open for a while without a refresh.

Changing `inverse:null` is not relevant for concurrency in this issue as it only helps to keep your local models up to date with changes you do.

_"only serialize dirty attributes/relations"_ 
Most likely the way to, but even then you can overwrite properties/relations and lose data...

erikap:
To be clear, it's ok for me to merge this PR now, but we need to shift focus in the future.

erikap (approved):

ValenberghsSven:
> To be clear, it's ok for me to merge this PR now, but we need to shift focus in the future.

I know, just wanted to clarify my reasoning a bit.
I would like to avoid using a band aid but there isn't really a solution readily available.
Unless I push the deletion of the old status-change to a backend service.
I was initially thinking of always using a query for getting the `status-change`, but seemed difficult because we need it in the overview.
maybe one or both of those might make a cleaner solution than having to reload before delete

PR #1282, opened 2022-04-05
Feature/kas 3294 interne opmerking enkel zichtbaar voor secretarie toevoegen aan een agendapunt

sergiofenoll:
Backend PR: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/265
Yggdrasil PR: https://github.com/kanselarij-vlaanderen/yggdrasil/pull/32

Adds a privateComment to agendaitems and pulls in some visual changes from the usertest branch.

Some formatter changes accidentally snuck for one of the files, sorry!


app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles.hbs
```diff
@@ -100,17 +100,23 @@
       </div>
     </Auk::Panel::Body>
     {{#if @agendaitem.explanation }}
```
erikap: Trailing whitespace

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles.hbs
```diff
@@ -100,17 +100,23 @@
...
-      </Auk::Modal::Footer>
+      <Auk::Panel::Footer>
+        <p>
+          <strong class="auk-u-text-bold">
+            {{t "remark-title"}}:
+          </strong>{{@agendaitem.explanation}}
+        </p>
+      </Auk::Panel::Footer>
+    {{/if}}
+    {{#if @agendaitem.privateComment }}
```
erikap: Trailing whitespace

app/models/agendaitem.js
```diff
@@ -34,8 +34,8 @@ export default ModelWithModifier.extend({
   title: attr('string'),
   formallyOk: attr('string'),
   isApproval: attr('boolean'), // isGoedkeuringVanDeNotulen
-  explanation: attr('string'),
-  // More information: https://github.com/kanselarij-vlaanderen/kaleidos-frontend/pull/469.
+  explanation: attr('string'), // More information: https://github.com/kanselarij-vlaanderen/kaleidos-frontend/pull/469
```
erikap: I know it doesn't originate from this PR, but can you remove the link to GitHub. No idea what the meaning of that comment is :woman_shrugging:

erikap (changes requested):
Some small comments.
Backend PR has already been merged, so custom Jenkins build is no longer need for new commits.

sergiofenoll:
Should I add tests for this? Perhaps a simple one to make sure that the private comment is there & is editable?
I'm not sure if Yggie runs in the test environment though, so I don't know if we can test the (lack of) propagation behaviour.

erikap:
> Should I add tests for this? Perhaps a simple one to make sure that the private comment is there & is editable? I'm not sure if Yggie runs in the test environment though, so I don't know if we can test the (lack of) propagation behaviour.

There are already some tests on Yggdrasil. Some of them seem to be failing now, but I'm not sure whether it's caused by the changes in this PR. 

If we want to add the private comment to a test (which is probably a good idea), we will first need a new release of Yggdrasil and a feature-branch on app-kaleidos containing the bumped version.

ValenberghsSven:
Some tests were failing on meeting-kind and agenda-approve-service, this wasn't up to date with dev branch.

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles.hbs
```diff
@@ -99,18 +99,24 @@
...
-        </Auk::Toolbar>
-      </Auk::Modal::Footer>
+    {{#if @agendaitem.explanation}}
+      <Auk::Panel::Footer>
+        <p>
+          <strong class="auk-u-text-bold">
+            {{t "remark-title"}}:
+          </strong>{{@agendaitem.explanation}}
+        </p>
+      </Auk::Panel::Footer>
```
ValenberghsSven: test-selector is gone: `data-test-agendaitem-titles-explanation`.
A test is failing because of it.
Currently testing locally to see if just restoring it (on the `p` element) will be enough.

app/components/agenda/agenda-overview/agenda-overview-item.hbs
```diff
@@ -104,39 +113,62 @@
...
+          <p>
+            <strong class="auk-u-text-bold">
+              {{t "remark-title"}}:
+            </strong>
+            {{@agendaitem.explanation}}
+          </p>
+        {{/if}}
         <div class="auk-hr">
-          <hr/>
+          <hr />
```
ValenberghsSven, line 125: The explanation is now above the `hr` as seen on the usertest (not in notion link) in agenda overview items
Looks a little funny to me.
example with documents.
image
But if this is desired then carry on.

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles-edit.hbs
```diff
@@ -45,6 +45,18 @@
       />
     </div>
   </div>
+  <div class="auk-u-mb-4">
+    <div class="auk-form-group">
+      <Auk::Label id="private-comment-title" class="auk-u-icon-left"><Auk::Icon @name="lock-closed" @size="small" /> {{t "private-comment-title"}}</Auk::Label>
```
ValenberghsSven: new line for `auk:icon`?
Tried prettier locally, doesn't do a good job for this one.
Manually:
```
<Auk::Label id="private-comment-title" class="auk-u-icon-left">
    <Auk::Icon @name="lock-closed" @size="small" />
    {{t "private-comment-title"}}
</Auk::Label>
```

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles-edit.hbs
```diff
@@ -45,6 +45,18 @@
...
+      <Textarea
+        data-test-agendaitem-titles-edit-private-explanation={{true}}
```
ValenberghsSven: You don't really have to add test selectors if you're not using them in this PR.
It will just result in unused and untracked selectors (selector files in cypress).
Opt to remove any new ones from this PR.

ValenberghsSven (changes requested):
Small comments, rest is ok

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles.hbs
ValenberghsSven: Hmm no, the behavior of this field changed.
Before this, we did not trim whitespace on view. Strings started with a blank enter would show as such.
In this PR, the removal of the `pre` with classes means we do trim whitespace when viewing.
When editing, the whitespaces are still there.
Not sure if this is desired

If it is, then change this line
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/0f71e99ad7d42ffcc064596d7fc3b7219eee0614/cypress/integration/unit/agenda.spec.js#L167

to 
```
cy.get(agenda.agendaitemTitlesView.explanation).contains(`Opmerking: ${explanation}`);
```

ValenberghsSven:
As for cypress tests:
I would recommend testing locally (manually) if other profiles can see the value.
I will add it to the todo list for cypress (higher prior since it's clear what needs to be tested)

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles.hbs
sergiofenoll: Ah, I didn't think about the changes to the whitespace. Just blindly took over the `<p>` from the usertest branch, which explains why the test keeps failing locally even though I re-added the test selector. Thanks for the catch!

app/components/agenda/agenda-overview/agenda-overview-item.hbs
sergiofenoll: Agreed that it looks a little funky, but I took over the styling from the usertest branch. See: https://usertest.kanselarij-dev.s.redpencil.io/vergadering/6225FB46EB3C02DA50165888/agenda/6225FB47EB3C02DA50165889/agendapunten

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles-edit.hbs
sergiofenoll: Updated the agenda.spec.js test to check the private comment alongside the normal comment, so left this in

sergiofenoll:
Have some failing tests locally which didn't fail on Jenkins. I'm guessing it's because the delta notifier takes too long to send out deltas (because some of the services aren't running locally), will see what it gives on Jenkins.

About adding new tests: `agenda.spec.js` got a little addition to add a private comment & then check that it gets rendered.
@ValenberghsSven I can try adding some checks in the propagatie tests, but if you think it'd best be handled by you/Mathias (would probably be quicker for you to do it), I will leave it like this.

ValenberghsSven:
> Have some failing tests locally which didn't fail on Jenkins. I'm guessing it's because the delta notifier takes too long to send out deltas (because some of the services aren't running locally), will see what it gives on Jenkins.
> 
> About adding new tests: `agenda.spec.js` got a little addition to add a private comment & then check that it gets rendered. @ValenberghsSven I can try adding some checks in the propagatie tests, but if you think it'd best be handled by you/Mathias (would probably be quicker for you to do it), I will leave it like this.

The agenda spec changes look good, unsure about the whitespace but jenkins will show.
The propagation would be the place to put the profile tests, but I'll plan a follow-up ticket after the easter holiday so we can limit the testing work done in this PR.

ValenberghsSven:
"Have some failing tests locally which didn't fail on Jenkins. I'm guessing it's because the delta notifier takes too long to send out deltas (because some of the services aren't running locally), will see what it gives on Jenkins."

Good guess! Deltanotifier taking ages is a known issue when the delta receiving services are not running (yggdrasil, musearch, others..).
You could locally turn off those delta rules.
What we should do is replace the service by a mocked delta service, a service that just receives and discards them to prevent deltanotifier from being a slowpoke.
The cache not invalidating fast enough is usually the consequence of this.

app/components/agenda/agendaitem/agendaitem-case/agendaitem-titles-edit.hbs
ValenberghsSven: I reviewed before any test was changed so this is good now.

app/components/agenda/agenda-overview/agenda-overview-item.hbs
ValenberghsSven: All the same for me, if usertest is the desired view then we leave it like this.

sergiofenoll:
> "Have some failing tests locally which didn't fail on Jenkins. I'm guessing it's because the delta notifier takes too long to send out deltas (because some of the services aren't running locally), will see what it gives on Jenkins."
> 
> Good guess! Deltanotifier taking ages is a known issue when the delta receiving services are not running (yggdrasil, musearch, others..). You could locally turn off those delta rules. What we should do is replace the service by a mocked delta service, a service that just receives and discards them to prevent deltanotifier from being a slowpoke. The cache not invalidating fast enough is usually the consequence of this.

Maybe instead of setting the entrypoint of the disabled services to `echo 'service disabled'`, we could set it to a netcat port listener that always returns a 200/204. Something along the lines of `while true; do echo -ne "HTTP/1.0 200 OK\r\n\r\n" | nc -lp 9000 ; done`. But that does depend on all those containers having netcat installed, which maybe isn't the case.

ValenberghsSven (approved):

erikap:
Awaiting release of https://github.com/kanselarij-vlaanderen/yggdrasil/pull/32 before merging this PR.

erikap (approved):

PR #1289, opened 2022-04-08
KAS-2305 Remove session-number-service

sergiofenoll:
Backend PR: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/273

After this gets merged, service should get archived on Github. https://github.com/kanselarij-vlaanderen/session-number-service


ValenberghsSven (approved):
Does this make session-service obsolete?
Was never a fan of the name because of the actual user session <> meeting (called session in early days)

sergiofenoll:
> Does this make session-service obsolete? Was never a fan of the name because of the actual user session <> meeting (called session in early days)

Yep! The JIRA Story was to either refactor it to a JSON:API compliant service, or archive the repo and remove it as a dependency if it's no longer needed. Given the changes to agendas and agenda creations, this service is no longer needed.

PR #1293, opened 2022-04-14
Feature/kas 3305 vervangen van transition to route met router service in controllers

sergiofenoll:
also applied octane refactors to the affected controllers/routes/templates except for agendas/overview because this one should get deleted if #1288 gets merged

### DIFF
--- a/app/pods/newsletters/template.hbs
+++ b/app/pods/newsletters/template.hbs
@@ -1,5 +1,3 @@
-{{!-- TODO: octane-refactor --}}
-{{!-- template-lint-disable no-action --}}
 {{page-title (t "newsletter")}}
 <Newsletter::NewsletterHeader />
 {{#if (gt this.model.length 0)}}
@@ -8,7 +6,7 @@
       <div class="auk-u-m-8">
         <DataTable
           data-test-route-newsletters-data-table={{true}}
-          @content={{this.model}}
+          @content={{@model}}
           @class="auk-table auk-table--striped"
           @page={{this.page}}
           @size={{this.size}}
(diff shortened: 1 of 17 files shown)


app/pods/newsletters/template.hbs
```diff
@@ -8,15 +6,15 @@
       <div class="auk-u-m-8">
         <DataTable
           data-test-route-newsletters-data-table={{true}}
-          @content={{this.model}}
+          @content={{@model}}
```
erikap, line 9: Did you get a warning for this one? I think we use `this.model` in most places atm. This kind of change occurs a couple of times in this PR.

sergiofenoll: Not a warning, it's just [stated in the ember docs](https://guides.emberjs.com/release/routing/specifying-a-routes-model/) to use `@model` in the template and `this.model` in the controller, which is why I tend to change it, though both do seem to work.

Although `@model` does seem to be used in some pods, most of the app uses `this.model`, so I can revert those changes for consistency.

erikap: No, no problem if `@model` is preferred by the official docs. Thanks for the pointer :pray:

erikap (approved):
Changes ok. Awaiting Jenkins run before merging.

app/pods/settings/users/index/route.js
```diff
@@ -23,14 +22,11 @@ export default Route.extend(DataTableRouteMixin, {
...
-    refresh() {
-      this._super(...arguments);
-      this.refresh();
-    },
-  },
-});
+  @action
+  refreshRoute() {
+    this.super(...arguments);
+    this.refresh();
```
erikap, line 29: Falende test lijkt mij met deze aanpassing te maken te hebben. Kan je hier `super.refresh()` van maken?

sergiofenoll: Issue lag bij de calls naar `this.send('refreshRoute')` vanuit de controller. Ik dacht dat die `.send()` een methode was om Route methodes op te roepen, maar dat is dus om `action`s op te roepen, zowel uit de route als de controller. Aangezien de controller ook een `refreshRoute` actie had kwamen er recursie-related errors tevoorschijn.

PR #1296, opened 2022-04-15
KAS-3307: Combine {Edit,New}MeetingModal into one component

sergiofenoll:
Combines the two modal components into a single component which takes a meeting record and enables a user to change its properties/relationships as was previously possible in the `NewMeetingModal` component.

The caller is responsible for passing in either a "real" meeting record, i.e. of a meeting that has been fetched from the backend, or a "new" meeting record, i.e. an object instantiated with `store.createRecord('meeting')`. The component makes use of the record's `isNew` member to determine whether it should perform the extra steps for a meeting creation when saving.

In theory this behaviour could be extracted to the caller as well (i.e., the `agendas.overview` controller creates the agendaitem for the meeting once it's been created), leaving this component purely as a meeting editor. Maybe that's a cleaner solution.

Other changes in this PR:

- The modal changes can't be saved unless all necessary fields are filled in.
- The `numberRepresentation` input field can be emptied (but not saved in this sate)
- The `meetingNumber` input field can be emptied (but not saved in this state)

The last two are to prevent behaviour where emptying the input caused it to be reset to the default value, which looks weird.


app/components/meeting/edit-meeting-modal.hbs
```diff
@@ -1,61 +1,130 @@
 <Auk::Modal @size="large">
-  <Auk::Modal::Header @title={{t "edit-meeting"}} @onClose={{@onCancel}} />
+  <Auk::Modal::Header
+    @title={{if @meeting.isNew (t "new-agenda") (t "edit-meeting")}}
+    @closeAble={{true}}
```
erikap: `closeable` instead of `closeAble`? But since `true` is the default it can be omitted.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -2,39 +2,60 @@ import Component from '@glimmer/component';
...
+  @service store;
+  @service newsletterService;
   @service toaster;
 
+  @tracked isAnnexMeeting = false;
+  @tracked isEditingNumberRepresentation = false;
+
+  @tracked selectedMainMeeting;
+  @tracked isFinal;
   @tracked kind;
```
erikap: Rename to `selectedKind` to be in line with `selectedMainMeeting`?

app/components/meeting/edit-meeting-modal.js
```diff
@@ -47,40 +68,175 @@ export default class MeetingEditMeetingComponent extends Component {
...
+  }
+
   @action
   selectStartDate(val) {
     this.startDate = val;
   }
 
   @action
-  setKind(kind) {
+  async setKind(kind) {
```
erikap: Rename to `selectKind` to be in line with `selectMainMeeting`?

app/pods/agendas/controller.js
```diff
@@ -20,11 +22,14 @@ export default class AgendasController extends Controller.extend(DefaultQueryPar
...
   closeNewSessionModal() {
     this.isCreatingNewSession = false;
+    this.newMeeting.deleteRecord();
+    this.newMeeting = this.store.createRecord('meeting');
   }
 
   @action
   successfullyAdded() {
     this.isCreatingNewSession = false;
+    this.newMeeting = this.store.createRecord('meeting');
```
erikap: Maybe better to move to `setupController` in the route since we want to reinitialize this each time the user visits the route?

app/pods/agendas/controller.js
```diff
@@ -7,10 +7,12 @@ import { action } from '@ember/object';
 import DefaultQueryParamsMixin from 'ember-data-table/mixins/default-query-params';
 
 export default class AgendasController extends Controller.extend(DefaultQueryParamsMixin) {
+  @service store;
   @service currentSession;
   @service router;
 
   @tracked isCreatingNewSession = false;
+  @tracked newMeeting = this.store.createRecord('meeting');
```
erikap: Maybe better to move to `setupController` in the route since we want to reinitialize this each time the user visits the route?

app/components/meeting/edit-meeting-modal.hbs
```diff
@@ -1,61 +1,130 @@
...
-        />
+    {{else}}
+      <div class="auk-u-mb-2">
+        <div class="auk-form-group" data-test-edit-meeting-datepicker>
+          <Auk::Label>{{t "start-date"}}</Auk::Label>
+          {{! TODO: Replace this with Auk::Datepicker, which isn't fully compliant yet (e.g. no time support) }}
+          <WebComponents::VlDatepicker
+            @enableTime={{true}}
+            @date={{array this.startDate}}
+            @dateChanged={{this.selectStartDate}}
```
erikap: Using inline `set` like [here](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/18b3af8f59030c17dcc3d117c4aa3fa344ff3009/app/components/publications/new-publication-modal.hbs#L109)

app/components/meeting/edit-meeting-modal.js
```diff
@@ -47,40 +68,175 @@ export default class MeetingEditMeetingComponent extends Component {
...
+        this.initializeKind.isRunning ||
+        this.initializeMainMeeting.isRunning ||
+        this.saveMeeting.isRunning
+      );
+  }
+
+  @task
+    *initializeMainMeeting() {
+      this.selectedMainMeeting = yield this.args.meeting.mainMeeting;
+    }
```
erikap: Indentation

app/components/meeting/edit-meeting-modal.js
```diff
@@ -47,40 +68,175 @@ export default class MeetingEditMeetingComponent extends Component {
...
+
   @task
   *initializeKind() {
     this.kind = yield this.args.meeting.kind;
+    this.kind ??= yield this.store.findRecordByUri('concept', CONSTANTS.MEETING_KINDS.MINISTERRAAD);
```
erikap: Didn't know this `??=` syntax. This is the same as `this.kind = (yield this.args.meeting.kind) || (yield this.store.findRecordByUri('concept', CONSTANTS.MEETING_KINDS.MINISTERRAAD);`, right? (just to be sure I understood it correctly)

app/components/meeting/edit-meeting-modal.js
```diff
@@ -2,39 +2,60 @@ import Component from '@glimmer/component';
...
+    this.initializeMeetingNumber.perform();
+    this.initializeMainMeeting.perform();
 
-    this.meetingYear = this.args.meeting.plannedStart.getFullYear();
-    this.startDate = this.args.meeting.plannedStart;
+    this.meetingYear = this.args.meeting.plannedStart?.getFullYear() || this.currentYear;
+    this.startDate = this.args.meeting.plannedStart || new Date(now.getFullYear(), now.getMonth(), now.getDate(), 10, 0, 0);
     this.extraInfo = this.args.meeting.extraInfo;
-    this.meetingNumber = this.args.meeting.number;
+    this.isFinal = this.args.isFinal ?? false;
```
erikap: `this.isFinal` seems not to be used. Argument `isFinal` is never passed to the component.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -47,40 +68,175 @@ export default class MeetingEditMeetingComponent extends Component {
...
+    const isNewMeeting = this.args.meeting.isNew;
     const now = new Date();
 
     this.args.meeting.extraInfo = this.extraInfo;
     this.args.meeting.plannedStart = this.startDate || now;
     this.args.meeting.kind = this.kind;
     this.args.meeting.number = this.meetingNumber;
     this.args.meeting.numberRepresentation = this.numberRepresentation;
+    this.args.meeting.mainMeeting = this.selectedMainMeeting;
+    this.args.meeting.isFinal = false;
```
erikap: A bit weird to overwrite this flag while editing the meeting. Was it like this before?

erikap (changes requested):
Since behaviour greatly differs for new/existing meeting, I have a preference for moving the creation of the agenda/first agenda item to the `agendas.overview` controller as you suggest.

app/components/meeting/edit-meeting-modal.js
sergiofenoll: yes, or more specifically `(yield this.args.meeting.kind) ?? (yield this.store...` since it only checks for null/undefined instead of all false-y values. but writing it like that results in `this.kind` being `null` when creating a new meeting. I suspect it's because the evaluation of `||` (or `??`) happens against the promises instead of the values themselves. this results in the first promise being accepted as non-false-y (or non-nullish), but once it resolves it turns out to be `null` after all. that's why i split it up in two statements, so that the promises are settled as expected.

app/components/meeting/edit-meeting-modal.js
sergiofenoll: was meant to use `this.isFinal` here, thanks for catching that!

ValenberghsSven:
1 of the failing tests seems legit in decision spec. We are creating PVV agenda right after the main agenda.
I'm assuming we end up editing the first agenda to be PVV rather then creating a new PVV agenda.
need to confirm manually.

sergiofenoll:
> 1 of the failing tests seems legit in decision spec. We are creating PVV agenda right after the main agenda. I'm assuming we end up editing the first agenda to be PVV rather then creating a new PVV agenda. need to confirm manually.

Yup, you're right. Currently checking it out. It's because after saving the meeting, the `setupController` hook doesn't get called again, so when we create another meeting we're in fact just editing the previous meeting.

In checking this, I also noticed that the title of the modal changes from "Nieuwe agenda aanmaken" to "Vergadering wijzigen", because the record gets saved. Will make the check on the title more explicit as well, with an `isNew` argument

app/components/meeting/edit-meeting-modal.js
erikap: Still confused about the `this.isFinal`. Since `this.args.isFinal` is always `undefined` (because not passed to the component) `this.isFinal` will always be `false`. The property is nowhere manipulated; only used to set `this.args.meeting.isFinal = this.isFinal` when saving the meeting. Is the tracked property needed? And does it need to be set to `false` on each save of the meeting? I don't see how it relates to the saved changes.

Behaviour I would expect is `isFinal` to be `false` on creation of a new meeting (that can be covered by adding the property on `createRecord('meeting', ...)` and leaving it untouched on edit of the meeting? Or am I missing something here @ValenberghsSven ?

app/pods/agendas/controller.js
```diff
@@ -20,12 +27,86 @@ export default class AgendasController extends Controller.extend(DefaultQueryPar
...
+        agenda,
+        this.newMeeting,
+        closestMeeting
+      );
+    }
+    await this.newsletterService.createNewsItemForMeeting(this.newMeeting);
+    this.newMeeting = this.store.createRecord('meeting');
+    // TODO: Should fix sessionNrBug
+    // Import from meeting-utils.js
+    // await assignNewSessionNumbers();
```
erikap: Is this comment still applicable @ValenberghsSven (and if yes, what does it entail) or can it be removed?

app/pods/agendas/controller.js
```diff
@@ -20,12 +27,86 @@ export default class AgendasController extends Controller.extend(DefaultQueryPar
...
+    this.newMeeting = this.store.createRecord('meeting');
   }
 
   @action
-  successfullyAdded() {
+  async successfullyAdded() {
+    const agenda = await this.createAgenda(this.newMeeting);
+
+    const closestMeeting = await fetchClosestMeetingAndAgendaId(this.newMeeting.plannedStart);
+    const broader = await this.newMeeting.kind?.get('broader');
```
erikap: Try to avoid usage of `get()` but explicitly await promises (even if you now it will immediately resolve because the data has already been fetched)
```javascript
const kind = await this.newMeeting.kind;
const broader = await kind?.broader;
```

app/components/meeting/edit-meeting-modal.js
sergiofenoll: That line should be `this.isFinal = this.args.meeting.isFinal ?? false`, so it follows the expected behaviour, I just missed the `.meeting`

app/pods/agendas/controller.js
```diff
@@ -20,12 +27,86 @@ export default class AgendasController extends Controller.extend(DefaultQueryPar
   @action
   closeNewSessionModal() {
     this.isCreatingNewSession = false;
+    this.newMeeting.deleteRecord();
+    this.newMeeting = this.store.createRecord('meeting');
   }
 
   @action
-  successfullyAdded() {
+  async successfullyAdded() {
```
erikap: `createAgendaAndNewsletter` is maybe a better name to cover the content?

app/pods/agendas/template.hbs
```diff
@@ -26,7 +26,9 @@
 </Auk::Navbar>
 {{outlet}}
 {{#if this.isCreatingNewSession}}
-  <Meeting::NewMeetingModal
+  <Meeting::EditMeetingModal
+    @isNew={{true}}
```
erikap: Maybe better to embed this logic in `Meeting::EditMeetingModal` by adding a tracked property `isNew` in the meeting modal component which is initialized in its constructor based on `this.args.meeting.isNew`. Keeps the component interface simpler.

app/pods/agendas/controller.js
ValenberghsSven: Not sure what that bug is or was.
It seems that the relevant code in the session-number-service would delete the triple and reinsert a decimal `adms:identifier` (`meeting.number` in frontend). Knowing what we know now, decimal is not a correct type since we should use integer. 
So the comment can go, but we might have to look into cleaning up the type differences (and maybe find the bug) in new ticket?

app/components/meeting/edit-meeting-modal.js
```diff
@@ -47,40 +67,99 @@ export default class MeetingEditMeetingComponent extends Component {
...
     const now = new Date();
 
     this.args.meeting.extraInfo = this.extraInfo;
     this.args.meeting.plannedStart = this.startDate || now;
-    this.args.meeting.kind = this.kind;
+    this.args.meeting.kind = this.selectedKind;
     this.args.meeting.number = this.meetingNumber;
     this.args.meeting.numberRepresentation = this.numberRepresentation;
+    this.args.meeting.mainMeeting = this.selectedMainMeeting;
+    this.args.meeting.isFinal = this.isFinal;
```
ValenberghsSven: The usage of "isFinal" in this component is mainly for applying the default of `isFinal:false` to a new meeting I guess?
Could we apply the defaults to the creation of the model from args instead?

Just want to prevent us accidentally overwriting the isFinal value + you can't actually change this value in this component

app/pods/agendas/controller.js
```diff
@@ -20,12 +27,87 @@ export default class AgendasController extends Controller.extend(DefaultQueryPar
   @action
   closeNewSessionModal() {
     this.isCreatingNewSession = false;
+    this.newMeeting.deleteRecord();
+    this.newMeeting = this.store.createRecord('meeting');
```
ValenberghsSven: To me, it doesn't really make sense to create something on closing of a modal.
also, doing it this way is resulting in us having 3 places where we create a meeting (here, in `createAgendaAndNewsletter` and in the route's `setupController`)
Could we just create the meeting (with defaults, see other comments) on the opening of the modal `openNewSessionModal`? Should be the only place

app/pods/agendas/controller.js
ValenberghsSven: See other comment, do we need to initialize this when visiting route?
We could just create a model for meeting before opening the `editMeetingModal`.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -2,39 +2,59 @@ import Component from '@glimmer/component';
...
   @tracked startDate;
   @tracked extraInfo;
   @tracked _meetingNumber;
   @tracked _numberRepresentation;
 
+  currentYear = new Date().getFullYear();
+
   constructor() {
     super(...arguments);
+    this.isNew = this.args.meeting.isNew;
```
ValenberghsSven, line 33: Instead of relying on the ember model to be new, we could pass `@isNew={{true}}` when creating a new one. We just have to default to false for editing

ValenberghsSven (changes requested):
Template `new-meeting-modal` still exists, guess we can delete?

app/components/meeting/edit-meeting-modal.js
sergiofenoll: @erikap [asked for the opposite change](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1296#discussion_r856082386) (i.e. don't pass `isNew`, compute it in the component)

I see the value in both, but don't have a strong preference for one or the other.
Ik zou zeggen, vecht het tussen elkaar uit :stuck_out_tongue_winking_eye:

ValenberghsSven: goh, ik maak niet zo vaak gebruik van de ember model properties buiten `hasDirtyAttributes`.
Ik bekijk het eerder als volgt:
Bij creatie van een meeting is dit true.
Bij edit van een meeting is dit false.
In dat opzicht zullen opties beiden werken. Voor mij is het gelijk.
Voor de ember properties moet je weten dat ze bestaan en wat ze juist doen om ze correct te gebruiken.

ValenberghsSven: Hou het maar zoals het is.

ValenberghsSven (approved):

PR #1301, opened 2022-04-21
remove unused server-side provided roles

MikiDi:
Currently the only roles we use are defined client-side https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/0b4dc8c7c2a976cbd7bbdec36d2a2e616ef695b9/app/services/current-session.js#L54-L56 .

Remove the unused server-side defined ones.


ValenberghsSven (approved):
gefaalde newsletter testen negeren

PR #1302, opened 2022-04-21
Feature/kas 3310 refactor data loading bij detail van een agendapunt

sergiofenoll:
AgendaitemNav is now part of the agendaitem template. Data needed for said nav is loaded in the afterModel hook and set on the controller.


app/pods/agenda/agendaitems/agendaitem/route.js
```diff
@@ -14,17 +15,23 @@ export default class AgendaitemAgendaitemsAgendaRoute extends Route {
...
-  afterModel(model, transition) {
+  async afterModel(model, transition) {
     const arrayToSearch = model.showAsRemark ? this.modelFor('agenda.agendaitems').announcements : this.modelFor('agenda.agendaitems').notas;
     if (!arrayToSearch.includes(model) && arrayToSearch.length) { // This can happen when the selected item no longer is visible in the sidebar after filtering
       this.transitionTo('agenda.agendaitems.agendaitem', arrayToSearch[0]);
     }
     this.transition = transition; // set on the route for use in setupController, since the provided "transition" argument there always comes back "undefined"
+
+    this.treatments = await model.treatments;
+    this.newsletterInfo = await this.treatments.firstObject.newsletterInfo;
```
ValenberghsSven: `treatments` can be null (1000 records on prod), possibly resulting in a 'cannot get newsletterInfo of undefined' error.
`this.treatments?.firstObject?.newsletterInfo;`

In the current app, having 0 treatments shouldn't happen since we don't allow for deletion of the last one.
You can use 2 browsers with 2 treatments and delete 1 of them in both browsers to end up with 0.
From that point on, the treatment view would break.

ValenberghsSven (changes requested):
Behaviour changed after refactor:
Open agenda with 2 agendaitems, 1 approval and 1 normal one.
Open the "kort bestek" tab with agendaitem 2
in the detail tab, open the approval agendaitem.
On DEV: we default to the "dossier" tab, since approval should not have newsletter-info
In this PR: We stay on "kort bestek" tab

sergiofenoll:
> Behaviour changed after refactor:
> Open agenda with 2 agendaitems, 1 approval and 1 normal one.
> Open the "kort bestek" tab with agendaitem 2
> in the detail tab, open the approval agendaitem.
> On DEV: we default to the "dossier" tab, since approval should not have newsletter-info
> In this PR: We stay on "kort bestek" tab

This behaviour is also present on the development branch, so not caused by changes in this PR, probably only came up now because dev got merged back into this branch. I can try to fix it, but I imagine it'll broaden the scope of this PR quite a bit.

ValenberghsSven:
> > Behaviour changed after refactor:
> > Open agenda with 2 agendaitems, 1 approval and 1 normal one.
> > Open the "kort bestek" tab with agendaitem 2
> > in the detail tab, open the approval agendaitem.
> > On DEV: we default to the "dossier" tab, since approval should not have newsletter-info
> > In this PR: We stay on "kort bestek" tab
> 
> This behaviour is also present on the development branch, so not caused by changes in this PR, probably only came up now because dev got merged back into this branch. I can try to fix it, but I imagine it'll broaden the scope of this PR quite a bit.

Ah, I see what happened.
I opened an agendaitem on DEV that had 0 treatments (probably from old publication bug)
So I never actually had the "kort bestek" tab open when I tested.

Ignore this review comment then, should be fine after adding the `?`

ValenberghsSven (approved):

ValenberghsSven:
This change will not break test and tests were green before so we can merge

PR #1307, opened 2022-04-25
Feature/kas 3327 paginering op overzicht procedurestappen

sergiofenoll:
Adds pagination to subcases & reduces the number of backend calls made.


app/components/subcases/subcase-item.js
```diff
@@ -89,6 +95,22 @@ export default class SubcaseItemSubcasesComponent extends Component {
     this.subcaseDocuments = sortPieces(pieces);
   }
 
+  @task
+  *loadSubcasePhases() {
+    this.phases = yield this.subcasesService.getSubcasePhases(this.args.subcase);
```
sergiofenoll, line 100: This loading task replaces a call to [`subcase.phases`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/d9b7831ae7494bf5e052d8a62b02fe82c64cb699/app/models/subcase.js#L57).

app/components/subcases/subcase-item.js
```diff
@@ -89,6 +95,27 @@ export default class SubcaseItemSubcasesComponent extends Component {
...
+  }
+
+  @task
+  *loadSubcaseIsApproved() {
```
sergiofenoll, line 104: This loading task replaces a call to [`subcase.approved`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/d9b7831ae7494bf5e052d8a62b02fe82c64cb699/app/models/subcase.js#L79)

app/pods/cases/case/subcases/overview/controller.js
```diff
@@ -5,6 +5,8 @@ import { action } from '@ember/object';
 
 export default class CasesCaseSubcasesOverviewController extends Controller {
   @tracked case;
+  @tracked page = 0;
+  @tracked size = 25;
```
sergiofenoll, line 12: The JIRA story's description specifies that we should show 20 subcases. I was a bit lazy and set it to 25 (because that's the closest value that's provided by `Auk::Pagination` by default). If showing 20 subcases instead of 25 is desired, this can be changed of course.

app/services/subcases-service.js
```diff
@@ -54,6 +54,10 @@ export default class SubcasesService extends Service {
             date: geplandeStart,
           });
           if (phaseData.decisionResultId) {
+            // Why don't we load in this data in the custom-subcases-service?
```
sergiofenoll: Maybe this shouldn't actually be a comment in the code, but just wrote it down somewhere when it occurred to me :sweat_smile:

app/pods/cases/case/subcases/overview/controller.js
tdn: Doesn't really matter, as long as they can change it. 25 is fine for me

app/services/subcases-service.js
tdn: Good question... The performance already seems quite OK as it is though. So let's leave this in there for now.

tdn:
Works well and looks good to me. If Jenkins passes this'll be good to merge as far as I can see.

package.json
```diff
@@ -116,6 +116,7 @@
     "ember-qunit": "^5.1.5",
     "ember-resolver": "^8.0.3",
     "ember-responsive": "^3.0.6",
+    "ember-router-service-refresh-polyfill": "^1.0.1",
```
ValenberghsSven, line 119: sneaky :D, was expecting this in a seperate PR.
Guess we'll just go for it now and plan refactor stories later.

app/services/subcases-service.js
ValenberghsSven: Instead of a question in comments, you could create a Jira ticket and mention it here
`// KAS-xxxx load in this data in the custom-subcases-service instead`

Part of the audit last year also suggested to not have "TO DO" or question comments in code unless there is linked ticket available so we should try to avoid adding them.

app/pods/cases/case/subcases/subcase/overview/route.js
```diff
@@ -9,7 +9,16 @@ export default class CasesCaseSubcasesSubcaseOverviewRoute extends Route {
...
+    //  We want to sort descending on date the subcase was concluded.
+    //  In practice, reverse sorting on created will be close
+    const allSubcases = await this.store.query('subcase', {
+      filter: {
+        case: {
+          id: this.case.id,
+        },
+      },
+      sort: '-created',
+    });
```
ValenberghsSven, line 21: I get that the pagination was not compatible with the current setup where a parent model from route `cases/case/subcase` is passed down to this afterModel.
Querying these here will impact the performance of the subcase detail overview?
use-case:
Case with 10 subcases, I want to view all of the info
before PR: loaded all subcases once when loading the parent route, opening detail view 10 times by going back to the overview and opening another subcase just passes on that model.
after PR: We query subcases on the overview route, we open the first subcase and we need to query all subcases again. We go back to the overview and have to query all subcases, opening second subcase is another query to all subcases. and so on.

Am I seeing this wrong?
If I'm not, perhaps it would be a good idea to keep the call in the parent route to avoid doing it multiple times.
The queries will be cached after the first time, but cache invalidations happen often on active cases/subcases.

app/pods/cases/case/subcases/subcase/controller.js
```diff
@@ -1,9 +1,12 @@
...
 import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
 
 export default class CasesCaseSubcasesSubcaseController extends Controller {
+  @service router;
+
   @action
   refreshSubcases() {
-    this.send('refreshParentModel');
+    this.router.refresh('cases.case.subcases');
```
ValenberghsSven: this only made sense when the subcases were in the model of `cases.case.subcases`, if we keep the current approach, you would just need to refresh the current route `cases.case.subcases.subcase`

app/components/subcases/subcase-item.js
```diff
@@ -89,6 +95,27 @@ export default class SubcaseItemSubcasesComponent extends Component {
...
+        'filter[subcase][id]': this.args.subcase.id,
+        'filter[decision-result-code][:id:]': [
+          CONSTANTS.DECISION_RESULT_CODE_IDS.GOEDGEKEURD,
+          CONSTANTS.DECISION_RESULT_CODE_IDS.KENNISNAME,
+        ].join(','),
+      }));
+    } else {
+      this.approved = false;
+    }
+  }
```
ValenberghsSven: This aims to replace the computed property `approved` on subcase.js, which i'm fine with.
We use it in more places, so eventually this could maybe move to a util instead, good for now atleast.

The computed `phases` on subcase was listening to the `approved` computed, which is now being called less so I found an instance where the computed `phases` was not updated (refresh of page fixed it)
Have a subcase on a design agenda (only 1 or 2 phases, "ingediend op" en "geagendeerd op")
in the same browser window:
- look at the subcase detail
- go to agenda, set formal ok and approve and close the agenda
- go back to subcase detail, the phases should now have "goedgekeurd op", but they don't, only after refresh.

Possible solution:
add `this.phases = yield this.subcasesService.getSubcasePhases(this.args.subcase);` to the afterModel hook of `cases/case/subcases/subcase/` and pass on that args to `SubcaseDescriptionView`, or load it that component I guess.
Then the `phases` computed should no longer be referenced.

app/pods/cases/case/subcases/overview/controller.js
```diff
@@ -13,6 +15,24 @@ export default class CasesCaseSubcasesOverviewController extends Controller {
 
   @action
   refreshSubcases() {
-    this.send('refreshParentModel');
+    this.send('refreshSubcasesRoute');
+  }
```
ValenberghsSven: why send when we can router with the polyfill? :D

ValenberghsSven (changes requested):

ValenberghsSven:
Managed to fix tests for now but still waiting on CSS, could we copy some of the needed classes into auk_additions like we did in other PR's?

app/pods/cases/case/subcases/subcase/overview/route.js
sergiofenoll: The problem with keeping it in the parent route is that we force the `subcase` rout to only show whatever page we're on in the `subcases` route. Additionally, in a follow-up story we add pagination to the `subcase` route too, which should be separate from the pagination in the `subcases` route.

I don't know what the real life use case is like though, if there will be tons of `cases` with >25 `subcases`, so I don't know how often pagination is actually going to be used here.

package.json
sergiofenoll: Had to choose between adding it in now, or reviving `cases/case/subcases/subcase/route.js` just to have some kind of refresh action (and then also make sure it doesn't have name collissions, make sure it's being called correctly from the route, ...) so I thought just adding it was the easiest solution :P

app/pods/cases/case/subcases/overview/controller.js
sergiofenoll: hah, good catch!

app/components/subcases/subcase-item.js
sergiofenoll: The changes in this PR shouldn't change the behaviour you described :thinking: 
The detail view of a subcase `cases.case.subcases.subcase.overview` is still using the computed `phases` & `approved` properties of the `subcase` model. So, unless I'm mistaken, in your example you're going form subcase overview to agenda back to subcase overview, so whatever is happening has nothing to do with the changes here (because those happen in `cases.case.subcases.overview`, and you're never passing that page.

sergiofenoll:
> Managed to fix tests for now but still waiting on CSS, could we copy some of the needed classes into auk_additions like we did in other PR's?

Is this comment for a different PR? No CSS changes were made in this PR :thinking:

app/components/subcases/subcase-item.js
ValenberghsSven: hmm, had to verify on dev branch and seems like you are right.
The problem is already present, my bad.

Could do the fix in a bug ticket since I think that fix will work, but ignore here

ValenberghsSven:
> > Managed to fix tests for now but still waiting on CSS, could we copy some of the needed classes into auk_additions like we did in other PR's?
> 
> Is this comment for a different PR? No CSS changes were made in this PR thinking

yep it is....  ignore this.

ValenberghsSven:
I was going to comment something, started querying for some data to answer the question "I don't know what the real life use case is like though, if there will be tons of cases with >25 subcases, so I don't know how often pagination is actually going to be used here."

But it seems like I found anomalies in the data.
I tried querying to get a count of cases who had more than 20 subcases.
The data surprised me, I found number as high as 381!.
After digging deeper, something seems wrong with legacy data.
the number of 381 subcases is the case from the Jira ticket (which should only have 122?). I queried all cases, then their subcases (with the relation `dossier:doorloopt`)
Only 122 have a type `dossier:Procedurestap`, so the other 259 are some legacy conversion error.

Talked with Johan, subcases were incorrectly linked to 1 case in the original conversion.

It's out of scope but just wanted to indicate what I just spend some time on to find out ...

ValenberghsSven:
FYI, adjusted my query and from what I can tell:
Without looking at legacy data, the current highest amount of subcases in a case is 8 since the start of kaleidos.

app/pods/cases/case/subcases/subcase/overview/route.js
ValenberghsSven: I'll keep the possible performance hit in mind going forward.
Now that I know this is mainly a legacy data issue I guess querying the same thing a few times will be fine in terms of performance on average cases.

ValenberghsSven (approved):
fine as it is now if jenkins is green.

edit: 1 test is failing on dev "Vertrouwelijk" in a pill not found, ignore here

app/pods/cases/case/subcases/overview/template.hbs
```diff
@@ -6,11 +6,41 @@
...
+            @nbOfItems={{@model.length}}
+            @total={{@model.meta.count}}
+            @onNextPage={{this.nextPage}}
+            @onPreviousPage={{this.prevPage}}
+          />
+        </Group.Item>
+      </Toolbar.Group>
+      <Toolbar.Group @position="right" as |Group|>
+        <Group.Item>
+          {{t "amount-showed"}}
```
ValenberghsSven: forgot about this, "dossiers per pagina" => "procedurestappen per pagina"?
Might be confusing

ValenberghsSven, line 35: forgot about this, "dossiers per pagina" => "procedurestappen per pagina"?

sergiofenoll: thanks! made it "Aantal resultaten per pagina", same thing we use in agendas overview

app/config/constants.js
```diff
@@ -55,6 +55,10 @@ export default {
     KENNISNAME: 'http://kanselarij.vo.data.gift/id/concept/beslissings-resultaat-codes/9f342a88-9485-4a83-87d9-245ed4b504bf',
     INGETROKKEN: 'http://kanselarij.vo.data.gift/id/concept/beslissings-resultaat-codes/453a36e8-6fbd-45d3-b800-ec96e59f273b',
   },
+  DECISION_RESULT_CODE_IDS: {
+    GOEDGEKEURD: '56312c4b-9d2a-4735-b0b1-2ff14bb524fd',
+    KENNISNAME: '9f342a88-9485-4a83-87d9-245ed4b504bf',
+  },
```
erikap: Sorry, late to the review party, but I think this is a dangerous precedent to end up again in an unstructured config file with a mix of ids and URIs. We only consider URIs to be stable. 

I get that you need IDs to filter on multiple values in mu-cl-resources, but I would configure URIs here and get their IDs in the component where you need them. By using [`store.findRecordByUri`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/store.js#L29) we ensure only 1 backend call is made per concept so performance impact is minimal.

ValenberghsSven: hmm, kinda passed over this but we should try to avoid going back to ID's indeed.
We can filter on uri in the query in `subcase-item.js` (we did that in the computed in subcase.js)
```
 'filter[decision-result-code][:uri:]': [
          CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD,
          CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME,
        ].join(','),
```

ValenberghsSven: @sergiofenoll already made the change to start jenkins. Perhaps we can get this PR done for ACC release

sergiofenoll: mu-cl-resources support zo'n "or" niet voor URIs, maar ik push zo een change waar we de IDs gewoon binnenhalen

ValenberghsSven: "mu-cl-resources support zo'n "or" niet voor URIs" echt? Ik twijfelde even, had het beter eerst lokaal getest...

sergiofenoll: ja... ik weet dat ook alleen omdat ik het gisteren in rocket chat had gevraagd (heb ook eerst zo'n query proberen te schrijven, maar dat gaat dus niet)
https://chat.semte.ch/channel/mu-semtech?msg=2SQi8azvmuDs3NLie
maar hopelijk wordt support ervoor at some point toegevoegd :triumph:

ValenberghsSven: dat heb je met snel willen zijn op het einde van de dag :sweat_smile:  sorry..

PR #1309, opened 2022-04-25
Feature/kas 3354 paginering bij getoonde procedurestappen in zelfde deeldossier

sergiofenoll:
Adds pagination to subcases in subcase/overview route.

Draft because I'm making some changes here which I don't love and I'd rather discuss, and it's probably easier to check the changes via PR than by looking at a branch.


app/pods/cases/case/subcases/subcase/template.hbs
```diff
@@ -11,8 +11,4 @@
     @onMoveSubcase={{this.refreshSubcases}}
   />
 </div>
-<div class="auk-scroll-wrapper__body">
```
sergiofenoll, line 14: The wrapper has been moved to the subroutes of `cases.case.subcases.subcase` because in the overview, we have the pagination at the bottom which we want to render outside of the wrapper. I assume it would be possible to let the pagination element break out of the wrapper instead of moving it to the subroute, but that seemed far too hackish to me, so I went for this solution.

app/pods/cases/case/subcases/subcase/overview/route.js
```diff
@@ -17,22 +27,35 @@ export default class CasesCaseSubcasesSubcaseOverviewRoute extends Route {
...
+      },
       sort: '-created',
     });
-    this.siblingSubcases = allSubcases.filter(
-      (subcase) => subcase.id !== model.id
-    );
-    this.mandatees = (await model.mandatees).sortBy('priority');
-    this.submitter = await model.requestedBy;
+    const siblingSubcases = allSubcases.filter((sibling) => sibling.id !== subcase.id);
+    this.siblingSubcasesCount = allSubcases.meta.count;
```
sergiofenoll, line 37: When you query the backend, Ember keeps the `meta` property from the response in the resulting object. Normally we use this property, specifically `meta.count` to display the total amount of results.

When calling `.filter()` here, the result we get is a plain JS array, so we lose the count information and we have to store that "next" to the actual array. Ideally at some point it just becomes possible to query the backend and filter out the result we don't want so we don't need to do any filtering in the frontend.

sergiofenoll: Apart from this, when doing pagination here we will show one the page that contains the current subcase with one less item than the selected page size. This might be confusing to the users (e.g. because the size is set to 25, but we're only showing 24 items).

ValenberghsSven: since the subcases are not numbered, one less might not even be noticeable. Not to worried about this.

Ideally we should be able filter out certain id's in the `store.query`, which would solve this problem.

ValenberghsSven:
assuming this needs updating with the base branch that had some fixes?

app/pods/cases/case/subcases/subcase/overview/route.js
ValenberghsSven: Another option would be to show something special to indicate which of the subcases is the one being looked at instead of just hiding it.

ValenberghsSven (commented):
The current changes look ok (but needs a fix from the base branch to work properly), jenkins failed on startup so waiting with approving

ValenberghsSven (approved):

PR #1323, opened 2022-05-05
Feature/kas 3342 datum mandatees selector

TaneeG:
Jira: https://kanselarij.atlassian.net/browse/KAS-3342

Note: Mandatees panel was also used for Case and publication Flow so that's why they're also edited


app/components/utils/mandatee-selector.js
```diff
@@ -33,26 +40,32 @@ export default class MandateeSelector extends Component {
...
     const queryOptions = {
-      ...this.defaultQueryOptions,  // clone
+      ...this.defaultQueryOptions, // clone
     };
     if (searchTerm) {
       queryOptions['filter[person][last-name]'] = searchTerm;
     }
-    queryOptions['filter[government-body][:uri:]'] = CURRENT_GOVERNMENT_BODY;
+    queryOptions['filter[:lte:start]'] =
+      this.governmentBodyOfDate.toISOString();
```
tdn: **Possibly** this could create an issue in some edge cases where the meeting is on the last/first day of a new government composition if the Date locale is not the same. (That's why I suggested using Date.UTC()). 

However, that'll be easily remedied by KAS-3344 when they can select **all** mandataries, so let's leave this like that.

tdn (approved):
Tested, working well, and code looks understandable and logical.

app/components/mandatees/mandatees-panel.js
```diff
@@ -9,6 +9,7 @@ export default class MandateesMandateesPanelComponent extends Component {
    * @argument submitter
    * @argument allowEditing
    * @argument onSave
+   * @GovernmentBodyOfDate Date of Case, Agenda or PublicationFlow to find the Mandatees of the right date
```
erikap: Incorrect documentation syntax

app/components/mandatees/mandatees-panel/mandatees-panel-edit.js
```diff
@@ -10,6 +10,7 @@ export default class MandateesMandateesPanelEditComponent extends Component {
    * @argument submitter
    * @argument onCancel
    * @argument onSave
+   * @GovernmentBodyOfDate Date of Case, Agenda or PublicationFlow to find the Mandatees of the right date
```
erikap: Incorrect documentation syntax

app/components/mandatees/mandatees-selector-modal.js
```diff
@@ -6,6 +6,7 @@ import { task } from 'ember-concurrency';
  * @callback {() => Promise} onClose
  * @callback {(mandatee: Mandatee) => Promise} onLink
  * @dependsOn {Mandatee[]} mandatees ('mandatee,mandatee.person')
+ * @GovernmentBodyOfDate Date of Case, Agenda or PublicationFlow to find the Mandatees of the right date
```
erikap: Incorrect documentation syntax

app/components/publications/publication/case/mandatees/mandatees-panel.hbs
```diff
@@ -65,6 +65,7 @@
 
 {{#if this.showSelectMandateeModal}}
   <Mandatees::MandateesSelectorModal
+    @governmentBodyOfDate={{@publicationFlow.created}}
```
erikap: For publications via MR this should be the date of the decision (i.e. start date of the decision-activity)
For publications not via MR `publicationFlow.openingDate` is semantically probably the best guess.

app/components/publications/publication/case/mandatees/mandatees-panel.js
```diff
@@ -4,6 +4,7 @@ import { tracked } from '@glimmer/tracking';
 
 /**
  * @argument {PublicationFlow} publicationFlow (publication-flow,publication-flow.mandatees,publication-flow.mandatees.person)
+ * @GovernmentBodyOfDate Date of PublicationFlow to find the Mandatees of the right date
```
erikap: Incorrect documentation syntax

app/pods/agenda/agendaitems/agendaitem/index/controller.js
```diff
@@ -25,6 +25,13 @@ export default class IndexAgendaitemAgendaitemsAgendaController extends Controll
 
   @tracked isEditingAgendaItemTitles = false;
 
+  get agendaCreatedDate() {
+    if (this.meeting.isPreKaleidos) {
+      return this.meeting.plannedStart;
+    }
+    return this.agenda.created;
+  }
```
erikap: Should always be `meeting.plannedStart` imo. Agenda creation date is a technical date which has no semantic meaning regarding government-bodies/mandatees.

app/pods/cases/case/subcases/subcase/overview/controller.js
```diff
@@ -16,6 +16,14 @@ export default class CasesCaseSubcasesSubcaseOverviewController extends Controll
   @tracked governmentAreas;
   @tracked siblingSubcasesCount;
 
+  get caseCreatedDate() {
+    if (this.case.created) {
+      return this.case.created;
+    } else {
+      return this.model.subcase.created;
+    }
+  }
```
erikap: As stated in a previous comment creation date is a technical date which has no semantic meaning regarding government-bodies/mandatees. In case the subcase is related to a meeting (`subcase.requestedForMeeting`)  the planned-start date of the meeting is the best guess.

Otherwise  subcase creation date is probably the best fallback we have. But shouldn't subcase always get preference over case creation date then?

Naming of the getter may also be improved to cover the load.

app/components/utils/mandatee-selector.js
```diff
@@ -2,19 +2,25 @@ import Component from '@glimmer/component';
...
-  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03e0' // Minister
+  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03e0', // Minister
 ];
-
+/**
+ * @argument singleSelect
+ * @argument readOnly
+ * @argument selectedMandatee
+ * @argument chooseMandatee
+ * @GovernmentBodyOfDate Date of Case, Agenda or PublicationFlow to find the Mandatees of the right date
```
erikap: Incorrect documentation syntax

erikap: Argument name is confusing since it's used as filter on mandatee start/end date and not on government-body. Maybe something like `dateFilter` (and `mandateeDateFilter` in wrapping components)? 
Description can also be simplified to something like 'Date to get active mandatees for'. Doesn't matter where that date originates from in this component.

app/components/utils/mandatee-selector.js
```diff
@@ -33,26 +40,32 @@ export default class MandateeSelector extends Component {
...
-    // Many versions of a mandatee exist within a government-body.
-    // We only want the mandatees with no end-date or an end-date in the future.
-    // mu-cl-resources doesn't have :has-no:-capability for properties.
-    return results.filter((mandatee) => {
-      return !mandatee.end || (mandatee.end > new Date());
-    });
+    if (results.length === 0) {
+      results = yield this.loadCurrentBodyMandatees(searchTerm);
+    }
+    return results;
```
erikap: The condition of this fallback is weird. Results may for example not be found because the user is searching for a mandatee name that wasn't active in the government at that time. It's unexpected that we start hinting currently active mandatees in that case.

To get the expected behaviour, a query needs to be done before the mandatee-query to get the current active government-body (regeringssamenstelling) based on the date passed as argument. The URI of that government-body should be added to the next filter for mandatees (as was done in the previous implementation with the hardcoded `CURRENT_GOVERNMENT_BODY`).

As stated in the comment that has been removed `mu-cl-resources doesn't have :has-no:-capability for properties.` That filter flag is [only applicable for relationships](https://github.com/mu-semtech/mu-cl-resources#special-filters). The filtering on end-date should therefore be done in the frontend like before.

To query for the goverment-body at the right level (regeringssamenstelling) it's probably worth reading [the data model documentation about the different levels](https://themis.vlaanderen.be/docs/catalogs#ministers).

app/components/utils/mandatee-selector.js
```diff
@@ -2,19 +2,25 @@ import Component from '@glimmer/component';
...
 import { tracked } from '@glimmer/tracking';
-import { timeout, task, restartableTask } from 'ember-concurrency';
+import { restartableTask, task, timeout } from 'ember-concurrency';
 import { inject as service } from '@ember/service';
-import { CURRENT_GOVERNMENT_BODY } from 'frontend-kaleidos/config/config';
 
 const VISIBLE_ROLES = [
   'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03de', // Minister-president
-  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03e0' // Minister
+  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03e0', // Minister
```
erikap, line 10: This list will not be sufficient when searching for mandatees from before 06/1995. Roles at that time were 'Voorzitter' and 'Gemeenschapsminister'.

erikap (changes requested):

app/components/utils/mandatee-selector.js
tdn: This one was done like this because I specified in the ticket to get rid of the `CURRENT_GOVERNMENT_BODY` constant, since it was a hardcoded solution.

About `:has-no:`: there's a whole thread in the #mu-semtech channel on this. (NHtxEJ9j5SjeBnwvW)

Basically this is [translated ](https://github.com/mu-semtech/mu-cl-resources/blob/master/framework/response-generation.lisp#L226-L231) to `FILTER(NOT EXISTS { ?mandataris mandaat:einde ?mandaatEinde })` instead of `FILTER(!BOUND(?mandaatEinde))`. But I think the two are equivalent, no?

In any case, these two queries return the same result, so I'm quite sure this'll keep on working:

```
PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
PREFIX prov: <http://www.w3.org/ns/prov#>
PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

SELECT DISTINCT ?bestuursOrgaan ?mandataris ?mandaatStart ?mandaatEinde WHERE {
  ?bestuursOrgaan a besluit:Bestuursorgaan;
	prov:hadMember ?mandataris .
  ?mandataris mandaat:start ?mandaatStart .
  OPTIONAL { ?mandataris mandaat:einde ?mandaatEinde . }
  FILTER (!BOUND(?mandaatEinde))
}
```

and

```
PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
PREFIX prov: <http://www.w3.org/ns/prov#>
PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>

SELECT DISTINCT ?bestuursOrgaan ?mandataris ?mandaatStart ?mandaatEinde WHERE {
  ?bestuursOrgaan a besluit:Bestuursorgaan;
	prov:hadMember ?mandataris .
  ?mandataris mandaat:start ?mandaatStart .
  OPTIONAL { ?mandataris mandaat:einde ?mandaatEinde . }
  FILTER (NOT EXISTS { ?mandataris mandaat:einde  ?mandaatEinde })
}
```

Given how much of a refactor/hassle this is to implement without `: has-no:`, I think it's worth the 'risk' here.

tdn: > The condition of this fallback is weird. Results may for example not be found because the user is searching for a mandatee name that wasn't active in the government at that time. It's unexpected that we start hinting currently active mandatees in that case.

This is true. I would suggest to do this without constants though, by making the call to get the current government body first, using the current Date. Then we can check whether the reference date is earlier or later than the start date of the mandatees in that government composition.

If earlier: do the call using the reference date.
If later: use the current government body we just got.

This might even be more efficient, since the current composition will be used in 99% of the cases.

app/components/utils/mandatee-selector.js
tdn: Small thought about this:

- What about vice-minister-president and vice-voorzitter? Shouldn't they be visible as well?

- Also, if this is not parameterized & basically all roles are currently shown: what's the point of having this filtering?

tdn: Also, in KAS-3344 you'll see that this component will be upgraded to have an option to show **all** mandatees, including those not in a `besluit:Bestuursorgaan`. So in that case the filtering will have to be bypassed as well.

erikap: I assume the filters on role are there because a vice-minister can never be assigned to an agendaitem in Kaleidos. They always act in their role of minister.
Just adding the URIs of 'Voorzitter' and 'Gemeenschapsminister' should fix it. And, as you mention, in a next story the users will have an option to bypass the filter.

app/components/utils/mandatee-selector.js
erikap: As @madnificent mentioned in the referenced chat discussion there is a risk that `:has-no:` to filter on properties will break randomly in a minor release since it's unspecified behaviour. Because the filtering logic in mu-cl-resources is actively being developed on at the moment, I do see a substantial risk. If we decide to take that risk, we should at least document it inline and provide a test case that fails if the filter doesn't work anymore such that the breaking change becomes apparent.

About the logic to get the mandatees, the logical steps in my head are:
1. Get the government body based on the reference date  (or current date if none is passed)
2. Get the mandatees filtered by:
    - the government body from step 1
    - mandate start date < reference date
    - mandate end date > reference date or undefined (preferably in frontend)
    - roles

We're missing mandatees not linked to a government composition in that case, but I'm wondering whether those should be selectable?  When checking production data using the SPARQL query below, it only yields non-Themis mandatees as a result. Aren't these planned to be phased out/mapped to a Themis mandatee? To ensure data quality it seems acceptable to me that they cannot select those mandatees in the dropdown. Any opinion on this @MikiDi ? 
```sparql
SELECT DISTINCT ?s WHERE {
    ?s a <http://data.vlaanderen.be/ns/mandaat#Mandataris>  .    
    FILTER NOT EXISTS { ?b <http://www.w3.org/ns/prov#hadMember> ?s }
}
```

app/components/mandatees/mandatees-panel.js
```diff
@@ -9,6 +9,7 @@ export default class MandateesMandateesPanelComponent extends Component {
    * @argument submitter
    * @argument allowEditing
    * @argument onSave
+   * @argument {ReferenceDate} Date to get active Mandatees for
```
tdn: Is there a reason/convention for the curly braces here? Just wondering, since I don't know which is the preferred documentation method in the app (I see many different styles).

The one with the braces I see elsewhere in the app has the following structure:

```
@argument {TypeOfArgument} nameOfArgument: comments
```

So I would prefer to see this as:

```
@argument {Date} referenceDate: Date to get active Mandatees for
```

Same applies to the other comments below. 
Tnx!

app/components/mandatees/mandatees-panel/mandatees-panel-edit.js
```diff
@@ -10,6 +10,7 @@ export default class MandateesMandateesPanelEditComponent extends Component {
    * @argument submitter
    * @argument onCancel
    * @argument onSave
+   * @argument {ReferenceDate} Date to get active Mandatees for
```
tdn: See previous comment: `@argument {Date} referenceDate: Date to get active Mandatees for` would be more uniform style-wise with the rest of the app.

app/components/mandatees/mandatees-selector-modal.js
```diff
@@ -6,6 +6,7 @@ import { task } from 'ember-concurrency';
  * @callback {() => Promise} onClose
  * @callback {(mandatee: Mandatee) => Promise} onLink
  * @dependsOn {Mandatee[]} mandatees ('mandatee,mandatee.person')
+ * @argument {ReferenceDate} Date to get active Mandatees for
```
tdn: See previous comment: `@argument {Date} referenceDate: Date to get active Mandatees for` would be more uniform style-wise with the rest of the app.

app/components/publications/publication/case/mandatees/mandatees-panel.js
```diff
@@ -4,10 +4,19 @@ import { tracked } from '@glimmer/tracking';
 
 /**
  * @argument {PublicationFlow} publicationFlow (publication-flow,publication-flow.mandatees,publication-flow.mandatees.person)
+ * @argument {IsViaCouncilOfMinisters}
```
tdn: I think this should be `@argument {Boolean} isViaCouncilOfMinisters`, based on the same logic for documentation style

app/components/utils/mandatee-selector.js
```diff
@@ -2,19 +2,27 @@ import Component from '@glimmer/component';
...
+  'http://themis.vlaanderen.be/id/bestuursfunctie/60d0dc2ab1838d01fca7db65', // Voorzitter
+  'http://themis.vlaanderen.be/id/bestuursfunctie/60d0dc2ab1838d01fca7db67', // Gemeenschapsminister
 ];
-
+/**
+ * @argument singleSelect
+ * @argument readOnly
+ * @argument selectedMandatee
+ * @argument chooseMandatee
+ * @argument {ReferenceDate} Date to get active Mandatees for
```
tdn: See previous comment: `@argument {Date} referenceDate: Date to get active Mandatees` for would be more uniform style-wise with the rest of the app.

app/components/utils/mandatee-selector.js
tdn: > If we decide to take that risk, we should at least document it inline and provide a test case that fails if the filter doesn't work anymore such that the breaking change becomes apparent.

Agreed. I personally vastly prefer this risk over a hardcoded solution, but I agree there should be a mechanism in place to catch it quickly if it should ever fail.

Now that I think about it more, an alternative option would have been to fetch 1 mandatary with the most recent start date, and then use that start date to get the current government body. Bit of a detour, but no hard-coding required. However, the current method is still more efficient and more understandable imo.

So since this works well for now, I'd add the following as comment, and request @ValenberghsSven to write a test to make sure `:has-no:` keeps working in this case:

```
/* WARNING: use of the :has-no:  filter is officially only documented for use with relations, not properties. However, at the time of writing, this method works in the case of xsd:datetime as well. If this test case should ever fail, the loadCurrentBodyMandatees method can be adjusted to fetch the mandatee with the most recent start date, and then use that start date to get the current government body. This is less efficient than the current method, but would require no hard-coding and does not rely on :has-no: */
```

I also added these remarks to the updated PR.

app/components/utils/mandatee-selector.js
```diff
@@ -67,4 +72,31 @@ export default class MandateeSelector extends Component {
       this.mandateeOptions = this.loadMandatees.perform();
     }
   }
+
+  async loadCurrentBodyMandatees(searchTerm) {
```
tdn, line 76: @TaneeG please add a comment here along the lines of: 

```
/* WARNING: use of the :has-no:  filter is officially only documented for use with relations, not properties. However, at the time of writing, this method works in the case of xsd:datetime as well. A test case will be made to check if this keeps working. If this test case should ever fail, the loadCurrentBodyMandatees method can be adjusted to fetch the mandatee with the most recent start date, and then use that start date to get the current government body. This is less efficient than the current method, but would require no hard-coding and does not rely on :has-no: */
```

@ValenberghsSven in the next sprint, could a test case please be added to check whether the `:has-no:` filter keeps working for xsd:datetimes? Created KAS-3402 for this. Thanks!

app/components/utils/mandatee-selector.js
```diff
@@ -32,27 +41,23 @@ export default class MandateeSelector extends Component {
...
-    // We only want the mandatees with no end-date or an end-date in the future.
-    // mu-cl-resources doesn't have :has-no:-capability for properties.
-    return results.filter((mandatee) => {
-      return !mandatee.end || (mandatee.end > new Date());
-    });
+    const currentBodyStartDate = results.firstObject.start;
+    if (currentBodyStartDate > this.referenceDate) {
+      results = yield this.loadPastBodyMandatees(searchTerm);
+    }
+    return results;
```
tdn, line 50: @erikap are you okay with the current order of calls here, given that a test case will be added for `loadCurrentBodyMandatees` and a warning is added? 

This fetches the current government composition first in all cases, since we don't know the start date for it beforehand. If the `referenceDate` is after this start date, we're done (will be the case most of the time).
If the `referenceDate` is before this start date, a new request is made to get the correct past government composition, of which we're sure it'll have an end date.

Advantages I see for this approach vs. [your previous suggestion](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1323/files/ad205a0838ef4a8321c4239101ac7b514760e3ec#r866659408):
- no need for a fallback if getting the government body based on the reference date fails due to the mandatees having no end property (would happen *a lot*, basically for every new agenda item).
- all filtering happens in the backend, which I assume is more efficient than in the browser.

Unless I'm missing something?

tdn (changes requested):

tdn (approved):

app/components/utils/mandatee-selector.js
erikap: I'm having difficulties with the mixture of current government composition and mandatee start/end date. We talk about government bodies in methods/variable names, but in the queries we only filter on mandatee dates. Mandatee and government composition (< government body) are separate resources which both have a start/end date. 

In the current implementation we fetch mandatees without an end-date first and assign the start date of the first result entry (which is a mandatee and not a government body) to `currentBodyStartDate`, but these dates are not necessarily the same. E.g. the current active ministers have a start date of 10/5/2021 while the start date of the current government composition is 02/10/2019.

Just renaming `loadCurrentBodyMandatees` to `loadCurrentActiveMandatees` and `loadPastBodyMandatees` to `loadPastMandatees` will resolve the confusion.

Thinking further the current implementation also assumes the start date is the same for all current active mandatees (the ones without end date), which is the case in the current data, but might not hold in the future. Maybe even the very near future with the resignation of minister Beke. We will land in a case where based on the reference date we need to offer an options list with a mixture of mandatees with **and** without end date. That will not work in the current implementation.

E.g. the mandatee of minister Beke gets an end date of 16/5/2022. If they modify the mandatees on an agendaitem of 29/4/2022 they will want to get a list of the current active mandatees, but also including the mandatee of minister Beke which was active at that moment.

PR #1336, opened 2022-05-13
computed are no longer retriggered, added .content to fix

ValenberghsSven:
During cypress refactor noticed that certain computeds were no longer being re-triggered.
fe. approve agenda with not formal ok > warning message > change to formal ok > warning message still present when it shouldn't be.
Traced it back to the `agenda` computed not recalculating after the `agendaitem` model had a new `formallyOk`.
Added `.content` to the computed before the `@each` fixed it.
Added `.content` to all computed with `@each`, assuming they might also be broken.

### DIFF
--- a/app/models/agenda.js
+++ b/app/models/agenda.js
@@ -59,14 +59,14 @@ export default Model.extend(LoadableModel, {
 
   isFinal: computed.alias('status.isFinal'),
 
-  canBeApproved: computed('agendaitems.@each.formallyOk', function() {
+  canBeApproved: computed('agendaitems.content.@each.formallyOk', function() {
     return this.get('agendaitems').then((agendaitems) => {
       const approvedAgendaitems = agendaitems.filter((agendaitem) => [CONSTANTS.ACCEPTANCE_STATUSSES.OK].includes(agendaitem.get('formallyOk')));
       return approvedAgendaitems.get('length') === agendaitems.get('length');
     });
   }),
 
-  allAgendaitemsNotOk: computed('agendaitems.@each.formallyOk', function() {
+  allAgendaitemsNotOk: computed('agendaitems.content.@each.formallyOk', function() {
     return this.get('agendaitems').then((agendaitems) => {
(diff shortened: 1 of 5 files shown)


ValenberghsSven:
ran 3 times, only fail was a flaky test.

MikiDi:
In general, this is playing with :fire: . I don't think the `content` part of a proxy is [public api](https://api.emberjs.com/ember-data/release/classes/PromiseArray). This is the classic story of derived async data I have no good answer for within our current way of working (see https://chat.semte.ch/group/ember?msg=nFKz2od88dvA9JGSj ), but I very much doubt this is a good idea. Hope someone else can help look in to this.

ValenberghsSven:
> In general, this is playing with fire . I don't think the `content` part of a proxy is [public api](https://api.emberjs.com/ember-data/release/classes/PromiseArray). This is the classic story of derived async data I have no good answer for within our current way of working (see https://chat.semte.ch/group/ember?msg=nFKz2od88dvA9JGSj ), but I very much doubt this is a good idea. Hope someone else can help look in to this.

The current computed properties with `@each` are not retriggering properly, possibly due to ember upgrades?
I made this PR to verify if my "fix" did not break anything else.
The best solution would probably be to avoid using `@each` from now on, since it seems like it no longer works as it used to.

The immediate problem now is that it's possible to get an incorrect list of agendaitems that are not "formally ok" when approving agendaitems which could cause a problem when the frontend shows "everything ok to approve" when in reality we will be doing rollbacks in the backend because of agendaitems that have a "not yet formally ok" status.
I have not looked at he impact of other computed properties.

I could refactor the used computed properties for the agenda actions in favor of this :fire: PR

erikap:
Same feeling as @MikiDi. Might resolve some bugs, but maybe also introduce others. Refactoring the async computed properties on models should probably get a much higher priority...
Before refactoring the ones on "formally OK"  we need to check with design, because they (rightly) questioned whether this feature is being used effectively by the secretary. But I don't know what the outcome of that discussion was.

ValenberghsSven:
Need to dive deeper into this.
I had this issue locally and can't seem to reproduce it on DEV server.
It almost seems like the computed properties behave differently on a local machine, will have to verify and try to break it on DEV to determine the urgency of the refactor.

I would argue that secretary uses the feature, partially because they have to (baked into the app) and partially to control the flow of checked information. There are a set of rules in place by the secretary for when the formally ok status should change (f.e. title change, new piece > yes // only comments get changed > no)
Dropping it would mean losing a way to check whether or not any important changes have recently happened by other members of the team.

PR #1343, opened 2022-05-18
Feature/kas 2670 ingediend door procedurestap zonder ministers decreten vp

sergiofenoll:
Don't display "Nog niet aangevraagd" and "Er zijn nog geen ministers toegevoegd." message for subcases of type "bekrachtiging Vlaamse Regering". Instead, respectively leave the field empty and display "Niet van toepassing".

image


app/pods/cases/case/subcases/subcase/overview/route.js
```diff
@@ -18,6 +18,7 @@ export default class CasesCaseSubcasesSubcaseOverviewRoute extends Route {
 
   async model(params) {
     const subcase = this.modelFor('cases.case.subcases.subcase');
+    const type = await subcase.type;
```
erikap: By 'decoupling' the type from the subcase here, we loose the relation with subcase. Consequently if the user modifies the type of the subcase, this value doesn't get updated. 
Can be resolved by refreshing the model hook, but that's a lot of data reloading for nothing imo. I'm more inclined to use the relationship in the getter so we make use of the benefits of ember-data: `return [CONSTANTS.SUBCASE_TYPES.BEKRACHTIGING].includes(this.model.type?.get('uri'));`

erikap (changes requested):

app/pods/cases/case/subcases/subcase/overview/route.js
sergiofenoll: thanks, didn't take the editing into account.

app/models/subcase-type.js
```diff
@@ -1,12 +1,10 @@
...
-// eslint-disable-next-line ember/no-classic-classes
-export default Model.extend({
-  label: attr('string'),
-  altLabel: attr('string'),
-  scopeNote: attr('string'),
-  subcases: hasMany('subcase', {
-    inverse: null,
-  }),
-});
+export default class Subcase extends Model {
```
erikap: Model name is incorrect

app/pods/cases/case/subcases/subcase/overview/controller.js
```diff
@@ -16,6 +17,10 @@ export default class CasesCaseSubcasesSubcaseOverviewController extends Controll
   @tracked governmentAreas;
   @tracked siblingSubcasesCount;
 
+  get showMandateesNotApplicableMessage() {
+    return [CONSTANTS.SUBCASE_TYPES.BEKRACHTIGING].includes(this.model.subcase.get('type')?.get('uri'));
```
erikap: Is the `get()` for `'type'` required?  `this.model.subcase` is not a proxy object, isn't it?

erikap (changes requested):

app/models/subcase-type.js
sergiofenoll: Darn, I need to pay more attention :facepalm:

app/pods/cases/case/subcases/subcase/overview/controller.js
sergiofenoll: You're right, not needed here. The semantics when it comes to proxy objects still confuse me a bit :woozy_face:

erikap (approved):

PR #1348, opened 2022-05-19
Feature/kas 3354 pagination clean up

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-3354?focusedCommentId=16828

Pagination looked a bit weird for subcases, eg. the pagination panel was fixed at the bottom, so you could see it even before you reached the subcases & the count displayed was wrong.

This PR moves the pagination inside the `SubcasesHistory` component, so that it's only visible once the subcases are scrolled into view. It also includes the current subcase in the list of "sibling" subcases to prevent the wrong number in the pagination panel.

Right now, the current subcase that is shown in the list is shown like all other subcases. We could potentially gray it out/add some other form of styling to visually show that said subcase is the one that's open, but I'm not sure what would make sense (since the item we show in the list can be interacted with) or if it's even necessary to make the distinction.


erikap (approved):

erikap:
Failing tests are not related to this PR.

PR #1350, opened 2022-05-19
Feature/kas 3362 toegangsniveau van vorige document versies aanpassen indien toegangsniveau huidige versie wijzigt

sergiofenoll:
Updates previous access levels of documents.

Not sure if service is necessary here, or if there's a way to access the store inside a normal util.


erikap (commented):
Service is (sadly) also my best answer in Ember atm when you need the ember-data store.

I would just rename the service a little more specifically (e.g. `piece-access-level-service`) to avoid it becoming a mishmash of everything related to pieces. And maybe add some human-readable documentation summarizing the logic that is applied to the method.

erikap (approved):

erikap (approved):

PR #1352, opened 2022-05-20
Feature/kas 3364 wanneer een procedurestap vertrouwelijk wordt krijgt bijhorende beslissingsfiche toegangsniveau ministerraad

sergiofenoll:
Merges into #1350 because I decided to place the new function in the same service for modifying piece access levels.


app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
     }
   }
+
+  /*
+   * This method is used to update the access levels of the pieces of a subcase's treatments. When a subcase has
+   * been updated to being confidential, all the pieces connected to it through agenda-item-treatments get the
+   * "Ministerraad" access level.
+   */
+  async updateDecisionsAccessLevelOfSubcase(subcase) {
+    await subcase.preEditOrSaveCheck();
```
erikap: I'm not familiar with the `preEditOrSaveCheck()`. What's the specific reason to call it here?

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
+
+    if (!subcase.hasDirtyAttributes) {
+      return;
+    }
+
+    let { confidential: [, confidential] = [] } = subcase.changedAttributes();
+    if (!confidential) {
+      return;
+    }
```
erikap: I'm a bit worried about relying on dirty state in a service while saving the record is managed outside of the service. It seems prone to (silent) errors in the future to me. I.e. moving the save operation before calling this method will make the logic silently fail.
I would check the precondition (subcase becoming confidential) outside of this service.

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
+
+    const ministerraad = await this.store.findRecordByUri('access-level', CONSTANTS.ACCESS_LEVELS.MINISTERRAAD);
+
+    const treatments = await subcase.treatments;
+    for (const treatment of treatments.toArray()) {
+      const piece = await treatment.report;
```
erikap: Maybe more efficient to use `this.store.query('document', ...)` with a filter on path to subcase id?
If we include `access-level` in the response we can even easily check, without additional network requests, whether the access level needs to be updated.

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
+
+      piece.accessLevel = ministerraad;
+      await piece.save();
```
erikap: Optimization: executing save-operations in parallel.

erikap (changes requested):

app/services/piece-access-level-service.js
ValenberghsSven: `preEditOrSaveCheck` is the concurrency check on agendaitem/subcase/newsletter-info model.
Typically you use this if you risk saving the model with stale attributes or serialized relations and will block model save until you refresh the page. We check if the `modified` property is still the same in DB as in your local ember store.
Haven't looked any further at the relevant code yet to see if it's needed.

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
+    }
+
+    let { confidential: [, confidential] = [] } = subcase.changedAttributes();
+    if (!confidential) {
+      return;
+    }
+
+    const ministerraad = await this.store.findRecordByUri('access-level', CONSTANTS.ACCESS_LEVELS.MINISTERRAAD);
+
+    const treatments = await subcase.treatments;
```
ValenberghsSven: query for getting all pieces of all treatments of a subcase:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/c7bdbc0bec0e7b5c2d3132ce926b3c275a110859/app/pods/cases/case/subcases/subcase/decision/route.js#L8

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,32 @@ export default class PieceAccessLevelService extends Service {
...
+    for (const treatment of treatments.toArray()) {
+      const piece = await treatment.report;
+
+      piece.accessLevel = ministerraad;
```
ValenberghsSven: Only change + save the piece if the current accesslevel actually needs to be changed.

app/services/piece-access-level-service.js
ValenberghsSven: could Include accessLevel as Erika suggests.

ValenberghsSven: You could also include a filter on accessLevels so that you only get pieces that need changing

app/services/piece-access-level-service.js
ValenberghsSven: Hmm, I'm a bit conflicted here. the use of `preEditOrSaveCheck` seems to be refactored out of some components.
I was still assuming that we run that check when you press the "edit" button.
So you would get a warning that you wont be able to save **before** you **start** your editing work and then a second check **before** actually **saving** since somebody might have just changed it at the same time.
The first check seems to be removed, so we only check on save and do it twice now (also included in `saveSubcaseTitles`)
If you would move the `updateDecisionsAccessLevelOfSubcase` after `saveSubcaseTitles`, the check already happened.
You don't save the subcase, so technically this code doesn't need it.

app/services/piece-access-level-service.js
ValenberghsSven: I would just check if the subcase is confidential and then continue.
We will hit this code more often on those subcases, but the pieces query with possible filter would run and return 0 pieces to change.

ValenberghsSven (changes requested):

app/services/piece-access-level-service.js
sergiofenoll: don't think filter on access level would work, because we can't query for `not ministerraad` and neither for `intern secretarie or intern regering or intern overheid or publiek`, given that we are working with URIs, but I did use a query instead of accessing through relations.

app/services/piece-access-level-service.js
```diff
@@ -62,4 +66,24 @@ export default class PieceAccessLevelService extends Service {
...
+   */
+  async updateDecisionsAccessLevelOfSubcase(subcase) {
+    const ministerraad = await this.store.findRecordByUri('access-level', CONSTANTS.ACCESS_LEVELS.MINISTERRAAD);
+    const pieces = await this.store.query('piece', {
+      'filter[agenda-item-treatment][subcase][:id:]': subcase.id,
+      include: 'access-level',
+    });
+
+    await Promise.all(pieces.toArray().map((piece) => {
+      if (piece.accessLevel.get('uri') !== ministerraad.uri) {
```
erikap: Although this will work (using `get('uri')` will immediately resolve since we included the access-level in the query), we have the habit of still awaiting the promise explicitly to avoid the introduction of bugs in the future when someone changes the `include` statement.
```javascript
const accessLevel = await piece.accessLevel;
if (accessLevel.uri !== ministerraad.uri) {
  ...
}
```

app/services/piece-access-level-service.js
```diff
@@ -80,7 +80,8 @@ export default class PieceAccessLevelService extends Service {
     });
 
     await Promise.all(pieces.toArray().map((piece) => {
-      if (piece.accessLevel.get('uri') !== ministerraad.uri) {
+      const accessLevel = await piece.accessLevel;
```
MikiDi, line 83: Mag je hier `await` gebruiken zonder de lambda-functie ook `async` te maken?

sergiofenoll: gaat inderdaad niet, ben vanochtend te snel geweest met de aanpassing. thanks for the catch!

erikap (approved):

PR #1353, opened 2022-05-20
Feature/kas 3366 beslissingsfiche opladen voor een vertrouwelijke procedurestap krijgt als standaard toegangsniveau ministerraad

sergiofenoll:


app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -28,7 +28,11 @@ export default class AgendaitemDecisionComponent extends Component {
 
   @task
   *loadCodelists() {
-    this.defaultAccessLevel = yield this.store.findRecordByUri('access-level', CONSTANTS.ACCESS_LEVELS.INTERN_REGERING);
+    this.defaultAccessLevel = yield this.store.findRecordByUri(
+      'access-level', this.args.agendaitemIsConfidential
```
ValenberghsSven: `subcaseIsConfidential`?
agendaitem does not have such a property, may be confusing.

app/pods/print-overviews/decisions/agendaitems/template.hbs
```diff
@@ -34,6 +34,7 @@
         {{#each (await agendaitemRow.treatments) as |treatment|}}
           <Agenda::Agendaitem::AgendaitemDecision
             @subcase={{await agendaitemRow.content.agendaActivity.subcase}}
+            @agendaitemIsConfidential={{await agendaitemRow.content.agendaActivity.subcase.confidential}}
```
ValenberghsSven: since we are using `agendaitemRow.content.agendaActivity.subcase` twice now, could we extract it to a `{{#let}}`?
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/8a4b848fcebcef9c751e0ced9311e4a70f2d66f6/app/components/agenda/agenda-header/agenda-action-popup-agendaitems.hbs#L18

Just a suggestion, whatever works is fine.
I can live with this current code since this route is mostly unused by the users I think.

erikap (changes requested):
I think it's better to calculate the required access level only at the moment the user uploads a new decision report. As such the model hook is not busy loading the subcase each time the user navigates to the decision tab. Data is only loaded if needed (ie. during upload of a new report) which will only be in a limited subset of the cases.

erikap (approved):

PR #1355, opened 2022-05-23
Feature/kas 3368 codelijst van toegangsniveau aanpassen naar concepts met concept scheme

sergiofenoll:
Backend: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/283


sergiofenoll:
Custom jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/2/

MikiDi:
Just a thought (probably ok here since all seem replaced): I wonder how equality is determined for ember-data records, or: in case you missed a place where you still fetch `access-level` instead of `concept`, would that result as equal when comparing to its concept-counterpart, or not.  
Note that the situation mentioned above could occur in more subtle ways than by leaving an `this.store.query('access-level' ...` untouched. Also a leftover `belongsTo('access-level')` combined with `myEntity.accessLevel` would do that.

sergiofenoll:
There were some ModelSelectors that used access level which I had overlooked, replaced those.
New Jenkins run: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/4/

sergiofenoll:
> Just a thought (probably ok here since all seem replaced): I wonder how equality is determined for ember-data records, or: in case you missed a place where you still fetch `access-level` instead of `concept`, would that result as equal when comparing to its concept-counterpart, or not. Note that the situation mentioned above could occur in more subtle ways than by leaving an `this.store.query('access-level' ...` untouched. Also a leftover `belongsTo('access-level')` combined with `myEntity.accessLevel` would do that.

Had a quick look, maybe Ember uses some of the mechanisms detailed [here](https://github.com/emberjs/rfcs/blob/master/text/0403-ember-data-identifiers.md) to determine record equality. For what it's worth, I think in general we always check equality on a record's ID, which means in case like this we wouldn't have a problem since the resources UUIDs also get changed.

erikap:
> Custom jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/2/

Result of this custom built is not available anymore :see_no_evil:  Can you trigger a new one @sergiofenoll (after merging development in this branch first)?

sergiofenoll:
> > Custom jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/2/
> 
> Result of this custom built is not available anymore see_no_evil Can you trigger a new one @sergiofenoll (after merging development in this branch first)?

New jenkins run:
~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/5/~ (this one was failing)
~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/6/~ (idem)
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/7/
Also reminder (to myself) to undo the changes to `ci/docker-compose.override.yml` before merging

erikap (approved):

sergiofenoll:
Jenkins build is green, undoing changes to `ci/docker-compose.override.yml`

app/components/utils/access-level-selector.js
```diff
@@ -0,0 +1,26 @@
...
+   * @argument allowClear
+   * @argument multiple
+   * @argument selected
+   * @argument onChange
+   * @argument filterOptions: a function that will filter out results from the dropwdown menu
+   */
+  get filter() {
+    return {
+      'concept-schemes': {
+        ':uri:': CONSTANTS.CONCEPT_SCHEMES.ACCESS_LEVELS,
```
ValenberghsSven, line 22: right now, this line is the only code making this component only usable for `access-levels`.
We could eventually abstract this component to a `ConceptSelector`, and pass in the relevent uri for the scheme we want to search for.

cypress/integration/unit/cancel-editing-document.spec.js
```diff
@@ -24,7 +24,7 @@ function uploadFileToCancel(file) {
 
 function openAgendaitemDocumentBatchEdit() {
   cy.intercept('GET', '/document-types?**').as('getDocumentTypes');
-  cy.intercept('GET', '/access-levels?**').as('getAccessLevels');
```
ValenberghsSven, line 27: this might not do the trick.
the intercept is very vague. Might no longer be needed though, need to check.

ValenberghsSven (approved):
looks ok, need to verify all tests work non-flaky with the accessLevel concepts also loading in the application route now.

app/components/utils/access-level-selector.js
sergiofenoll: Definitely, though we'd have to allow filtering on extra fields. E.g. for the kind selector, we also need to filter on [`:has-no:narrower`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/utils/kind-selector.js#L35). Definitely worth it to think about an elegant interface for a `ConceptSelector`.

ValenberghsSven:
> looks ok, need to verify all tests work non-flaky with the accessLevel concepts also loading in the application route now.

the branch is quite stable so if nothing else pops up we could merge.
yggie > backend > this one

PR #1356, opened 2022-05-23
KAS-3342 dynamic mandatee selection based of reference date

MikiDi:
successor of https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1323
Gathers shared mandatee-fetching logic in a services and follows a `referenceDate -> governmentBody -> active mandatees`-flow with client-side name-filtering.

Goes with https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/282

:information_source: Probably still needs some tweaks to Cypress scenarios


erikap (approved):

app/components/publications/publication/case/mandatees/mandatees-panel.hbs
```diff
@@ -65,6 +65,10 @@
 
 {{#if this.showSelectMandateeModal}}
   <Mandatees::MandateesSelectorModal
+    @referenceDate={{if @viaCouncilOfMinisters
```
ValenberghsSven: The correct variable from args is `isViaCouncilOfMinisters`

app/components/mandatees/mandatee-selector.js
```diff
@@ -0,0 +1,52 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { isEmpty } from '@ember/utils';
+import { tracked } from '@glimmer/tracking';
+import { restartableTask, task, timeout } from 'ember-concurrency';
+import { inject as service } from '@ember/service';
+
+/**
+ * @argument multiple
+ * @argument disabled
```
ValenberghsSven, line 10: Only 1 template uses the `Mandatees::MandateeSelector` and does not pass in `multiple` or `disabled` as args.
So they currently serve no function but I guess we can keep this in for now though. Perhaps we need this later.

But it does mean we can't really test the `multiple` part of the template, so it might not work as expected should we ever need it in the future.

app/pods/settings/ministers/controller.js
```diff
@@ -1,13 +1,13 @@
...
+// import { inject as service } from '@ember/service';
 // import { task } from 'ember-concurrency';
 // import { isPresent } from '@ember/utils';
 
 // TODO: octane-refactor
 /* eslint-disable ember/no-get */
 // eslint-disable-next-line ember/no-classic-classes
 export default Controller.extend({
-  currentSession: service(),
+  // currentSession: service(),
```
ValenberghsSven: This was the last piece of actual code in this controller.
Out of scope, but I'm guessing we can just fall back to ember's default empty controller.
We kept most of this code 'in case we enable mandatee editing again'.
I'm pretty sure the current structure of government bodies will be too complex to allow users to do this without extra tooling or help from our part, so just opt the remove the legacy code?
Follow ticket/task is fine so this PR can move.

app/services/mandatees.js
```diff
@@ -0,0 +1,93 @@
+import Service, { inject as service } from '@ember/service';
+import { task } from 'ember-concurrency';
+import { PAGE_SIZE } from 'frontend-kaleidos/config/config';
+
+const VISIBLE_ROLES = [
+  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03de', // Minister-president
+  'http://themis.vlaanderen.be/id/bestuursfunctie/5fed907ce6670526694a03e0', // Minister
+  'http://themis.vlaanderen.be/id/bestuursfunctie/60d0dc2ab1838d01fca7db65', // Voorzitter
+  'http://themis.vlaanderen.be/id/bestuursfunctie/60d0dc2ab1838d01fca7db67', // Gemeenschapsminister
+];
```
ValenberghsSven, line 10: I could argue to have these in config, but they are only used here to this is fine.

app/services/mandatees.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+    const [closedBody, openBody] = yield Promise.all([closedRange, openRange]);
+    const existingBody = closedBody || openBody;
+    if (existingBody) {
+      if (closedBody && openBody) {
+        console.warn('Multiple active government bodies for given referenceDate. Probably something wrong in data');
+      }
+      return existingBody;
+    }
+    return null;
```
ValenberghsSven: This one is a bit confusing on a shallow skim of the code.
You would suspect that the filters on `start` and `end` are attributes, but they are relations to `generation` and `invalidation` models that are only defined in backend.

Do we want to document this more clearly? Spent like 10 minutes questioning why this worked ... :sweat_smile: 
Would it make more sense to see `filter[generation]` `filter[invalidation]` `:has-no:invalidation` ?

I was also confused because we do have attributes `mandaat:bindingStart` and `mandaat:bindingEinde` for older `government-bodies` in a migration, but they are never used outside of that migration. I first thought that we are referencing these attributes.

app/services/mandatees.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @task
+  *fetchMandateesForGovernmentBody(governmentBody, referenceDate=null) {
```
ValenberghsSven, line 55: since we fallback to a `new Date` in `getMandateesActiveOn`, referenceDate would never be null.
Is there a situation where this would be needed? If not, this is just adding dead code?

ValenberghsSven (changes requested):
Only 1 place needs changing (the args typo).

The rest are comments and questions.

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/239/
New run with testdata zip from backend and fix in cypress commands
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/240/

app/services/mandatees.js
MikiDi: > referenceDate would never be null.

Correct. It seemed a valuable addition for the upcoming stories regarding possibilities for more flexible (broader) minister-selection though. `fetchMandateesForGovernmentBody` without specifying a referenceDate can be seen as an "inbetween" compared to "only active ministers"/"áll ministers".

app/services/mandatees.js
MikiDi: It's a naming thing (copied from how it's specified in Themis). Will adapt to `generation` and `invalidation`. :+1:

app/pods/settings/ministers/controller.js
MikiDi: I've been thinking this as well. Might as well get it done. Will strip out what's no longer needed/accurate.

app/components/mandatees/mandatee-selector.js
MikiDi: Correct regarding both using and testing. Expect to need them (or at least `multiple`) in the future. Would prefer to keep them for now.

MikiDi:
added https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1356/commits/34b221570abc84f1602b82e6a78179700b67d29e to avoid race condition where gov bodies attempt loading before visible roles are available

app/components/publications/publication/case/mandatees/mandatees-panel.hbs
MikiDi: I'm now wondering if the reasoning behind this is even correct to begin with. Don't we always want  `@publicationFlow.agendaItemTreatment.startDate` if it's available, regardless of "via MR" or not? Regardless it's the mandatee in charge on the decision-date that's responsible, no?

app/services/mandatees.js
MikiDi: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1356/commits/8fd181a7967731418d29b803374e97ceda4d120d

app/pods/settings/ministers/controller.js
MikiDi: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1356/commits/23eb68d6171a259c5f5a8c30b0ba1c5828cd48f7

app/services/mandatees.js
ValenberghsSven: So we keep it in, just keep in mind it's untested when implementing it

app/components/mandatees/mandatee-selector.js
ValenberghsSven: :+1:

PR #1361, opened 2022-05-25
Feature/kas 3385 zoekfunctie tab onthouden

sergiofenoll:
Makes use of a service to keep track of the last visited search tab. Tried setting the `@tracked` directly on the service (so without the `properties = new class {...}` intermediate object), but that didn't seem to work.

Also replaced the page title of ["Korst bestek" search](https://kaleidos-dev.vlaanderen.be/zoeken/kort-bestek). It used to be `kort bestek`, now it's `Kort bestek` (capitalized, like the other search tabs). Change can be reverted if this was working as intended before.


erikap:
As discussed with the team, this is the first time a custom user pattern like this is applied in the app. The proposed change works, but doesn't have a high sustainability score. A more thorough analysis of the exact user need is necessary and will be included in the reworking of the search functionality in the near future. Therefore going to close this PR without merging for now.

erikap:
As discussed with the team, this is the first time a custom user pattern like this is applied in the app. The proposed change works, but doesn't have a high sustainability score. A more thorough analysis of the exact user need is necessary and will be included in the reworking of the search functionality in the near future. Therefore going to close this PR without merging for now.

PR #1364, opened 2022-05-27
Feature/kas 3408 opheldering datum statusaanpassing vs publicatiedatum

sergiofenoll:
No longer display the last edited date in the publication status pill.

Refreshes the publication page when editing a pub status to prevent stale data from being rendered (I didn't find a more elegant without changing a lot of the innards of the components, because we're side loading lots of relations in components, so refreshing seemed the easiest solution here).

From the ticket, I couldn't easily find an example of 

> 3. Datum op de timeline. Hier wordt momenteel in het geval van een gepubliceerd dossier de publicatiedatum uit 2. getoond, maar in alle andere gevallen de wijzigingsdatum uit 1. Dit is heel verwarrend, en hier moet dus sowieso verduidelijkt worden wat er getoond wordt.

Locally, cases that haven't been published yet look like this for me, but perhaps I'm missing something. Note the lack of a wrong publication date in the timeline.
image


app/components/publications/editable-publication-status-pill.js
```diff
@@ -75,6 +76,7 @@ export default class PublicationStatusPill extends Component {
       );
       this.loadStatus.perform();
       this.loadDecision.perform();
+      this.router.refresh('publications.publication')
```
ValenberghsSven: do we need to both performs since we are refreshing the route?
I would expect a re-render and this component to be reloaded, hitting the constructor again.

ValenberghsSven (approved):
This seems ok

app/components/publications/editable-publication-status-pill.js
erikap: In case the route refreshes and the component will re-render the constructor will indeed be hit again. 

I still need to have a look in depth at this PR. Refreshing the router from inside a component looks like a risky pattern to introduce at first sight.

sergiofenoll: Agreed that it's quite drastic, but was the only solution I could find (though there might be better ways that I just haven't found).

For context: [this component](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/publications/publication/publication-activities/publications-info-panel.hbs#L59) is rendered in the "Publicaties" tab and it contains the "Publicatiedatum" field that we want to show the user. Here, the publication date is fetched in task before showing it to the user. When editing the publication status to published via the pill at the top of the page, we want the date to get updated and displayed in this page.

I couldn't find a way to force the task to re-fire (so that the new data gets loaded), so I went for a refresh.

ValenberghsSven: "Refreshing the router from inside a component looks like a risky pattern to introduce at first sight."
I get what you mean, if this component is reused in some other route this would not work.
Passing a controller method down to do the component and call that with `this.args.refreshRoute` seems like a better option. Then we do the refreshing in the controller of the route we are in.

erikap (approved):

PR #1365, opened 2022-05-30
Feature/kas 2636 uitfaseren van pod structuur voor routes

sergiofenoll:
Moves from pod structure to classic structure for the routes

Probably good to await further discussion with team to see if we want to go through with this. PR is open for 1) tests and 2) in case anyone wants to try out the classic structure locally

EDIT: Setting to draft since this is one we should merge when most other PRs have been merged (and maybe we still want/need to discuss whether this is desired). So if you're looking for PRs to review, **ignore this one** (for now) :wink:


ValenberghsSven (commented):
Difficult to approve, since there will be constant conflicts.
I would aim to wait for some of the larger PR's that are close to being merged.
Then we should just go for it and update any other branch that is open, being very wary of files moving back to pods because of how git merges sometimes.

PR #1367, opened 2022-05-31
Feature/kas 3205 cleanup van computed properties in subcase model

sergiofenoll:
Removes computed properties from subcase model & refactors model to Octane syntax


app/utils/agendaitem-utils.js
```diff
@@ -70,7 +70,7 @@ export const setModifiedOnAgendaOfAgendaitem = async(agendaitem) => {
  * @param resetFormallyOk
  * @returns {Promise<void>}
  */
-export const saveChanges = async(agendaitemOrSubcase, propertiesToSetOnAgendaitem, propertiesToSetOnSubcase, resetFormallyOk) => {
+export const saveChanges = async(agendaitemOrSubcase, propertiesToSetOnAgendaitem, propertiesToSetOnSubcase, resetFormallyOk, store) => {
```
sergiofenoll: I don't love this change, it feels weird to pass in the store to this util so that it can fetch the needed agendaitems.

An alternative is fetching the agendaitems through the relations on the subcase, i.e. `subcase.agendaActivities.mapBy('agendaitems')`, getting their agenda & checking status, but that implies a bunch of extra queries.

Alternatively this function could be split into two, one for saving changes "from" a subcase, and other one for saving "from" an agendaitem, and then it's the caller's responsibility to fetch the necessary agendaitems, but that means some code duplication in ~5 places.

app/components/subcase/subcase-titles-panel/subcase-titles-panel-view.js
```diff
@@ -17,7 +18,26 @@ export default class SubcaseTitlesPanelView extends Component {
 
   @task
   *loadApproved() {
-    this.approved = yield this.args.subcase.approved;
+    const meeting = yield this.args.subcase.requestedForMeeting;
```
sergiofenoll: Maybe this should be moved to a util, but again the store is needed and it feels weird to pass in the store for fetching. Input welcome.

app/utils/agendaitem-utils.js
ValenberghsSven: We could toss this function in a new service instead?
The entire syncing of agendaitem/subcase was always a troublesome piece of code.
In any case, caution is advised.

cypress/integration/all-flaky-tests/subcase.spec.js
```diff
@@ -107,6 +107,7 @@ context('Subcase tests', () => {
     cy.visit('/dossiers/5F02E3F87DE3FC0008000002/deeldossiers');
     cy.addSubcase(type, shortSubcaseTitle, subcaseTitleLong, subcaseType, subcaseName);
     cy.openSubcase(0);
+    cy.wait(2000);
```
sergiofenoll, line 110: also stated in the commit message but wanted to reiterate here:

Not the ideal solution, but it seems to fix an error that's caused by
Cypress immediately deleting the subcase while some requests are still
in flight.

Error at hand:
```
1) Subcase tests
    should add a subcase and then delete it:
   Error: The following error originated from your application code, not from Cypress.

> Attempted to handle event `pushedData` on <subcase:629675108AADBDDBE93A3D46> while in state root.deleted.inFlight.
```
I am guessing that the timing of the requests are off, and that perhaps
the way computed properties work make it so deleting the subcase record
while the requests are still in flight causes an error here.

ValenberghsSven: if this is caused by deleting subcase, it would make more sense to await in the test where we acutally delete the subcase. Will look at this when reviewing

app/components/subcase/subcase-titles-panel/subcase-titles-panel-view.js
sergiofenoll: Moved to a service

app/components/subcases/subcase-header.js
```diff
@@ -24,17 +24,17 @@ export default Component.extend({
...
     } = this;
     const {
       subcase,
     } = this;
     const requestedForMeeting = await subcase.get('requestedForMeeting');
-    const hasActivity = await subcase.get('hasActivity');
+    const activities = await subcase.get('agendaActivities');
 
-    if (hasActivity || requestedForMeeting || isAssigningToOtherAgenda || isLoading) {
+    if (activities?.legnth || requestedForMeeting || isAssigningToOtherAgenda || isLoading) {
```
erikap: Typo in `legnth`

app/services/subcase-is-approved.js
```diff
@@ -0,0 +1,30 @@
+import Service, { inject as service } from '@ember/service';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+
+export default class SubcaseIsApprovedService extends Service {
+  @service store;
+
+  async approved(subcase) {
```
erikap: Can you change the name to something that clarifies that it will return a boolean?  `isApproved`?

app/services/subcase-is-approved.js
```diff
@@ -0,0 +1,30 @@
...
+      const approvedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD
+      );
+      const acknowledgedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME
+      );
+
+      return !!(await this.store.queryOne('agenda-item-treatment', {
```
erikap: There is also a `this.store.count` in our extented data store service

app/services/subcase-is-approved.js
```diff
@@ -0,0 +1,30 @@
...
+        'filter[subcase][id]': subcase.id,
```
erikap: Better to use `:id:` (since `id` is not officially supported by mu-cl-resources)

app/models/subcase.js
```diff
@@ -1,99 +1,32 @@
...
-        return !!approvedTreatment || !!acknowledgedTreatment;
-      }
-    }
-    return false;
-  }),
-
-});
+export default class Subcase extends ModelWithModifier {
+  @service store;
+  @service intl;
```
erikap: Not used anymore?

app/services/save-agendaitem-changes.js
```diff
@@ -0,0 +1,38 @@
+import Service, { inject as service } from '@ember/service';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+import { setNewPropertiesToModel, setModifiedOnAgendaOfAgendaitem } from 'frontend-kaleidos/utils/agendaitem-utils';
+
+export default class SaveAgendaitemChangesService extends Service {
+  @service store;
+
+  async saveChanges(agendaitemOrSubcase, propertiesToSetOnAgendaitem, propertiesToSetOnSubcase, resetFormallyOk) {
+    const item = agendaitemOrSubcase;
+    const isAgendaitem = item.get('modelName') === 'agendaitem';
```
erikap: No need to use `get()`

app/services/save-agendaitem-changes.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+      await setNewPropertiesToModel(item, propertiesToSetOnAgendaitem, resetFormallyOk);
+      await setModifiedOnAgendaOfAgendaitem(item);
+    } else {
+      await setNewPropertiesToModel(item, propertiesToSetOnSubcase, false);
+      const agendaitemsOnDesignAgendaToEdit = await this.store.query('agendaitem', {
+        'filter[agenda-activity][subcase][:id:]': item.id,
+        'filter[agenda][status][:uri:]': CONSTANTS.AGENDA_STATUSSES.DESIGN,
+      });
+      if (agendaitemsOnDesignAgendaToEdit && agendaitemsOnDesignAgendaToEdit.get('length') > 0) {
```
erikap: No need to use `get()`?

app/components/web-components/vl-decisions-column.js
```diff
@@ -1,21 +1,49 @@
...
+  @tracked textToShow;
+
+  constructor() {
+    super(...arguments);
+
+    this.loadTextToShow.perform();
+  }
+
+  @task
+  *loadTextToShow() {
```
erikap, line 19: Use `SubcaseIsApprovedService` or does this contain different logic? In that case, same comments as in `SubcaseIsApprovedService` apply here.

app/components/subcases/subcase-header.js
```diff
@@ -24,17 +24,17 @@ export default Component.extend({
   caseToDelete: null,
   onMoveSubcase: null, // from args, the parent route subcases needs to be refreshed after subcase is moved/deleted
 
-  canPropose: computed('subcase.{requestedForMeeting,hasActivity}', 'isAssigningToOtherAgenda', async function() {
+  canPropose: computed('subcase.{requestedForMeeting,agendaActivities}', 'isAssigningToOtherAgenda', async function() {
```
erikap: Is it an option to refactor this component to a Glimmer component in this PR or will that lead us too far?

app/components/subcases/subcase-header.js
```diff
@@ -24,17 +24,17 @@ export default Component.extend({
...
     const {
       isAssigningToOtherAgenda, isLoading,
     } = this;
     const {
       subcase,
     } = this;
     const requestedForMeeting = await subcase.get('requestedForMeeting');
-    const hasActivity = await subcase.get('hasActivity');
+    const activities = await subcase.get('agendaActivities');
```
erikap: No need to use `get()`  (occurs a couple of times in this file)

erikap (changes requested):

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
```diff
@@ -54,11 +54,11 @@ export default class AgendaitemCasePanelEdit extends Component {
       confidential: this.args.subcase?.confidential,
     };
 
-    yield saveSubcaseTitles(
+    yield this.saveAgendaitemChanges.saveChanges(
```
ValenberghsSven: Not fully convinced about the name of the service.
still thinking what would be a better name (+ maybe comments in the service)

just semantics, could run with it for now

app/components/subcases/subcase-header.js
ValenberghsSven: I prefer the latter, this is quite a big component.

app/components/web-components/vl-decisions-column.js
ValenberghsSven: If we do use the service, we may want to make sure we use `subcase?.requestedForMeeting;` with the optional token.
The bottom part of the logic seems the same except for that optional token.

app/pods/cases/case/subcases/subcase/documents/controller.js
```diff
@@ -147,7 +147,8 @@ export default class CasesCaseSubcasesSubcaseDocumentsController extends Control
     if (documentContainer) {
       const lastPiece = await documentContainer.get('lastPiece'); // TODO: what is the purpose of getting lastPiece here
       if (this.subcase && lastPiece) {
-        const latestActivity = await this.subcase.get('latestActivity');
+        const agendaActivities = await this.subcase.agendaActivities;
+        const latestActivity = agendaActivities.sortBy('startDate')?.lastObject;
         if (latestActivity) {
           await latestActivity.hasMany('agendaitems').reload(); // This fixes a case where approving an agenda did not update latestAgendaitem
           const latestAgendaitem = await latestActivity.get('latestAgendaitem');
```
ValenberghsSven, line 154: The computed replace is ok (from looking at code, not manually confirmed yet)

Perhaps this whole part is a little complex that might be better solved with 1 or more queries (not now, later)
This logic dates back to a time where multiple agenda-activities were possible for the purpose of postponing subcases and reusing them on a new meeting (so 1 agenda-activity per meeting), but the subcases are being copied instead so we normally only have 1 activity (need to query to make sure it's always 1).

What we need here= if the piece in question was a "next-version" linked to an agendaitem, we have to attach the "previous-version" to the same agendaitem or the whole document might no longer be linked.
So we have to find a way to get the agendaitem, via subcase, via its (latest) activity, via its (latest) agendaitem.
Pretty sure that we can find a query that covers that.

app/services/subcase-is-approved.js
```diff
@@ -0,0 +1,30 @@
+import Service, { inject as service } from '@ember/service';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+
+export default class SubcaseIsApprovedService extends Service {
+  @service store;
+
+  async approved(subcase) {
+    const meeting = await subcase.requestedForMeeting;
```
ValenberghsSven: to make this method more error safe: `subcase?.`

cypress/integration/all-flaky-tests/subcase.spec.js
ValenberghsSven: > if this is caused by deleting subcase, it would make more sense to await in the test where we acutally delete the subcase. Will look at this when reviewing

Judged this a bit too quick, this is indeed the test it should be in.
I will check if there is a more viable solution where we await certain calls, since the showing of the delete button is tied to some computed properties
Possibly hiding the entire action menu when we are loading would be a good fix.

app/pods/cases/case/subcases/subcase/documents/controller.js
ValenberghsSven: perhaps 
```
if (this.subcase && lastPiece) {
const latestAgendaitem = await this.store.queryOne('agendaitem', {
        'filter[agenda-activity][subcase][:id:]': this.subcase.id,
        'filter[:has-no:next-version]': 't',
        sort: '-created',
      });
      
      await restorePiecesFromPreviousAgendaitem(latestAgendaitem, documentContainer);
          // TODO: make sure we're not loading stale cache
      await latestAgendaitem.hasMany('pieces').reload();
  }
```

I see you changed `latestActivity.get('latestAgendaitem');` in the other PR, so maybe it's better to do it there instead

app/components/subcases/subcase-header.js
ValenberghsSven: well nevermind, it's done now :D

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
erikap: Something with `agendaitem`, `subcase` and `sync`?
`AgendaitemAndSubcasePropertiesSyncService`?

app/services/save-agendaitem-changes.js
```diff
@@ -0,0 +1,38 @@
+import Service, { inject as service } from '@ember/service';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+import { setNewPropertiesToModel, setModifiedOnAgendaOfAgendaitem } from 'frontend-kaleidos/utils/agendaitem-utils';
```
erikap: Can we embed these utils in this service since it seems the only place where they are used.

app/services/agendaitem-and-subcase-properties-sync.js
```diff
@@ -0,0 +1,75 @@
...
+  return model.reload();
+};
+
+/**
+ * @description Zet de modified date property van een agenda op basis van de doorgegeven agendaitem
+ * @param agendaitem Het agendaitem om de agenda mee op te vragen.
+ * @returns {Promise<void>}
+ */
+const setModifiedOnAgendaOfAgendaitem = async(agendaitem) => {
+  const agenda = await agendaitem.get('agenda');
```
erikap: No need to use `get()`

app/services/agendaitem-and-subcase-properties-sync.js
```diff
@@ -0,0 +1,75 @@
...
+  const isDesignAgenda = await agenda.asyncCheckIfDesignAgenda();
+  if (agenda && isDesignAgenda) {
+    agenda.set('modified', moment().utc()
+      .toDate());
```
erikap: Just use `new Date()` instead of `moment`

app/services/agendaitem-and-subcase-properties-sync.js
```diff
@@ -0,0 +1,75 @@
...
+    setNotYetFormallyOk(model);
+  }
+
+  const keys = Object.keys(propertiesToSet);
+  for (const key of keys) {
+    model.set(key, propertiesToSet[key]);
+  }
+
+  await model.save();
+  return model.reload();
```
erikap, line 23: I didn't notice this before. Any idea why we do the `model.reload()` explicitely after the `model.save()` @ValenberghsSven ?

app/services/subcase-is-approved.js
```diff
@@ -0,0 +1,30 @@
...
+      const approvedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD
+      );
+      const acknowledgedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME
+      );
+
+      return await this.store.count('agenda-item-treatment', {
```
erikap: Maybe add  `> 0` such that we always return a boolean?

erikap (changes requested):

app/services/agendaitem-and-subcase-properties-sync.js
ValenberghsSven: not really sure, it's been in there for a long time.
Could be concurrency related
Could leave it in for now and use a seperate PR to see what jenkins does with it.

Some of those legacy utils are ... questionable to current code quality.
Lots of generic pieces of code that were used in shared generic components for agendaitem/subcase.

I wouldn't mind taking a stab at refactoring some of the utils to be more specific (save changes to subcase specifically instead of calling a method that also works for agendaitems)
In a follow-up ticket of course

erikap: You're right. We can keep diving and improving :see_no_evil: Better to handle further improvements in another story (or in the margin of related stories)

erikap (approved):

app/pods/cases/case/subcases/subcase/documents/controller.js
erikap: Will be resolved in #1368

PR #1368, opened 2022-06-01
Feature/kas 3427 refactoren en verwijderen van computed properties op ember data models

sergiofenoll:


app/components/agenda/agenda-header/agenda-action-popup-agendaitems.js
```diff
@@ -0,0 +1,21 @@
...
+  @service store;
+  @service agendaService;
+
+  @bind
+  async newsletterInfo(agendaitem) {
+    const newsletterInfos = await this.store.query('newsletter-info', {
+      'filter[agenda-item-treatment][agendaitem][:id:]': agendaitem.id,
+    });
+    return newsletterInfos.firstObject;
+  }
```
ValenberghsSven: I know this was the exact computed, do we want to queryOne instead? (should only be one unless data was incorrectly connected)

app/components/agenda/agenda-header/agenda-action-popup-agendaitems.js
```diff
@@ -0,0 +1,21 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import bind from 'frontend-kaleidos/utils/bind';
+
+export default class AgendaAgendaHeaderAgendaActionPopupAgendaitemsComponenet extends Component {
+  @service store;
+  @service agendaService;
+
+  @bind
```
ValenberghsSven, line 8: First thought this was an ember thing, but it's our own (borrowed) code.
Ember already has something called bind, will this cause confusion?
https://api.emberjs.com/ember/3.28/functions/@ember%2Frunloop/bind


If we want to get rid of `await` in the template, we could add a component per agendaitem and load data in it's constructor but that will be out of scope.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -79,6 +81,14 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
   }
 
+  @bind
+  async allAgendaitemsNotOk() {
+    const agendaitems = await this.args.currentAgenda.agendaitems;
+    return agendaitems
+          .filter((agendaitem) => [CONSTANTS.ACCEPTANCE_STATUSSES.NOT_OK, CONSTANTS.ACCEPTANCE_STATUSSES.NOT_YET_OK].includes(agendaitem.get('formallyOk')))
+          .sortBy('number');
+  }
+
```
ValenberghsSven: asking myself if we would be better off with a query.

The only issue I can think of is that the agendaitems from the query are different from the ones in store (in some cases we either set formally ok's or rolling back dirty attributes)

app/components/agenda/agenda-header/agenda-version-actions.hbs
```diff
@@ -225,7 +225,7 @@
           @skin="error"
           @message={{t
             "agenda-design-delete-message"
-            agenda=this.latestAgenda.agendaName
+            agenda=(await (agenda-name this.latestAgenda))
```
ValenberghsSven: does awaiting work properly in translations? I thought not

ValenberghsSven (commented):
Was reviewing, but it got late.
Completing this partial review for now

app/models/agenda.js
```diff
@@ -1,105 +1,41 @@
...
-      const previousVersion = await agendaitem.get('previousVersion');
-      if (previousVersion) {
-        approvedAgendaitems.pushObject(agendaitem);
-      }
-    }
-    return approvedAgendaitems;
-  }),
+  // TODO: Either refactor these two or figure out how to make them accessible using @tracked ?
+  @reads('status.isDesignAgenda') isDesignAgenda;
+  @reads('status.isApproved') isApproved;
```
erikap: Just make get()-functions?
```javascript
get isDesignAgenda() {
  return this.status.get('isDesignAgenda');  // need to use get() since this.status returns a promise
}
```
But I don't see a lot of added value in these properties? Are they used/required?

app/models/agenda.js
```diff
@@ -1,105 +1,41 @@
...
 
-});
+  async asyncCheckIfDesignAgenda() {
+    await this.status;
+    return this.isDesignAgenda;
+  }
```
erikap: Seems to be used only in agendaitem-utils. Can it be embedded there? I don't see added value in this separate function.

app/models/meeting.js
```diff
@@ -44,52 +40,11 @@ export default Model.extend({
...
-    });
-  }),
-
-  sortedAgendas: computed('agendas.@each.agendaName', function() {
-    return PromiseArray.create({
-      promise: this.get('agendas').then((agendas) => agendas.sortBy('agendaName').reverse()),
-    });
-  }),
-
+  // TODO this computed property is used in:
```
erikap: Since there is only 1 simple computed left in this model, it can be refactored to Octane (and the computed property to a getter).

app/models/meeting.js
```diff
@@ -1,11 +1,7 @@
...
-import { PromiseArray, PromiseObject } from '@ember-data/store/-private';
 import { computed } from '@ember/object';
 import { inject } from '@ember/service';
 import { KALEIDOS_START_DATE } from 'frontend-kaleidos/config/config';
-import {
-  sortDocumentContainers
-} from 'frontend-kaleidos/utils/documents';
 
 // TODO: octane-refactor
 /* eslint-disable ember/no-get */
```
erikap: Do the eslint-disable rules on top of this file still apply?

app/models/agendaitem.js
```diff
@@ -59,36 +55,20 @@ export default ModelWithModifier.extend({
...
-              size: pieces.get('length'), // # documentContainers will always be <= # pieces
-            },
-            include: 'type,pieces,pieces.access-level,pieces.next-piece,pieces.previous-piece',
-          }).then((containers) => sortDocumentContainers(this.get('pieces'), containers));
-        }
-      }),
-    });
-  }),
-
   isDesignAgenda: reads('agenda.isDesignAgenda'),
```
erikap: Do we have an idea whether this is actually used?

app/utils/bind.js
```diff
@@ -0,0 +1,23 @@
+// Borrowed from https://github.com/NoHomey/bind-decorator
+// via https://github.com/statelyai/xstate/blob/231a6eb1dd47d05b8bf37181e435bb2d99afdeac/packages/core/src/utils.ts#L737
+export default function bind(target, propertyKey, descriptor) {
```
erikap, line 3: Can you explain what this util does?

app/pods/cases/case/subcases/subcase/documents/controller.js
```diff
@@ -150,11 +150,22 @@ export default class CasesCaseSubcasesSubcaseDocumentsController extends Control
...
+          const agendaitems = await latestActivity.hasMany('agendaitems').reload(); // This fixes a case where approving an agenda did not update latestAgendaitem
+          const meeting = await this.subcase.requestedForMeeting;
+          const agendas = await meeting.agendas;
+          const sortedAgendas = agendas.sortBy('serialnumber').reverse();
+          const latestAgenda = sortedAgendas.firstObject;
+          for (let index = 0; index < agendaitems.length; index++) {
+            const agendaitem = agendaitems.objectAt(index);
+            const agenda = await agendaitem.get('agenda');
+
+            if (agenda.get('id') === latestAgenda.get('id')) {
```
erikap: No need to use `get()`

app/helpers/agenda-name.js
```diff
@@ -0,0 +1,11 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(async function agendaName([agenda]) {
+  const isDesignAgenda = await (await agenda.status).isDesignAgenda;
+  const agendaName = agenda.serialnumber || '';
+  let prefix = 'Agenda';
+  if (isDesignAgenda) {
+    prefix = 'Ontwerpagenda';
+  }
```
erikap: Shorter `let prefix = isDesignAgenda ? 'Agenda' : 'Ontwerpagenda'`. These hardcoded strings probably need to come from translations.

app/helpers/agenda-name.js
```diff
@@ -0,0 +1,11 @@
...
+  return `${prefix} ${agendaName}`;
```
erikap: Nitpicking:  adding `trim()` to clean trailing whitespaces if `agendaName` is empty

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -102,6 +104,48 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
     );
   }
 
+  @bind
+  async canBeApproved() {
+    const agendaitems = await this.args.currentAgenda.agendaitems;
+    const approvedAgendaitems = agendaitems.filter((agendaitem) => [CONSTANTS.ACCEPTANCE_STATUSSES.OK].includes(agendaitem.get('formallyOk')));
```
erikap: No need to use `get()`

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -102,6 +104,48 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
...
+    return approvedAgendaitems.length === agendaitems.length;
+  }
+
+  async allAgendaitemsNotOk() {
+    const agendaitems = await this.args.currentAgenda.agendaitems;
+    return agendaitems
+          .filter((agendaitem) => [CONSTANTS.ACCEPTANCE_STATUSSES.NOT_OK, CONSTANTS.ACCEPTANCE_STATUSSES.NOT_YET_OK].includes(agendaitem.get('formallyOk')))
```
erikap: No need to use `get()`

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -79,6 +81,14 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
     return this.latestPublicationActivity != null;
   }
 
+  @bind
+  async allAgendaitemsNotOk() {
+    const agendaitems = await this.args.currentAgenda.agendaitems;
+    return agendaitems
+          .filter((agendaitem) => [CONSTANTS.ACCEPTANCE_STATUSSES.NOT_OK, CONSTANTS.ACCEPTANCE_STATUSSES.NOT_YET_OK].includes(agendaitem.get('formallyOk')))
```
erikap: No need to use `get()`

app/helpers/agenda-name.js
```diff
@@ -0,0 +1,11 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(async function agendaName([agenda]) {
+  const isDesignAgenda = await (await agenda.status).isDesignAgenda;
```
erikap: `isDesignAgenda` of `agendastatus` should not be awaited?

app/helpers/agenda-name.js
```diff
@@ -0,0 +1,11 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(async function agendaName([agenda]) {
```
erikap: Wondering whether templates automatically update (without the need for the `await` helper) when defining `agendaName` as an (async) getter on the `agenda` model like
```javascript
get agendaName() {
  let prefix = this.status.get('isDesignAgenda') ? 'Ontwerpagenda' : 'Agenda';
  let name = this.serialNumber || '';
  return `${prefix} ${name}`.trim();
}
```

I know it's an anti-pattern we would like to avoid, but this case seems 'legimate' to me. It's a very simple getter, it doesn't introduce additional network requests compared to the helper and it makes the template simpler since there is no need to use `await`.

erikap (changes requested):

app/utils/bind.js
sergiofenoll: This util is used to ensure a function's `this` is the one we expect when using said function as a helper in a template.

A number of the computed properties that got refactored make use of services or the model itself, but these properties were only ever used in templates. E.g. `agendaitem.canBeApproved`, which got refactored [here](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/f0ebecfdd73ec2c25a1088fc0a12a590230e1583/app/components/agenda/agenda-header/agenda-version-actions.js#L108).

My approach for refactoring these is to turn them into helpers, so we can use them in templates. Since we have [ember-functions-as-helper-polyfill](https://github.com/ember-polyfills/ember-functions-as-helper-polyfill) installed, I went for local helpers where possible (that way I can make use of services, args, etc. that the component/controller might have access to). But when a local function is passed to a template as a helper, it loses its context and `this` no longer "points" to the component/controller class.

I took the solution from [this issue](https://github.com/ember-polyfills/ember-functions-as-helper-polyfill/issues/103) in the polyfill repo, where it's suggested to explicitly bind the functions-as-helpers like [so](https://github.com/statelyai/xstate/pull/3061/files#diff-6552eacdc0f4378c00e24a96877df8352eeb0db006852dfdcf511f59d9b781aeR731).

If this is undesired we can pass the needed context to the helpers (pass in the model, services, ... as arguments) which I tried at first but that looked kind of odd to me.

Maybe some/all of the computed-properties-turned-helpers can be rethought using `@tracked` properties and `@task`s to load in data.

ValenberghsSven:
Basis wijzigen naar development, normaal gebeurt dit automatisch maar door conflicten niet gebeurt denk ik.

app/components/agenda/agenda-header/agenda-action-popup-agendaitems.js
```diff
@@ -0,0 +1,20 @@
...
+  async newsletterInfo(agendaitem) {
+    return await this.store.queryOne('newsletter-info', {
+      'filter[agenda-item-treatment][agendaitem][:id:]': agendaitem.id,
+    });
+  }
```
ValenberghsSven: tbh, the only reason I initially included `newsletter-info` here was because I wanted this component to resemble the sidebar component.
All of this code for showing an icon in case there is a `newsletter-info` and  `inNewsletter` is `false`.
The current code change doesn't actually work (but it may have been broken before this), so I would opt to just remove this code + the relevant block in the template.

app/components/agenda/agenda-header/agenda-action-popup-agendaitems.hbs
```diff
@@ -10,12 +10,12 @@
...
                 <Auk::Icon @name="calendar-new" @skin="muted" />
               {{/if}}
               {{!-- This does not seem to work properly without the #let, newsletterInfo.inNewsletter is undefined --}}
-              {{#if (await agendaitem.newsletterInfo)}}
-                {{#let (await agendaitem.newsletterInfo) as |newsletterInfo|}}
+              {{#if (await (this.newsletterInfo agendaitem))}}
+                {{#let (await (this.newsletterInfo agendaitem)) as |newsletterInfo|}}
                   {{#if (eq (await newsletterInfo.inNewsletter) false)}}
                     <Auk::Icon @name="not-visible" @skin="muted" />
                   {{/if}}
```
ValenberghsSven: see comment in component, I would just remove this entire block `{{#if (await agendaitem.newsletterInfo)}}`.
It doesn't really add much to the component and just adds to the complexity

ValenberghsSven: There also was never a design for this component (I think), so this is kinda "off the cuff"

app/components/agenda/agenda-header/agenda-actions.hbs
```diff
@@ -5,7 +5,7 @@
   data-test-agenda-header-show-options={{true}}
   as |Menu|
 >
-  {{#if this.canEditDesignAgenda}}
+  {{#if (and this.currentSession.isEditor @currentAgenda.status.isDesignAgenda)}}
```
ValenberghsSven: I feel a bit of inconsistency here.
in app/components/agenda/agenda-overview.hbs you changed:
 `{{#if (and this.currentSession.isEditor this.isDesignAgenda)}}` to `this.canEdit`
In this case you did the opposite and got rid of a getter.
Any context reasons that I may be missing? Perhaps something didn't work as expected?

app/models/meeting.js
```diff
@@ -1,96 +1,41 @@
...
-  }),
-
-  latestAgenda: computed('agendas.[]', function() {
-    return PromiseObject.create({
-      promise: this.get('agendas').then((agendas) => {
-        const sortedAgendas = agendas.sortBy('agendaName').reverse();
-        return sortedAgendas.get('firstObject');
-      }),
-    });
-  }),
```
ValenberghsSven, line 85: This is still seems used in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3427-refactoren-en-verwijderen-van-computed-properties-op-ember-data-models/app/pods/newsletters/template.hbs#L75

app/pods/cases/case/subcases/subcase/documents/controller.js
```diff
@@ -145,16 +145,27 @@ export default class CasesCaseSubcasesSubcaseDocumentsController extends Control
...
+            const agendaitem = agendaitems.objectAt(index);
+            const agenda = await agendaitem.agenda;
+
+            if (agenda.id === latestAgenda.id) {
+              await restorePiecesFromPreviousAgendaitem(agendaitem, documentContainer);
+              // TODO: make sure we're not loading stale cache
+              await agendaitem.hasMany('pieces').reload();
+              break;
+            }
+          }
```
ValenberghsSven: tested this and it still works.
can go with this for now, might want to make this less complex later.

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -86,8 +107,8 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
       isSessionClosable &&
       this.currentSession.isAdmin &&
       this.currentAgendaIsLatest &&
-      this.args.currentAgenda.isDesignAgenda
-    );
+      this.isDesignAgenda
+    )
```
ValenberghsSven, line 110: bit out of scope, but this getter `canReopenPreviousAgenda` has 1 incorrect check:
`!this.currentSession.isFinal`. This should be `!this.args.meeting.isFinal`

Would also prefer `isMeetingClosable`  (session was used for both meeting and current user session in legacy code, snippets of this remain)

If not done here, no worries, will do this on DEV or in small PR afterwards

ValenberghsSven (changes requested):
nearly there

app/components/agenda/agenda-header/agenda-actions.hbs
sergiofenoll: This one's just me being a bit sloppy and getting confused by ember autotracking, first I thought it would not work with the getter so I removed it here, then later I found out it does actually work, so I added it elsewhere. thanks for the catch

ValenberghsSven (approved):

PR #1369, opened 2022-06-01
KAS-3453 make relation read-only, update inverse

ValenberghsSven:
### :warning:  WARNING! branch started from production

# HOTFIX https://kanselarij.atlassian.net/browse/KAS-3453

### What has changed:
models/agenda-item-treatments.js:
- Make relation read-only by not serializing it.

agenda/agendaitems/agendaitem/decisions/controller.js:
- instead of creating a new treatment with a newsletter, we call the new method in the newsletter service.

agenda/agendaitems/agendaitem/news-item/route.js:
- Model was dependent on a possible stale relation, use query instead.

print-overviews/decisions/agendaitems/controller.js:
- instead of creating a new treatment with a newsletter, we call the new method in the newsletter service.
(connecting existing newsletter was actually missing here, assuming this route is not really used to add treatments)

services/newsletter-service.js:
- linter
- new method that checks if there is an existing newsletter for the agendaitem in params. If it exists, we refresh its treatments and save the newsletter with the list of treatments

--

Jenkins PR will fail, use custom builds against production backend

1st build
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/242/

### DIFF
--- a/app/models/agenda-item-treatment.js
+++ b/app/models/agenda-item-treatment.js
@@ -15,7 +15,7 @@ export default class AgendaItemTreatment extends Model {
   @belongsTo('agendaitem') agendaitem;
   @belongsTo('subcase') subcase;
   @belongsTo('piece') report;
-  @belongsTo('newsletter-info') newsletterInfo;
+  @belongsTo('newsletter-info', { serialize: false }) newsletterInfo;
   @belongsTo('decision-result-code', { inverse: null }) decisionResultCode;
   @hasMany('publication-flow') publicationFlows;
   @hasMany('sign-flow') signFlows;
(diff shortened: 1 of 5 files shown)


sergiofenoll:
Repro stappen uit de ticket gevolgd, resultaat werkt zoals verwacht.
Code ziet er goed uit.

sergiofenoll (approved):
Repro stappen uit de ticket gevolgd, resultaat werkt zoals verwacht.
Code ziet er goed uit.

ValenberghsSven:
Ik ga deze al mergen zodat ik lokaal eens de production tag kan uittesten

PR #1370, opened 2022-06-02
KAS-3419 Link existing publication (not via MR) to agendaitem document afterwards

erikap:
Add support to link a publication 'not via MR' to a document on the agenda afterwards.

Search box for publications now also returns 'publications not via MR'. In case a publication 'not via MR' is selected, the publication gets relinked to the case/agenda-item-treatment the document belongs to. The dummy data (case/agenda-item-treatment) [created before](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/publication-service.js#L70-L79) is removed.

Also updated to condition to classify a publication as 'via MR'/'not via MR'. Instead of checking for the presence of a subcase, the presence of an agendaitem is now checked, which is more correct semantically.

### DIFF
--- a/app/components/publications/batch-documents-publication-modal.js
+++ b/app/components/publications/batch-documents-publication-modal.js
@@ -104,15 +104,40 @@ export default class PublicationsBatchDocumentsPublicationModalComponent extends
 
   @action
   async linkPublicationFlow(piece, publicationFlow) {
+    let publicationFlowHasChanged = false;
+
     const regulationType = await publicationFlow.regulationType;
     if (!regulationType) {
       const regulationTypeFromDocument =
         await this.getRegulationTypeThroughReferenceDocument(piece);
       if (regulationTypeFromDocument) {
         publicationFlow.regulationType = regulationTypeFromDocument;
-        await publicationFlow.save();
+        publicationFlowHasChanged = true;
       }
     }
+
+    // Ensure the publication-flow is linked to the same agenda-item-treatment as
+    // the document. This might differ if the publication-flow has been created before
+    // the subcase was handled on an agenda and a dummy agenda-item-treatment was
+    // created back then. A dummy agenda-item-treatment can be recognized because it
+    // doesn't have a related agendaitem in Kaleidos.
+    const agendaItemTreatment = await publicationFlow.agendaItemTreatment;
+    if (agendaItemTreatment.id != this.agendaItemTreatment.id) {
+      const agendaitem = await this.store.queryOne('agendaitem', {
+        'filter[treatments][:id:]': agendaItemTreatment.id
+      });
+      if (!agendaitem) {
+        publicationFlow.agendaItemTreatment = this.agendaItemTreatment;
+        publicationFlowHasChanged = true;
+        // destroy the dummy agenda-item-treatment
+        await agendaItemTreatment.destroyRecord();
+      }
+    }
+
+    if (publicationFlowHasChanged) {
+      await publicationFlow.save();
+    }
+
     piece.publicationFlow = publicationFlow;
     await piece.save();
   }
--- a/app/components/publications/batch-documents-publication-modal.js
+++ b/app/components/publications/batch-documents-publication-modal.js
@@ -128,13 +128,14 @@ export default class PublicationsBatchDocumentsPublicationModalComponent extends
       } else {
         // Publication-flow is already linked to an agenda-item-treatment that has been
         // handled on an agenda. We cannot relink the publication-flow in that case.
-        // In practice, this scenario will not occur since they are not able to select
-        // such a publication-flow in the publication-flow-selector.
+        // In practice this scenario will only occur in concurrency scenarios where
+        // multiple users try to link the same publication at the same time to
+        // different cases.
         this.toaster.error(
           this.intl.t('unable-to-relink-publication-flow'),
           this.intl.t('warning-title')
         );
-        throw Error('Unable to relink publication flow');
+        return;
       }
     }
 
--- a/app/components/publications/batch-documents-publication-modal.js
+++ b/app/components/publications/batch-documents-publication-modal.js
@@ -12,6 +12,8 @@ import { task } from 'ember-concurrency';
 export default class PublicationsBatchDocumentsPublicationModalComponent extends Component {
   @service store;
   @service publicationService;
+  @service toaster;
+  @service intl;
 
   @tracked isOpenNewPublicationModal = false;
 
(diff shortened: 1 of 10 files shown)


erikap:
Failing tests seem related to this PR. Tried to fix but didn't succeed yet :worried: 

I assume the problem is the search operation in [publications/publication-flow-selector.js](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1370/files#diff-5ade19de290b9b9fc2e4841596fe49cb0f3baf6da88d8f2245f7ae94c25b298f) being split into 2 queries instead of one. Tried to update test flow in [cypress/integration/unit/publication-via-MR.spec.js](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3419-link-publication-to-document-afterwards/cypress/integration/unit/publication-via-MR.spec.js#L222-L225) to

```javascript
    cy.intercept({
      method: 'GET',
      url: '/publication-flows',
      query: {
        'filter[case][:id:]': '*',
        'filter[identification][id-name]': publicationNumber2,
      },
    }).as('getFilteredByCaseAndId');
    cy.intercept({
      method: 'GET',
      url: '/publication-flows',
      query: {
        'filter[agenda-item-treatment][:has-no:agendaitem]': 'yes',
        'filter[identification][id-name]': publicationNumber2,
      },
    }).as('getFilteredNotViaMRAndId');
    cy.intercept('PATCH', '/pieces/**').as('patchPieces');
    cy.get(dependency.emberPowerSelect.searchInput).type(publicationNumber2)
      .wait(['@getFilteredByCaseAndId', '@getFilteredNotViaMRAndId']);
```

but the `wait(['@getFilteredByCaseAndId', '@getFilteredNotViaMRAndId'])` never matches. Can you help @ValenberghsSven ?

ValenberghsSven:
I'll look at this

ValenberghsSven:
Cypress is not picky, defining 2 unique calls can be done with plenty of wildcards.
```
cy.intercept('GET', `/publication-flows**?filter**case**${publicationNumber2}**`).as('getFilteredIdCase');
cy.intercept('GET', `/publication-flows**?filter**agenda*item*treatment**${publicationNumber2}**`).as('getFilteredIdTreatment');
```
and then chain awaited
```
cy.get(dependency.emberPowerSelect.searchInput).type(publicationNumber2)
.wait('@getFilteredIdCase')
.wait('@getFilteredIdTreatment');
```

The main issue was that we sometimes found the loading option, but we can ignore that option since it has a different class then a normal selectable option (the 2 `cy.wait` were not always enough..)
`cy.get(dependency.emberPowerSelect.optionLoadingMessage).should('not.exist');`

Part of the tests failing was also that we can click the `edit` button on the case panel while the `initFields` action is still running, and we are hitting a `null.localIdentifier` in `openEditingPanel()`.
So an additional fix is to either wait longer in cypress (checking the url changing wasn't enough), so opted to alter the code a bit, so we wait for the `initFields` before showing the edit button.

app/components/publications/publication-flow-selector.js
```diff
@@ -16,23 +16,54 @@ export default class PublicationsPublicationFlowSelectorComponent extends Compon
...
     });
-    return publicationFlows;
+    const searchNotViaCouncilOfMinisters = this.store.query('publication-flow', {
+      filter: filterNotViaCouncilOfMinisters,
+      'page[size]': PAGE_SIZE.PUBLICATION_FLOWS,
+      sort: 'identification.id-name',
+      include: 'identification',
+    });
+
+    const [publicationsByCase, publicationsNotViaCouncilOfMinisters] = yield all([
```
ghost: Wat is de reden dat nu EmberConcurrency.all wordt gebruikt i.p.v. Promise.all?

app/components/publications/batch-documents-publication-modal.js
```diff
@@ -104,15 +106,52 @@ export default class PublicationsBatchDocumentsPublicationModalComponent extends
...
+      } else {
+        // Publication-flow is already linked to an agenda-item-treatment that has been
+        // handled on an agenda. We cannot relink the publication-flow in that case.
+        // In practice, this scenario will not occur since they are not able to select
+        // such a publication-flow in the publication-flow-selector.
+        this.toaster.error(
+          this.intl.t('unable-to-relink-publication-flow'),
+          this.intl.t('warning-title')
+        );
+        throw Error('Unable to relink publication flow');
```
ValenberghsSven: In order to test his, I managed to hit his with concurrency, both browsers already searched the same number and tried to link 2 docs from different cases to the same publicationflow and the second one failed so it works.
Not sure why we are throwing an error though, nothing really catches this error so it just ends up as an `uncaught exception`.
Could just return to exit the method?

app/services/publication-service.js
```diff
@@ -190,9 +190,10 @@ export default class PublicationService extends Service {
   }
 
   async getIsViaCouncilOfMinisters(publicationFlow) {
-    const _case = await publicationFlow.case;
-    const subcases = await _case.subcases;
-    return !!subcases.length;
+    const agendaitem = await this.store.queryOne('agendaitem', {
+      'filter[treatments][publication-flows][:id:]': publicationFlow.id,
+    });
+    return agendaitem != null;
```
ValenberghsSven: Oddly enough, this doesn't work.
the `queryOne` yields null even though I just linked a publication so it should return something.
`query` does return results

I think we are hitting a cache issue here:
`queryOne` only worked after I rebooted cache and resource.
I'm guessing the path to cache of agendaitem/treatments wasn't invalidated in this process so we keep hitting the stale one. Or something like that, kinda difficult to pinpoint
Using curl and adding something to the call also works to get results while the current call gets null

edit: I saw that there was no change to the "source", so newly linked publications still showed 'niet via MR'

ValenberghsSven (changes requested):

app/services/publication-service.js
ValenberghsSven: "I'm guessing the path to cache of agendaitem/treatments wasn't invalidated in this process so we keep hitting the stale one"

This is kinda the same issue we have with the yggdrasil government cache?

ValenberghsSven: I suggest using the same query as you do in the `linkPublicationFlow`
```
const agendaItemTreatment = await publicationFlow.agendaItemTreatment;
const agendaitem = await this.store.queryOne('agendaitem', {
  'filter[treatments][:id:]': agendaItemTreatment.id
});
```

app/components/publications/publication-flow-selector.js
erikap: No specific reason since the passed promises aren't ember-concurrency tasks. Replaced with Promise.all().

app/components/publications/batch-documents-publication-modal.js
erikap: Nice catch. It was throwing an error because I didn't expect to hit this case. Updated the inline documentation with the concurrency scenario and replaced the error-throwing with a regular return to exit the method :+1:

app/services/publication-service.js
erikap: Changed the query to the one you proposed and documented the reasoning inline :+1: 

Sounds a lot like the Yggdrasil government cache issue. May be related to the fuzzy relation-cardinality between agendaitem and agenda-item-treatment (defined as `has-one` in mu-cl-resources, but `has-many` in data)?|
I hope the issue will  be "automatically" resolved by the upcoming data migration stories where we're going to split agenda-item-treatment and decision-activities which will result in a simple one-to-one relation between agendaitem and agenda-item-treatment. I would at least pause further investigation on the cache issue until that data migration is done.

ValenberghsSven: "simple one-to-one relation between agendaitem and agenda-item-treatment"
this will not be the case since multiple agenda versions have different agendaitems that are linked to the same treatment, so it will still be a many-to-one. But since there will only ever be 1 treatment and multiple decisions, we wont have to deal with a many-to-many anymore

ValenberghsSven:
jenkins ok buiten 1 gekende flaky deze keer, ik dacht dat die opgelost was..

PR #1380, opened 2022-06-10
KAS-3466:  Integreren van nieuwe editor in kort bestek

sergiofenoll:
Bumps rdfa-editor to v0.57.0

Also adds peer deps which are needed for it to work right now due to a bug.


erikap:
@sergiofenoll Can you try to upgrade to the latest released version? https://github.com/lblod/ember-rdfa-editor/tags
I hope that also solves the issue with the peer deps.

sergiofenoll:
@erikap ~~latest version does indeed fix that issue, should've tried that before :grimacing: 
thanks!~~

scratch that, I bumped to version 0.59.1 but forgot to uninstall the peer deps, so it only worked because they were still installed. uninstalling the peer deps still gives the error.

i'll bump the version but still leave in the peer deps.

as an aside, I tried setting it up locally (`npm link`ing a local copy of the `@lblod/ember-rfa-editor` package) and there I don't get the error, which makes debuggin the problem pretty annoying

ValenberghsSven (commented):
With the new editor changes to `nbsp`, will it have any side effects?
See comments in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3466-bump-rdfa-editor/app/components/news-item/edit-item.js#L49 to replace some `nbsp`

sergiofenoll:
> With the new editor changes to `nbsp`, will it have any side effects? See comments in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3466-bump-rdfa-editor/app/components/news-item/edit-item.js#L49 to replace some `nbsp`

@ValenberghsSven I would assume not. The PR I linked in the test's comments seems to be about the editor normalizing user inputted whitespace. The comment you mention seems to be about the editor inserting `&nbsp;`s, but I assume that's a different mechanism that remains unchanged. I haven't looked at how the editor works in-depth though, so take this with a grain of salt.

erikap:
I've created a new story to investigate whether the `&nbsp;` workaround is still required in the latest editor version such that it doesn't block this PR. Going to merge this one as users have tested/accepted the editor upgrade on usertest including the workaround.

PR #1387, opened 2022-06-21
KAS-3488 Mandataris titel is soms undefined in dropdown

sergiofenoll:
Display the mandatee's role label if a mandatee has no title. Should only happen for some older Minister-president that lack the `dct:title` property. In those cases we will just display `Minister-president` (instead of the usual `Minister-president van de Vlaamse Regering`.


erikap (approved):
We should get rid of those computeds, but out of scope for this story. Good enough for now :+1:

PR #1389, opened 2022-06-22
Feature/kas 3494 usertest kort bestek editor

sergiofenoll:
Updates editor in Kort Bestek tab of agendapunt overview to match the styling of usertest.

PR might be a bit annoying to read because a couple of file get changes & are moved, but the changes and moving occur in separate commits, so it might be easier to go through it commit per commit.

Where possible, Auk components are replaced with Au components. Two notable exceptions:
- `AuToolbar` does not seem to work in the same way `Auk::Toolbar` does when it comes to taking up space and positioning. At a first glance, positioning items on the left or right side didn't seem to be supported, so I opted to keep the `Auk::Toolbar`s here
- `AuToggleSwitch` expects to receive a label to set up a wrapping `AuLabel` by itself, but it follows a different style than what we're using. There doesn't seem to be an option to disable this behaviour though, so the component would end up creating extra empty tags. I opted for sticking to `Auk::Toggle` instead.


erikap (approved):
Approving this one purely based on the urgently requested layout changes. Ember-wise these components/route need further clean up (e.g. reassigning model in the controller is an anti-pattern, improve DDAU for saving, ...). I'm going to make a new story that hopefully can be included this sprint, since the current progress looks good.

erikap:
Updated base branch to `acceptance` since this feature should be released separately from the other changes on dev

erikap:
Changing base branch did not work since development branch has been merged in this feature-branch (thx @ValenberghsSven for noticing). Going to create a new branch, starting from acceptance, and cherrypick the relevant commits for this feature.

erikap:
Closed in favor of #1402

PR #1391, opened 2022-06-23
KAS-3489 add search only on urgent publications option

ValenberghsSven:
# :warning: this PR is mutually exclusive with https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1392 :warning: 
Don't merge both

In this PR, Added the option to search on urgent publications only.

Dependent on https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/292

### DIFF
--- a/app/routes/search/publication-flows.js
+++ b/app/routes/search/publication-flows.js
@@ -11,6 +11,7 @@ import {
 } from 'frontend-kaleidos/utils/publication-auk';
 import { PAGE_SIZE } from 'frontend-kaleidos/config/config';
 import { warn } from '@ember/debug';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
 
 export default class PublicationFlowSearchRoute extends Route {
   @service store;
@@ -25,6 +26,10 @@ export default class PublicationFlowSearchRoute extends Route {
     publicationStatusIds: {
       refreshModel: true,
       as: 'statussen',
+    },
+    urgentOnly: {
+      refreshModel: true,
+      as: 'dringend',
     },
     page: {
       refreshModel: true,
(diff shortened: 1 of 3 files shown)


ValenberghsSven:
Custom jenkins run.
The testdata zip has not been updated yet, so we are still running against the old index.
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/251/

ValenberghsSven:
clicked the wrong button and closed by accident....

app/routes/search/publication-flows.js
```diff
@@ -109,6 +118,10 @@ export default class PublicationFlowSearchRoute extends Route {
       filter[':terms:statusId'] = params.publicationStatusIds;
     }
 
+    if (params.urgentOnly) {
+      filter[':terms:urgencyLevelId'] = this.urgencyLevelSpeed.id;
```
sergiofenoll: I think the filter here needs to be `:terms:urgencyLevelId.keyword`. Locally, I can't get the search to only return results marked as urgent. After appending `.keyword` to the filter, it does seem to work as expected.

I assume it's related to how Elasticsearch searches string fields. See this conversation in chat for reference: https://chat.semte.ch/channel/mu-semtech?msg=KEHQtfdWyLWEHj7f5

sergiofenoll (changes requested):

app/routes/search/publication-flows.js
ValenberghsSven: hmmm, I only changed that near the end, may have missed it.
But then we don't have to use `terms`, `filter['urgencyLevelId'] = this.urgencyLevelSpeed.id;` should work?

sergiofenoll (approved):

erikap (approved):

PR #1392, opened 2022-06-23
Feature/kas 3489 urgency search both values

ValenberghsSven:
# :warning: this PR is mutually exclusive with https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1391 :warning: 
Don't merge both

In this PR, added the option to search on urgency-level value
Due to bad data, this option does not work properly without migrations
most data has no `urgency-level` by default, urgent publications have `urgency-level` 'spoedprocedure', very few publications have `urgency-level` "standaard" (when you edit an urgent publication to no longer be urgent, we replace the `urgency-level` instead of deleting it)

This leaves us with 3 states, so searching for `urgency-level` "standaard" will yield far less results then expected

Dependent on https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/292

### DIFF
--- a/app/controllers/search/publication-flows.js
+++ b/app/controllers/search/publication-flows.js
@@ -13,6 +13,9 @@ export default class PublicationFlowSearchController extends Controller {
     publicationStatusIds: {
       type: 'array',
     },
+    urgencyLevelIds: {
+      type: 'array',
+    },
     page: {
       type: 'number',
     },
@@ -31,6 +34,7 @@ export default class PublicationFlowSearchController extends Controller {
   @tracked sort;
   @tracked regulationTypeIds = [];
   @tracked publicationStatusIds = [];
+  @tracked urgencyLevelIds = [];
 
   constructor() {
     super(...arguments);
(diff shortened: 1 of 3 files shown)


erikap:
Not sure what has been discussed in the standup, but according to me it has been an oversight in the past that publications don't have urgency level 'standaard' by default. Isn't it easier to fix the data in this story and make sure every publication has an urgency level (being 'standaard' or 'spoedprocedure') and newly created publications get 'standaard' by default?

ValenberghsSven:
> Not sure what has been discussed in the standup, but according to me it has been an oversight in the past that publications don't have urgency level 'standaard' by default. Isn't it easier to fix the data in this story and make sure every publication has an urgency level (being 'standaard' or 'spoedprocedure') and newly created publications get 'standaard' by default?

That is my question here exactly.

ValenberghsSven:
After Feedback from Johan. Searching only on "urgent" is fine.
For that option we don't need the `access-level` on every publication.
We could still add default levels later, but the "urgent only" option will just keep working whether we do or not.

PR #1393, opened 2022-06-23
KAS-3495 add transition to publication search for ovrb users

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3495

`isOvrb` was not used anywhere anymore since using roles, so decided to change this to make the conditional logic easier.

If not, we would need to check that the user was `isOvrb` and not also `isAdmin`

### DIFF
--- a/app/routes/search/index.js
+++ b/app/routes/search/index.js
@@ -5,7 +5,7 @@ export default class IndexSearchRoute extends Route {
   @service currentSession;
   @service router
 
-  beforeModel() {
+  beforeModel(transition) {
     /* It is assumed here that explicitly navigating to the index-route
      * expresses a desire to "reset", to "start a new search-session".
      * Therefore all parameters except filters -such as "type" of agenda-item- are cleared.
@@ -16,15 +16,24 @@ export default class IndexSearchRoute extends Route {
      * of the "search" route, even though the queryParams are marked `refreshModel: true`.
      * As a result "searchTextBuffer" doesn't get cleared.
      */
+    // We have to mention the arrays or we run into the problem that clicking multiple times
+    // results in these arrays being transformed to strings "[]"
     const queryParams = {
       searchText: null,
       mandatees: null,
       dateFrom: null,
       dateTo: null,
       page: 0,
+      regulationTypeIds: [],
+      publicationStatusIds: [],
     }
     // ovrb users get directed to publication search
     if (this.currentSession.isOvrb) {
+      // clicking search route while filters are selected will clear filters but not he checkboxes
+      // This workaround blocks transitions to the same route to prevent this.
+      if (this.router.currentRouteName === 'search.publication-flows') {
+        transition.abort();
+      }
       return this.router.transitionTo('search.publication-flows', {
         queryParams: queryParams,
       });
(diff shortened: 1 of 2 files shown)


sergiofenoll:
Looks good, only thing I found is that if you click the Zoeken tab twice while the page still loading as an OVRB user you will encounter an error.

Error: 
```
> Uncaught (in promise) TypeError: params.regulationTypeIds.join is not a function
```
Points to this line: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/routes/search/publication-flows.js#L102

Printing the params the model hook is getting, I get this result:
```
// First click
{
  "searchText": null,
  "mandatees": null,
  "dateFrom": null,
  "dateTo": null,
  "publicationDateTypeKey": "decisionDate",
  "regulationTypeIds": [], // This is an array, which has a .join() method
  "publicationStatusIds": [],
  "page": 0,
  "size": 20,
  "sort": "-opening-date"
}
// Second click
{
  "searchText": null,
  "mandatees": null,
  "dateFrom": null,
  "dateTo": null,
  "publicationDateTypeKey": "decisionDate",
  "regulationTypeIds": "[]", // This is a string now. doesn't have a .join() method
  "publicationStatusIds": "[]",
  "page": 0,
  "size": 20,
  "sort": "-opening-date"
}
```

No idea what the underlying cause is, I would guess that Ember does something weird when it comes to serializing query params while in the middle of a transition?

But we might want to handle this case (perhaps by checking that `params.regulationTypeIds` and `params.publicationStatusIds` have a `.join()` method) because clicking on the Zoeken tab take a little while and I imagine users might try to click it twice thinking nothing had happened.

ValenberghsSven:
When it occurs, we see a wrong template being used
image

ValenberghsSven:
i'm guessing there could a problem with the getter `isSearchingPublicationFlows` when switching fast
 `return this.router.currentRouteName === 'search.publication-flows';`
 
 Edit: nope, it happens before the getter is called

ValenberghsSven:
Changing queryParams in the `search/index` route
```
    const queryParams = {
      searchText: null,
      mandatees: null,
      dateFrom: null,
      dateTo: null,
      page: 0,
      regulationTypeIds: [],
      publicationStatusIds: [],
    }
```
Works to solve the error, but messes with the filter
when op pub search, filter on a type (get limited results)
Click "zoeken"
The filter is reset, but the checkbox is not (all results, type still checked)

This could be countered by blocking the transition to "search.publication-flow" while already on route "search.publication-flow" in the beforeModel hook of the search/index route
```
if (this.router.currentRouteName === 'search.publication-flows') {
    transition.abort();
}
return this.router.transitionTo('search.publication-flows', {
   queryParams: queryParams,
});
```

ValenberghsSven:
Fixed the bug,
fixing the bug introduces a new problem, where checkboxes are not cleared but the filter is when clicking the search route button.
To counter it, prevented the transition from `search.publication-flow` when already in `search.publication-flow` for ovrb users. (it's not a problem for other profiles because they go to `search.agenda-items` by default.

erikap (approved):

PR #1394, opened 2022-06-23
KAS-3499 - cypress/KAS-3499-link-not-via-MR-publication-to-MR

MathiasZabot:


ValenberghsSven (approved):
Tests are ok, setup needs to be put in zip but waiting on big refactor to be merged first.
Could merge as-is and do this later

PR #1396, opened 2022-06-23
KAS-3501 change default sort field for late-in-bs tab

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3501

- sort on `target-end-date` instead of `due-date` :heavy_check_mark: 
- keep filtering on "not final", was already the case :white_check_mark: 
- oldest first, was already the case :white_check_mark: 

image


erikap (changes requested):
Filter must also be updated to target-end-date instead of due-date

ValenberghsSven:
> Filter must also be updated to target-end-date instead of due-date

Yes, an oversight on my part

erikap (approved):

PR #1399, opened 2022-06-27
Feature/kas 3460 improve agenda search ux

sergiofenoll:
Adds inline loading, so that the search input does not get re-rendered while searching (and the user does not lose focus).
Adds buttons to explicitly start the search.
Keeps support for debounced search-as-you-type (wasn't sure to me if that was expected, but it seems useful to me), but typing Enter or clicking the search button performs the search immediately.
Uses `AuButton` for the search buttons instead of `Auk::Button`

Note: the icon size in `AuButton` looks somewhat small. Perhaps we want to make it larger with CSS overrides. Or maybe we could add an `@iconSize` argument to the `AuButton` component in https://github.com/appuniversum/ember-appuniversum (an `AuIcon` with `@size="large"` looks okay in our buttons).
image


app/routes/agenda/agendaitems.js
```diff
@@ -117,6 +117,24 @@ export default class AgendaAgendaitemsRoute extends Route {
...
+  loading(transition) {
+    // eslint-disable-next-line ember/no-controller-access-in-routes
+    const controller = this.controllerFor(this.routeName);
+    controller.isLoadingModel = true;
+    transition.promise.finally(() => {
+      controller.isLoadingModel = false;
+    });
+
+   // only bubble loading event when transitioning between tabs
+   // to enable loading template to be shown
```
erikap: Is this comment up-to-date?  'tabs' are not applicable here? 
I assume we want to say "don't bubble loading event if only query params change"?

app/templates/agenda/agendaitems.hbs
```diff
@@ -33,6 +37,7 @@
       <Agenda::AgendaitemSearch
         @searchText={{readonly this.filter}}
         @onSearch={{this.searchAgendaitems}}
+        @fullwidth={{true}}
```
erikap: `fullwidth` doesn't clearly reflect the intention of the argument for me. Maybe name it in line with the argument of `AuButton`?  `hideButtonText` or something like that?

app/routes/agenda/agendaitems.js
sergiofenoll: I just copy-pasted this snippet to be honest, but I think it still applies? We still want to show the template-level loading state when switching between the Overzicht, Detail, Vergelijken, and Documenten tab, right?

But the comment is somewhat confusing and really just an artifact of being copied around, I will try to make it a bit clearer

erikap:
> Note: the icon size in `AuButton` looks somewhat small. Perhaps we want to make it larger with CSS overrides. Or maybe we could add an `@iconSize` argument to the `AuButton` component in https://github.com/appuniversum/ember-appuniversum (an `AuIcon` with `@size="large"` looks okay in our buttons). image

@brenner-company What's your opinion about the icon-size? What's the best fix? Do we create a follow-up story for this or block this story?

brenner-company:
@erikap @sergiofenoll I'm inclined to switching the button back to the auk-version for the time being because:
- The component is, as you are stating, missing some styling overrides like icon size (I wouldn't go for 'large' size assuming it should be correctly sized by default), focus color, button height, etc.
- The component is not 100% complete for use within Kaleidos regarding available skins (missing 'success', etc.).

Normally I would pick this up along my other work but I'm on my last day before a rather long vacation 🏖. So a quick follow-up is not possible from my side I'm afraid.

brenner-company:
@sergiofenoll Thank you for you initiative by the way in using the components from `ember-appuniversum` 😊

erikap (approved):

PR #1405, opened 2022-06-30
Feature/kas 3373 decision activity

ValenberghsSven:
WIP,
easier to check all changes and prepare the merge


app/components/agenda/agenda-detail/sidebar-item.js
```diff
@@ -78,8 +78,7 @@ export default class SidebarItem extends Component {
 
   @task
   *loadNewsletterVisibility() {
-    const treatments = yield this.args.agendaitem.treatments;
-    const treatment = treatments.firstObject;
+    const treatment = yield this.args.agendaitem.treatment;
     if (treatment) { // TODO: this is only the case for the first item of the agenda (approval, older data)
       const newsletterInfo = yield treatment.newsletterInfo;
       if (newsletterInfo) {
```
ValenberghsSven: changing this to use `treatment?.newsletterInfo` and replacing the todo with comments why.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -119,7 +119,7 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
       // Reloading some relationships of agendaitem most likely to be changed by concurrency
       yield agendaitem.reload();
       yield agendaitem.hasMany('pieces').reload();
-      yield agendaitem.hasMany('treatments').reload();
+      yield agendaitem.belongsTo('treatment').reload();
```
ValenberghsSven: removing this, the relation can no longer be changed with concurrency.

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -193,7 +193,7 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
       // Reloading some relationships of agendaitem most likely to be changed by concurrency
       yield agendaitem.reload();
       yield agendaitem.hasMany('pieces').reload();
-      yield agendaitem.hasMany('treatments').reload();
+      yield agendaitem.belongsTo('treatment').reload();
```
ValenberghsSven: same, no longer needed

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -5,6 +5,13 @@ import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 
+/**
+ * TODO; subcase/agendaitem arguments are merely used for determining confidentiality of added report file. Simplify this. Maybe just determine subcase from decisionActivity and work from there?
+ * @argument subcase; optional
+ * @argument agendaitem; optional
+ * @argument decisionActivity
+ * @argument {Boolean} isTableRow: other layout when in edit mode
+ */
```
ValenberghsSven: fixing this todo by using the subcase from decisionActivity.
Will need to verify that all treatments that have a subcase are also linked to the subcase.
Should be the case for recent data, but maybe not legacy.

ValenberghsSven: also `isTableRow` was added here, but no longer used, removing I guess.

app/components/documents/document-preview/signatures-tab.js
```diff
@@ -27,15 +27,15 @@ export default class DocumentsDocumentPreviewDetailsSignaturesTabComponent exten
...
   }
 
   @task
   *markOrUnmarkForSignature() {
     if (!this.signMarkingActivity) {
-      const treatments = yield this.agendaitem.treatments;
-      const agendaItemTreatment = treatments.firstObject;
-      yield this.args.markForSignature(this.args.piece, agendaItemTreatment);
+      const agendaItemTreatment = yield this.agendaitem.treatment;
+      const decisionActivity = yield agendaItemTreatment.decisionActivity;
```
ValenberghsSven, line 37: do we need optional `agendaItemTreatment` here? out of scope, look into this when signflow progress is made.

app/models/agenda-item-treatment.js
```diff
@@ -1,22 +1,12 @@
...
-    Any reads should be on the inverse "agendaitem.treatments" or with a query via id.
-    see method getLatestAgendaitemFromTreatment in publicationService.
-  */
-  @belongsTo('agendaitem') agendaitem;
-  @belongsTo('subcase') subcase;
-  @belongsTo('piece') report;
+  @belongsTo('decision-activity') decisionActivity;
+
+  // Merely hasMany because agenda-item can have different versions (one per agenda-version)
+  @hasMany('agendaitem', { serialize: false }) agendaitems;
```
ValenberghsSven, line 10: `serialize: false` is technically currently not needed, but we should keep it.
Why it's not needed right now:
There is no ability to save the `treatment`, meaning we don't risk overwrites.
Why it could be needed eventually:
If we ever save the treatment somewhere, we could have concurrency issue where person 1 has 2 agendaitems linked and approves the agenda, so now there are 3.
Person 2 still has those 2 agendaitems and changes something that would save `treatment` > the list is saved with 2, removing the third relation from the DB.

app/models/newsletter-info.js
```diff
@@ -19,8 +19,8 @@ export default ModelWithModifier.extend({
   publicationDocDate: attr('datetime'),
   remark: attr('string'),
 
-  agendaItemTreatment: hasMany('agenda-item-treatment', {
-    serialize: true, // on creation of the newsletter-info, multiple decisions (for now also treatments) might already exist.
+  agendaItemTreatment: belongsTo('agenda-item-treatment', {
+    serialize: true, // TODO: research if this still needs expliciting. belongsTo is normally serialized anyway (this used to be a hasMany)
```
ValenberghsSven: The treatment will no longer change after the initial creation so saving it is fine.

app/models/newsletter-info.js
```diff
@@ -37,12 +37,12 @@ export default ModelWithModifier.extend({
   // - Newsletter:NewsletterItem::ItemContent
   // - Agenda::Agendaitem::AgendaitemNewsItem
   // Refactor these usages and remove this computed property
-  newsletterProposal: computed('agendaItemTreatment.{[],@each.subcase}', async function() {
+  newsletterProposal: computed('agendaItemTreatment.decisionActivity.subcase', async function() {
     // eslint-disable-next-line ember/no-get
-    const treatments = await this.get('agendaItemTreatment');
-    const treatment = treatments.firstObject;
+    const treatment = await this.get('agendaItemTreatment');
```
ValenberghsSven, line 40: works, but we should make a util out of this (no `store` needed) soon and refactor to octane

app/models/piece.js
```diff
@@ -36,9 +36,9 @@ export default class Piece extends Model {
   @belongsTo('submission-activity', {
     serialize: false
   }) submissionActivity;
-  @belongsTo('agenda-item-treatment', {
-    inverse: null
-  }) treatment;
+  @belongsTo('decision-activity', {
+    inverse: null // TODO: figure out why `inverse: null` here. If no reason, remove.
```
ValenberghsSven: are you ever going to use `piece.decisionActivity`?
Having it `inverse: null` means that ember will never load `piece.decisionActivity` when you set `decisionActivity.report`
So a save of `piece` will never update `piece.decisionActivity`.
A relation that is not loaded will not be saved

With `inverse: null` removed and concurrency you could get:
Report is uploaded by person 1 (ember sets `piece.decisionActivity`, but that's fine)
Person 2 goes to the report (ember does not load the relation since we never call it)
Person 2 upload a bis (ember sets `piece.decisionActivity` and )
Person 1 didn't reload yet and changes the accesslevel of the first version and saves the piece. (ember will now push the `piece.decisionActivity` to the backend. This causes there to be 2 relation from `decisionActivity` to pieces in a one-to-one relation.
The use-case is a bit exotic I know, but there are probably other use cases I am missing.

I wrote this text, then tested it

image

2 predicates for a one-to-one relation

app/models/subcase.js
```diff
@@ -24,5 +24,5 @@ export default class Subcase extends ModelWithModifier {
   @hasMany('submission-activity', { serialize: false }) submissionActivities;
   @hasMany('piece') linkedPieces;
   @hasMany('mandatee') mandatees;
-  @hasMany('agenda-item-treatment', { inverse: null }) treatments;
+  @hasMany('decision-activity', { inverse: null }) decisionActivities; // TODO: document why "inverse: null" is here or remove
```
ValenberghsSven: same argument as in piece.
Mainly for concurrency.
Not having the `inverse:null` means that ember would be possibly updating this relation and saving it, perhaps overwriting things with concurrency.
would even suggest making this a read-only relation by not serializing the hasMany.
Tried to test a use-case for this, but ended up breaking something else :D 

use case: 
1 subcase:
person 1 proposes it for an agenda
person 2 changes something that causes a save in subcase without reloading (should trigger concurrency warning but doesn't anymore?)
The connection between `agenda-activity` and `subcase` is gone (expected this with `decisionActivity` but is not the case)
So I would also suggest to make that relation `serialize: false` so it becomes read-only but that would require a small change in `new-subcase.js`, that seems like the only place we set `subcase.agendaActivities`

app/routes/agenda/agendaitems/agendaitem.js
```diff
@@ -22,14 +22,15 @@ export default class AgendaitemAgendaitemsAgendaRoute extends Route {
     }
     this.transition = transition; // set on the route for use in setupController, since the provided "transition" argument there always comes back "undefined"
 
-    this.treatments = await model.treatments;
-    this.newsletterInfo = await this.treatments?.firstObject?.newsletterInfo;
+    const agendaItemTreatment = await model.treatment;
+    this.newsletterInfo = await agendaItemTreatment?.newsletterInfo;
+    this.decisionActivity = await agendaItemTreatment?.decisionActivity;
```
ValenberghsSven, line 27: possible cache issue, should we query for `decisionActivity` instead like we do for the decision route?

app/routes/agenda/agendaitems/agendaitem/index.js
```diff
@@ -29,9 +29,8 @@ export default class DetailAgendaitemAgendaitemsAgendaRoute extends Route {
       const case_ = await this.subcase.case;
       await case_.governmentAreas;
     }
-    const agendaItemTreatment = await model.hasMany('treatments').reload();
-    const anyTreatment = agendaItemTreatment.firstObject;
-    this.newsletterInfo = await anyTreatment?.newsletterInfo;
+    const agendaItemTreatment = await model.belongsTo('treatment').reload(); // TODO: reload required?
```
ValenberghsSven: reload no longer needed

app/services/agenda-service.js
```diff
@@ -221,21 +226,20 @@ export default class AgendaService extends Service {
...
-          // TODO DELETE REPORT !
-          await treatment.destroyRecord();
-        }));
+      if (treatment) {
+        const newsletter = await treatment.get('newsletterInfo');
+        if (newsletter) {
+          await newsletter.destroyRecord();
+        }
+        // TODO DELETE REPORT !
+        // TODO: delete decision activity
```
ValenberghsSven: will do this part

app/services/agenda-service.js
```diff
@@ -246,7 +250,7 @@ export default class AgendaService extends Service {
       await subcase.set('requestedForMeeting', null);
       await subcase.save();
       await subcase.hasMany('agendaActivities').reload();
-      await subcase.hasMany('treatments').reload();
+      await subcase.hasMany('submissionActivities').reload();
```
ValenberghsSven: need to check this

app/templates/print-overviews/decisions/agendaitems.hbs
```diff
@@ -12,37 +12,18 @@
...
-              </Auk::Toolbar::Item>
-            </Auk::Toolbar::Group>
-          </Auk::Toolbar>
+    {{#let (await agendaitemRow.agendaActivity.subcase) (await agendaitemRow.treatment.decisionActivity) as |subcase decisionActivity|}}
+      {{#if (and subcase decisionActivity)}}
+        <div class="auk-u-m-8">
+          <Agenda::Agendaitem::AgendaitemDecision
+            @decisionActivity={{decisionActivity}}
+            @subcase={{subcase}}
+          />
```
ValenberghsSven: changing `AgendaitemDecision` to no longer need subcase args

app/templates/agenda/agendaitems/agendaitem/decisions.hbs
```diff
@@ -6,30 +6,10 @@
...
-      @onDeleteTreatment={{this.refresh}}
-      @allowDelete={{gt this.treatments.length 1}}
-    />
-  </div>
-{{/each}}
\ No newline at end of file
+<div class="auk-u-mb-2">
+  <Agenda::Agendaitem::AgendaitemDecision
+    @agendaitem={{this.agendaitem}}
+    @decisionActivity={{this.model}}
```
ValenberghsSven: changing `AgendaitemDecision` to no longer need agendaitem

app/models/newsletter-info.js
ValenberghsSven: doesn't update right after adding mandatees without a refresh, but pretty sure that was already the case since the computed doesn't listen to updates on `subcase.mandatees`

app/routes/agenda/agendaitems/agendaitem/news-item.js
```diff
@@ -35,10 +32,6 @@ export default class NewsitemAgendaitemAgendaitemsAgendaRoute extends Route {
     super.setupController(...arguments);
     const agendaitem = this.modelFor('agenda.agendaitems.agendaitem');
     controller.set('agendaitem', agendaitem);
-
-    const agendaItemTreatment = await agendaitem.get('agenda-item-treatment');
-    controller.set('agendaItemTreatment', agendaItemTreatment);
-
```
ValenberghsSven: Pretty sure `agendaitem.get('agenda-item-treatment')` was incorrect, but since the controller value was never used we never saw this.
removed it.

ValenberghsSven:
new run with latest changes.
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/254/  :heavy_check_mark:

app/services/agenda-service.js
ValenberghsSven: changed this to `decisionActivities`

app/models/subcase.js
ValenberghsSven: AHA! figured out why I couldn't break this.
The relation is defined as `decisions` in the resources, not `decision-activities`

ValenberghsSven: AHA! figured out why I couldn't break this.
The relation is defined as `decisions` in the resources, not `decision-activities`
Will update the backend resource to `decision-activities` so it matches

PR #1412, opened 2022-07-07
Feature/kas 3516 refactor kort bestek

sergiofenoll:


app/components/newsletter/newsletter-item.hbs
```diff
@@ -1,36 +1,36 @@
...
-{{#if (is-pending this.newsletterInfo)}}
-  <WebComponents::VlLoader @text={{t "newsletter-info-loading-text"}} />
-{{else if (not this.isEditing)}}
-  {{#if (not this.allowEditing)}}
-    {{#if (await this.newsletterInfo.inNewsletter)}}
+<div class="auk-u-mb-10">
+  {{#if (is-pending @newsletterInfo)}}
+    <WebComponents::VlLoader @text={{t "newsletter-info-loading-text"}} />
+  {{else if (not this.isEditing)}}
+    {{#if (not @allowEditing)}}
```
ValenberghsSven: Double negative conditional already existed, perhaps we could change this for clarity?

app/components/newsletter/newsletter-item.hbs
```diff
@@ -1,36 +1,36 @@
...
+      {{#if (await @newsletterInfo.inNewsletter)}}
```
ValenberghsSven: The `newsletter-info` is already awaited in the template of `templates/newsletter/print.hbs`

Then we do a `is-pending this.newsletterInfo` and then `await @newsletterInfo.inNewsletter`.
I think those final 2 are not needed.
Out of scope, I think we have to improve the `templates/newsletter/print.hbs` so we do all the heavy loading in route instead of awaiting in the template.

So ok for this PR, but wanted to mention it because we can do better :D

ValenberghsSven (approved):
can be approved as-is.
Had some suggestions for future work though

PR #1414, opened 2022-07-12
implement small delay right after test login to ensure data is loaded

ValenberghsSven:
active bug where fast switches means entering a route with wrong queryparams is giving test flakyness.
Overall it shouldn't take that much extra time to wait for that data to load


ValenberghsSven (commented):
seemed to help with search spec, did nothing for agendaitem-changes flakyness

PR #1416, opened 2022-07-19
KAS-3521: replace pdf view addon by native "object" pdf-view

MikiDi:
General maintenance + Partial fix for https://kanselarij.atlassian.net/browse/KAS-3521


sergiofenoll:
Very cool that we're going for built-in tools, love it! A little heads up though, last time I tried this (when we first talked about upgrading the PDF viewer), I tested the `<object>` solution locally and I found that it didn't work out of the box on Chrome (and presumably other Chromium-based browsers).

The reason for that is that when Chrome sends a requests for the file that the `<object>` tag points to, the Accept header it sends is too specifc (IIRC it was something like first HTML, then XML, and only then anything else), so our dispatcher routes the request to the frontend, which breaks the page.

It wasn't an issue on Firefox though because it sends `*/*` in its Accept header, which gets nicely routed to the file-service.

MikiDi:
That's good to know. Could you re-test with this PR? Given there is this dedicated HTML key `type`, I'd think the behavior you describe is a bug, or at least an incomplete implementation on Chrome's side? (which doesn't make your case less valid, since most of our users use chrome)

sergiofenoll:
@MikiDi still seems to be the case. 

In Firefox:
imagen
HAR: [PDF download Firefox.json.gz](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/files/9165680/PDF.download.Firefox.json.gz)


In Chrome:
imagen
HAR: [PDF download Chrome.json.gz](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/files/9165681/PDF.download.Chrome.json.gz)


Agreed that it looks like a bug. Given that we pass `type=application/pdf`, you'd expect the Accept header to be `application/pdf`. Haven't digged around to look for an explanation from the browsers though...

sergiofenoll:
Tested this again with the changes to the file-service and it works both on Firefox and Chrome if accessing the frontend *container*. Chrome is still broken when accessing the frontend through `ember serve` (because it does not proxy the request to the backend because Ember assumes all requests that accept HTML are for Ember itself and should not be proxied).

Might be worth it to check whether any of the devs use Chrome for development, as this might then be an annoying issue.

erikap:
Bumped file service in the backend in https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/301

File preview not working through `ember serve` for Chromium based browsers is indeed annoying. Started looking in adding some custom middleware to proxy the file requests based on [this blogpost](https://blog.201-created.com/secrets-of-the-ember-cli-server-bde80bb546dd) but didn't succeed yet. If someone else wants to give it a try, feel free...

PR #1417, opened 2022-07-21
KAS-3505: Draft access level pills

sergiofenoll:
Uses a "stippelijn" for access level pills when the access level hasn't been enforced yet.

Leaving it in draft for a bit because the solution isn't great and I would appreciate some input. Since we need quite a bit of information to determine if an access level pill is in draft state or not, I put it in a helper. All of the needed data is sideloaded there from the passed in piece. Although this works, we lack reactivity here. For the bottom three access levels, which change based on releasing decisions/documents/to Themis, we don't capture those changes at all in the helper and they would require adding refreshes to those actions just so the pill can be rerendered correctly.

Alternatively, the helper could be rewritten to pass in all the needed conditions as arguments, that way (as far as I understand) it would be reactive to changes. But this means drilling down a lot of props through the app, which I don't find very elegant either.

Any input for a good solution here would be appreciated.


sergiofenoll:
Merged in development. After the new changes, the "issue" is specifically for "Intern regering" access levels. If a user goes through the motions of finalizing an agenda (while the documents tab containing the piece with "Intern regering" access level is open) and publishes the decisions, the access level pill will not update until the user manually refreshes.

All in all it's a minor thing and it can be "fixed" by making the Release decisions button refresh the page (though I somewhat dislike that approach).

app/helpers/is-access-level-in-effect.js
```diff
@@ -0,0 +1,82 @@
...
+  async compute([accessLevel, piece]) {
+    if (accessLevel.uri === CONSTANTS.ACCESS_LEVELS.INTERN_SECRETARIE) {
+      // Always in effect
+      return true;
+    }
+
+    let inEffect = false;
+    if (piece) {
+      // Always assume that we're dealing with the latest agenda
+      const agendaitems = (await piece.agendaitems).toArray();
```
ValenberghsSven: Not a big fan of using the inverse relation between models and pieces in general with a many to many.
this inverse is 'read-only' (not serialized) and the `inverse: null` prevents ember from updating agendaitems.pieces in the store. So technically this should be fine, but it's possible some concurrency issues could ensue.

app/helpers/is-access-level-in-effect.js
```diff
@@ -0,0 +1,82 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+import { isPresent } from '@ember/utils';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+
+export default class IsAccessLevelInEffect extends Helper {
+  @service agendaService;
+
```
ValenberghsSven: This helper method is very specifically tuned to pieces of agendaitems.
This is not readily apparent from the helper name.

Document cards seen from subcase views won't show the same styling as document cards viewed from agendaitem views?

app/components/documents/document-card.hbs
```diff
@@ -62,6 +62,7 @@
               <div class="auk-o-flex auk-o-flex--vertical-center">
                 <AccessLevelPill
                   @accessLevel={{this.accessLevel}}
+                  @isDraft={{not (await (is-access-level-in-effect this.accessLevel this.piece))}}
```
ValenberghsSven: Not convinced this is the best way.
For most of the use cases, we start from a route which has all the info about meeting / agenda / agendaitems but then we go into the document card and now have to get all the info again because the document-card is so generic and knows nothing about the meeting / agenda / agendaitem we started from.

ValenberghsSven: Could we callback to the original route/controller to get this needed data?
just pass along the callback down the document-card tree and call it on construction of the component where it's needed?
A part of the helper logic requires just `meeting` and its `activities`, which could be pre-calculated on opening the route and made available to lower routes and components (since that part is the same for an entire meeting and could only be done once?).

ValenberghsSven (commented):
Bit late to the party, wanted to give my opinion.
Very specific await heavy "helper" function in a template doesn't seem like a sustainable way to implement this.
It's difficult to quickly see all possible ways `inEffect` should be true.

Not sure how I would approach this tbh

app/helpers/is-access-level-in-effect.js
sergiofenoll: Hmm, I hadn't though of this. Looking at the FP again, it only specifies rules for pieces related to agendas. From a subcase view, if a piece is linked to an agendaitem, it should still be displayed in the same way as from an agenda view (because the helper fetches all the related data and finds the linked agenda). If the piece is not also linked to an agendaitem, then it will show a draft status (unless the access level is `Intern secretarie`). I think this would be technically correct, given that Yggdrasil only propagates pieces that are linked to an agenda, right?

app/components/documents/document-card.hbs
sergiofenoll: Passing a callback could help here to make the helper less bulky, but that also means drilling down a prop through a lot of layers of components, although it might be the better option in this case. But I will give it a try (once the data migration stuff is finished). Thanks for the tip

PR #1421, opened 2022-07-23
KAS-2749 Add meeting kind prefix to page title on agenda

erikap:
Prefix page title with 'MR' or 'MR VV' depending on the kind of meeting.


ValenberghsSven (approved):
good, unrelated flaky tests aside

PR #1423, opened 2022-07-27
KAS-3530: replace Kort bestek labels with "Op de website"

sergiofenoll:
Label "Zichtbaar in Kort bestek" is "Op de website" geworden.

In tab "Kort bestek" van agendapunt ook, alsook is de pill "Niet zichtbaar in Kort bestek" "Niet op de website" geworden.


app/components/agenda/agenda-overview/agenda-overview-item.hbs
```diff
@@ -141,46 +152,6 @@
...
-                  @skin="muted"
-                  class="auk-u-ml-2"
-                >
-                  <EmberTooltip @side="top" @tooltipClass="tooltip-custom">
-                    <p>
-                      {{t "loading"}}
-                    </p>
-                  </EmberTooltip>
-                </Auk::Icon>
-              {{/if}}
```
ValenberghsSven, line 183: Do we still need this workaround to show a temp icon during loading?
Might be harder to test locally, this is a large agenda problem.
Or with timeouts on the task

ValenberghsSven (approved):

app/components/agenda/agenda-overview/agenda-overview-item.hbs
erikap: I would give it a try without the loading state and await user's feedback. If a loading icon must be shown, we should use another one than `circle-question` since that's also used for an undetermined formally OK status at the same location. This might confuse users.

erikap (approved):

PR #1425, opened 2022-07-29
used the get('uri') to check the current document release status instead of the whole concept object

tdn:
This caused the document list to be empty when it wasn't supposed to be. Simple fix, the whole object was compared to a URI constant instead of the uri property.


erikap (commented):
Oops, sorry for the stupid mistake :flushed:  
Since document publication status promises are awaited, there is no need to use `get()` to retrieve the URI. Just `documentPublicationStatus.uri` should work.

MikiDi:
Resolved by https://github.com/kanselarij-vlaanderen/frontend-kaleidos/commit/de2ddf9297a9b9cb5efe3cc0f21ee1424bfa991f

PR #1430, opened 2022-08-11
feature/KAS-3484-subcase-government-areas

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3484

- added governmentAreas to subcase (kept using areas because of domain/field concepts in 1 list)
- used those areas in subcase view and agendaitem view
- copy governmentAreas on new subcase with previous subcase (not mentioned in ticket, therefore added a TODO)


app/components/cases/new-subcase.js
```diff
@@ -158,6 +158,8 @@ export default class CasesNewSubcase extends Component {
     } else {
       subcase.linkedPieces = pieces;
     }
+    // TODO KAS-3484 is there a situation where we don't have to inherit the governmentAreas?
```
tdn: I don't think there is. Currently `governmentAreas` are changed on the case level, so they are implicitly inherited between subcases already. So this looks fine to me, as long as they can be changed per subcase by the user later on.

app/controllers/cases/case/subcases/subcase/overview.js
```diff
@@ -46,7 +46,7 @@ export default class CasesCaseSubcasesSubcaseOverviewController extends Controll
     const governmentAreas = this.governmentAreas;
```
tdn: It seems weird to me that `governmentAreas` is a tracked property here. This made sense when they came from the parent `case`, but now they are a part of the `subcase` itself. Why keep this property at all when they're in the model already?

tdn (changes requested):
One TODO comment may be removed, and see remark on `app/controllers/cases/case/subcases/subcase/overview.js`. Potentially we can clean up the code a bit here by using the subcase model directly instead of a tracked property.

Apart from that, this looks good to me!

EDIT: actually, upon further inspection, I have an additional remark/question: what happens when an OVRB user modifies the governmentAreas here: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3484-subcase-government-areas/app/controllers/publications/publication/case.js#L10-L16 ?

I'm assuming that if there's no subcases yet ("niet via ministerraad"), this just "pre-sets" the governmentAreas of the case. However, what if there are already subcases? Should this even be editable then? And if so, what happens to the governmentAreas of the individual subcases? ---> answering my own question: publication cases created by Kanselarij profiles are not editable for OVRB, so that's fine: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3484-subcase-government-areas/app/templates/publications/publication/case.hbs#L20

However, governmentAreas on OVRB-created cases remain editable even after subcases were added. This will need to be addressed (see comment further down the review).

app/components/cases/new-subcase.js
```diff
@@ -158,6 +158,8 @@ export default class CasesNewSubcase extends Component {
...
+    subcase.governmentAreas = await latestSubcase.governmentAreas;
```
tdn, line 164: This needs to support an edge-case where an OVRB user creates a case (niet via MR) and adds governmentAreas.

If a kanselarij user then adds a subcase later, I would copy the existing governmentAreas from the case. Since `latestSubcase` will be undefined in this scenario, we'll need an extra step here

tdn (changes requested):

app/components/cases/new-subcase.js
tdn: And actually, if a case is created by OVRB, the governmentAreas remain editable from the publication view even after a subcase has been added. So we need to decide between:

1. only keeping this change on the case level and not propagating (i.e., the current situation)
2. propagating this change on the case level back to all subcases
3. propagating this change on the case level back to the latest subcase 
4. Simply not allowing editing of the governmentAreas on the case level if there is at least 1 subcase here: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3484-subcase-government-areas/app/templates/publications/publication/case.hbs#L20

In my opinion, 1 is dangerous in terms of data consistency, also w.r.t. disclosure via OSLO. 2 and 3 could work if the users want it this way. 4 seems the easiest and safest option to me.

ValenberghsSven: Not following completely.
Kanselarij shouldn't add subcases on a case created by ovrb.
It's possible, but they shouldn't. That's not the proper workflow.
The correct way would be to have a case/subcase created by kanselarij, and connect that to the existing publication via the feature from documents on the agendaitem. (ovrb case will be deleted then)

The flag to determine if ovrb can edit the government areas is the existence of agendaitem.
So if in your example the subcase would be on the agenda, it would no longer be editable.

tdn: > Kanselarij shouldn't add subcases on a case created by ovrb.
> It's possible, but they shouldn't. That's not the proper workflow.

Ok, so this scenario shouldn't happen in practice. However, I feel weird leaving this in a way so the data could end up in an inconsistent state, even though it's not supposed to happen.

I do recall someone from OVRB saying in a meeting they sometimes start preparing the "publication" `case` for certain agenda items beforehand if they know it's going to be a lot of work. So it might not be so uncommon after all...

Also, the fix seems quite simple to avoid the issue altogether: 

1. add the something along the lines of 

```
else if (this.args.case.governmentAreas) {
   subcase.governmentAreas = await this.args.case.governmentAreas;
}
```
here: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/5a61282030b33e5088d0188e843048ac51a4da4b/app/components/cases/new-subcase.js#L130

2. include the existence of a subcase in the editable flag for OVRB.

Since it's minimal effort to include it, is there a reason we shouldn't? (this is a genuine question, maybe I'm overlooking something...)

erikap: @ValenberghsSven is right that the workflow where OVRB users create a case on which Secretarie adds subcases afterwards is not supported. It was indeed mentioned once in a meeting, but the conclusion back then was that those cases are so rare that we won't put effort in supporting it in the app. Therefore, I wouldn't bother about the possible data inconsistency this might introduce at the moment. 

Due to the unclear split between case/besluitvormingsaangelegenheid/publicatieaangelegenheid in the model at the moment, a lot of these kind of data inconsistencies are currently possible. I think it's more important to spend time in the near future to clean up the model and make it OSLO compliant than trying to to try to tie up all loose ends in the current model.

app/controllers/cases/case/subcases/subcase/overview.js
erikap: I assume this is done to handle ember-data relations being async. By using the tracked property and resolving the relation in the `afterModel`-hook of the route, the controller/template doesn't need to bother about the relation being async. In the template of the subcase route `this.governmentAreas` is passed as argument to the government-areas-panel component 
```hbs
      <Utils::GovernmentAreas::GovernmentAreasPanel
        @governmentAreas={{this.governmentAreas}}
        @allowEditing={{this.currentSession.isEditor}}
        @onSave={{this.saveGovernmentAreas}}
      />
```

But it's inconsistent with the way the data passing is done in the agendaitem template. There we do pass the async relation to the government-areas-panel component 
```hbs
  <Utils::GovernmentAreas::GovernmentAreasPanel
    @governmentAreas={{this.subcase.governmentAreas}}
    @allowEditing={{this.currentSession.isEditor}}
    @onSave={{this.saveGovernmentAreas}}
  />
```

We actually try to avoid passing promises as arguments to components, but it seems for this case the government-areas-panel component is aware that `@governmentAreas` might contain a promise since the value is [`yield`-ed in the constructor](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/utils/government-areas/government-areas-panel.js#L42)

TL;DR You're right that the tracked property isn't necessary in the subcase route and the relation can be directly passed to the government-areas-panel component  as done in the agendaitem route.

app/routes/agenda/agendaitems/agendaitem/index.js
```diff
@@ -26,8 +26,7 @@ export default class DetailAgendaitemAgendaitemsAgendaRoute extends Route {
     if (this.subcase) {
       this.submitter = await this.subcase.requestedBy;
       this.meeting = await this.subcase.requestedForMeeting;
-      const case_ = await this.subcase.case;
-      await case_.governmentAreas;
+      await this.subcase.governmentAreas;
```
erikap, line 30: Is this preloading necessary?  The data will automatically be fetched by the template where `this.subcase.governmentAreas` gets called.

app/components/cases/new-subcase.js
ValenberghsSven: In any case, we don't want government areas to "flow" from ovrb to kanselarij. That was made clear it should be a one-way from kanselarij to ovrb.
So even in the scenario above, if ovrb is editing a case and is adding government areas and a subcase is added (not on agenda yet), any save of subcase will trigger the sync-service, where we will also cleanup all government areas on the case that are not on any of its subcases, thus removing any extra government areas that kanselarij has not set themselves.

The ability to add more by ovrb will exist until that subcase is put on an agenda.
From that point, the sync service will remove any extra government areas not found on the subcase from the case.

tdn: Alright, then I concede my objection ;-)

app/routes/agenda/agendaitems/agendaitem/index.js
ValenberghsSven: it seems needed.
Broken without.
image

ValenberghsSven: In the util, when not preloading the proxy is not resolved as expected, resulting in 0 areas
this line `yield this.args.governmentAreas?.toArray();`
Is it trying to make an array out of the proxy?

If I change it to this:
```
const governmentAreasPromise = yield this.args.governmentAreas;
const governmentAreas = yield governmentAreasPromise?.toArray();
```
Then it does working without preloading (still testing if it's reliable in all cases)

ValenberghsSven: Will try to remove the promises though, not very robust.

app/controllers/cases/case/subcases/subcase/overview.js
ValenberghsSven: removed the tracked

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/277/

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/278/
New jenkins run after development merge and service refactor.

Both specs that work with fields (fields-assigning.spec and publication.spec) were ok locally.

PR #1431, opened 2022-08-11
KAS-3556 add parallel saving of many agendaitems

ValenberghsSven:
bugfix https://kanselarij.atlassian.net/browse/KAS-3556

Did not use the `rollbackattributes`.
If any error occurs, the view is most likely going to need to reload anyway.
Also couldn't force any errors from concurrency testing (adding document in tab 2 before or during approving all items) gave no error


app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -222,6 +221,20 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+  /**
+   * @name setAgendaitemFormallyOkThrottled
+   * @description set formally ok on agendaitems, throttled task.
+   * @param agendaitem
+   * @returns promise
+   */
+  @task({ maxConcurrency: 3, enqueue: true })
+  *setAgendaitemFormallyOkThrottled(agendaitem) {
+    if (agendaitem.get('formallyOk') !== CONSTANTS.ACCEPTANCE_STATUSSES.OK) {
+      agendaitem.set('formallyOk', CONSTANTS.ACCEPTANCE_STATUSSES.OK);
```
erikap: No need to use `get()` and `set()`

erikap (approved):
Looks good, besides some old Ember syntax.

app/components/agenda/agenda-header/agenda-actions.js
ValenberghsSven: changed it

ValenberghsSven:
cypress didn't run yet because of dependency issues.
Wait on that for merging.
During testing locally, every call in the old for loop took about 700ms to 1 second, so 70 agendaitems would result in a minute of saving.
With the concurrency of 3, this should already be reduced by 2/3 of that time.
Need to test this on DEV on the large agenda there.

PR #1433, opened 2022-08-12
feature/KAS-3550-move-search-routes

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3550

draft to see if my cypress changes are enough.


app/router.js
```diff
@@ -51,7 +51,11 @@ Router.map(function() {
     this.route('overview', { path: '', });
   });
 
-  this.route('newsletters', { path: '/kort-bestek', });
+  this.route('newsletters', { path: '/kort-bestek', }, function() {
+    this.route('overview', { path: '', });
```
erikap: Can we use the default `index` route here?

app/router.js
```diff
@@ -51,7 +51,11 @@ Router.map(function() {
...
+    this.route('search', { path: '/bericht-zoeken', });
```
erikap: Just `/zoeken` as path?

app/router.js
```diff
@@ -82,6 +86,7 @@ Router.map(function() {
       this.route('proofread', { path: '/nalezen' });
       this.route('late', { path: '/te-laat-in-bs' });
       this.route('reports', { path: '/rapporten' });
+      this.route('search', { path: '/dossier-zoeken' });
```
erikap: Just `/zoeken` as path?

app/routes/newsletters/search.js
```diff
@@ -136,14 +159,10 @@ export default class NewsletterInfosSearchRoute extends Route {
...
-    const controller = this.controllerFor(this.routeName);
-    controller.isLoadingModel = true;
-    transition.promise.finally(() => {
-      controller.isLoadingModel = false;
-    });
-    return true;
+  loading(/*transition, originRoute*/) {
+    // Disable bubbling of loading event to prevent parent loading route to be shown.
+    // Otherwise it causes a 'flickering' effect because the search filters disappear.
+    return false;
```
erikap, line 169: Don't we want to set `isLoadingModel` on the controller here as it's used in the template to show a loading state in the data table?

app/routes/publications/overview/search.js
```diff
@@ -144,14 +171,8 @@ export default class PublicationFlowSearchRoute extends Route {
...
-  loading(transition) {
-    // eslint-disable-next-line ember/no-controller-access-in-routes
-    const controller = this.controllerFor(this.routeName);
-    controller.isLoadingModel = true;
-    transition.promise.finally(() => {
-      controller.isLoadingModel = false;
-    });
-    return true;
+  loading() {
+    return false;
```
erikap, line 185: Don't we want to set `isLoadingModel` on the controller here as it's used in the template to show a loading state in the data table?

app/templates/newsletters/search.hbs
```diff
@@ -0,0 +1,196 @@
...
+            @enter={{fn this.search}}
+            id="dossierId"
+            data-test-route-search-input
+          />
+        </div>
+      </div>
+      <div class="auk-o-grid-col-2">
+        <div class="auk-form-group auk-u-position-relative">
+          <Auk::Label>{{t "from"}}</Auk::Label>
+          <WebComponents::VlDatepicker
```
erikap: Can we start using `Auk::Datepicker` here?

app/templates/newsletters/search.hbs
```diff
@@ -0,0 +1,196 @@
...
+            @enableClear={{this.dateFromBuffer}}
+            @onClear={{fn (mut this.dateFromBuffer) null}}
+            @allowInput={{true}}
+          />
+        </div>
+      </div>
+      <div class="auk-o-grid-col-2">
+        <div class="auk-form-group auk-u-position-relative">
+          <Auk::Label>{{t "until-inclusive"}}</Auk::Label>
+          <WebComponents::VlDatepicker
```
erikap: Can we start using `Auk::Datepicker` here?

app/templates/publications/overview/search.hbs
```diff
@@ -0,0 +1,243 @@
...
+                {{dateType.label}}
+              </option>
+            {{/each}}
+          </select>
+        </div>
+      </div>
+      <div class="auk-o-grid-col-2">
+        <div class="auk-form-group auk-u-position-relative">
+          <Auk::Label>{{t "from"}}</Auk::Label>
+          <WebComponents::VlDatepicker
```
erikap: Can we start using `Auk::Datepicker` here?

app/templates/publications/overview/search.hbs
```diff
@@ -0,0 +1,243 @@
...
+            @enableClear={{this.dateFromBuffer}}
+            @onClear={{fn (mut this.dateFromBuffer) null}}
+            @allowInput={{true}}
+          />
+        </div>
+      </div>
+      <div class="auk-o-grid-col-2">
+        <div class="auk-form-group auk-u-position-relative">
+          <Auk::Label>{{t "until-inclusive"}}</Auk::Label>
+          <WebComponents::VlDatepicker
```
erikap: Can we start using `Auk::Datepicker` here?

app/controllers/newsletters/search.js
```diff
@@ -0,0 +1,101 @@
...
+    this.sort = '-agendaitems.meetingDate';
+  }
+
+  deserializeDate(date) {
+    return date && moment(date, 'DD-MM-YYYY').toDate();
+  }
+
+  serializeDate(date) {
+    return date && moment(date).format('DD-MM-YYYY');
+  }
```
erikap: Move to a util like `serialize-date` that can be reused in all search routes?

Start using `format` of https://date-fns.org/v2.29.1/docs/format#usage instead of `moment`

app/controllers/newsletters/search.js
```diff
@@ -0,0 +1,101 @@
...
+  constructor() {
+    super(...arguments);
+    this.page = 0;
+    this.size = this.sizeOptions[2];
+    this.sort = '-agendaitems.meetingDate';
+  }
+
+  deserializeDate(date) {
+    return date && moment(date, 'DD-MM-YYYY').toDate();
+  }
```
erikap: Move to a util like `deserialize-date` that can be reused in all search routes?

Start using `parse` of https://date-fns.org/v2.29.1/docs/parse#usage instead of `moment`

app/routes/newsletters/search.js
```diff
@@ -126,8 +143,14 @@ export default class NewsletterInfosSearchRoute extends Route {
     });
   }
 
-  setupController(controller) {
-    super.setupController(...arguments);
+  setupController(controller, model) {
+    super.setupController(controller, model);
+    const params = this.paramsFor('search');
```
erikap: Should be reworked since all params are defined on this route now and no longer on a parent route.

app/routes/newsletters/search.js
```diff
@@ -126,8 +143,14 @@ export default class NewsletterInfosSearchRoute extends Route {
...
+    controller.searchTextBuffer = params.searchText;
+    controller.mandateesBuffer = params.mandatees;
+    controller.page = params.page;
+    controller.dateFromBuffer = controller.deserializeDate(params.dateFrom);
+    controller.dateToBuffer = controller.deserializeDate(params.dateTo);
     controller.emptySearch = isEmpty(this.paramsFor('search').searchText);
```
erikap: Define as getter on controller?

erikap (commented):
I saw too late that this one was still in draft. Have already put my preliminary comments. Some comments are only added in newsletter related files but also apply to publications.

In general I think we should spend some time cleaning up the search code for the 2 new routes as they will not be addressed anymore later when refactoring the general search.

app/templates/newsletters/search.hbs
ValenberghsSven: replacing `vlDatepicker` is out of scope. we can, but I shouldn't do that here.

app/templates/newsletters/search.hbs
ValenberghsSven: same

app/router.js
ValenberghsSven: Those routes are in the ticket on usertest, assumed that was requested
image

app/controllers/newsletters/search.js
ValenberghsSven: I would again argue that this is out of scope.
Nice to have, wont do in this ticket.

app/routes/newsletters/search.js
ValenberghsSven: true, I'll start with removing the `const searchParams = this.paramsFor('search');` since that wont do anything I imagine.

app/routes/newsletters/search.js
ValenberghsSven: will check, guessing the entire `setupController` needs to be relooked since `const params = this.paramsFor('search');` needs to go

app/routes/newsletters/search.js
ValenberghsSven: was unsure about this, need to check

app/routes/publications/overview/search.js
ValenberghsSven: same

app/router.js
ValenberghsSven: Was unsure what to pick since overview made more sense to me.
Do you mean a rename?
```
  this.route('newsletters', { path: '/kort-bestek', }, function() {
    this.route('index', { path: '', });
    this.route('search', { path: '/bericht-zoeken', });
  });
```

app/routes/newsletters/search.js
ValenberghsSven: done

app/routes/publications/overview/search.js
ValenberghsSven: we do need to return false in this case because of the parent reload.
I have added the setting of the `isLoadingModel` so the content of the datable switches between loading and showing results.

app/router.js
erikap: You get the index route by default in Ember. No need to define it explicitely in `router.js`. See https://guides.emberjs.com/v3.28.0/routing/defining-your-routes/#toc_index-routes

app/router.js
erikap: I wasn't aware that it came from there. Nonetheless, I think that naming was made ad hoc in usertest and we shouldn't necessarily take it literally. I prefer some consistency across the (sub)routes.

PR #1443, opened 2022-08-17
KAS-3561: Set access level to Intern Regering for old piece when uploading a new version

sergiofenoll:


erikap (approved):
Very clean  :+1:

PR #1446, opened 2022-08-19
KAS-3517 rename queryParams of agendas route to be unique

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3517

TLDR, switching to a sibling route while async model is loading is causing the queryparams from route A to be used for route B.
Since we use those params for out model queries we can get errors. Mainly because of the `sort` param.

Tried a lot of things including refactoring the newsletters route to stop using mixins, thought that might help (see second PR)
But in the end, the issue was solved by making the params on the `agendas` route unique.
Could be a temporary fix until the ember community comes up with a root cause fix.


erikap:
Failing test not related to this PR. Going to merge.

PR #1452, opened 2022-08-29
KAS-3579: Minimal `include`s

sergiofenoll:
Not the most efficient function, but it should suffice. To apply the minimal includes to all outgoing queries, the `application` adapter has two overriden methods, `buildQuery` and `buildURL`. It's not entirely clear to me based on the docs which method is applied to which store calls, but while testing it seems that `buildURL` is only used for certain methods, e.g. `findRecord`, while `buildQuery` is only used for others, e.g. `query`. So here both methods are overriden just to be sure.

### DIFF
--- a/app/utils/minimal-include.js
+++ b/app/utils/minimal-include.js
@@ -1,0 +1,27 @@
+/**
+ * Returns the a comma-separated string containing the minimal necessary include
+ * paths. When a path is a complete prefix of another path, the shorter path
+ * doesn't need to be part of the include parameter, so it gets removed by this
+ * function.
+ *
+ * @param {Array<string>} paths list of mu-cl-resources paths
+ * @returns {string} String comma-separated string with the minimal paths
+ */
+export default function minimalInclude(paths) {
+  return paths
+    .reduce((acc, path, i, arr) => {
+      let isPrefix = false;
+      arr.slice(i + 1).forEach((otherPath, j) => {
+        if (path.includes(`${otherPath}.`)) {
+          arr.splice(i + j + 1, 1);
+          i--;
+        }
+        if (!isPrefix) {
+          isPrefix = otherPath.includes(`${path}.`)
+        }
+      });
+      if (isPrefix) return acc;
+      return [...acc, path];
+    }, [])
+    .join(',');
+}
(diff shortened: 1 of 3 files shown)


app/utils/minimal-include.js
```diff
@@ -0,0 +1,38 @@
...
+          // We deleted an element further down the array, decrease the index so
+          // it still points to the right element
+          i--;
+        }
+        if (!isPrefix) {
+          isPrefix = otherPath.startsWith(`${path}.`)
+        }
+      });
+      if (isPrefix) return acc;
+      return [...acc, path];
```
erikap: `return isPrefix ? acc : [...acc, path];` to improve readability?

app/utils/minimal-include.js
```diff
@@ -0,0 +1,38 @@
...
+ * Returns the a comma-separated string containing the minimal necessary include
+ * paths. When a path is a complete prefix of another path, the shorter path
+ * doesn't need to be part of the include parameter, so it gets removed by this
+ * function.
+ *
+ * The function works by iterating over each path and storing only paths that
+ * aren't a prefix of other paths. For each path, we iterate over all other
+ * paths and make sure that (1) the current path is not a prefix of another path
+ * and (2) that other paths are not full prefix of the current path. If (1), we
+ * discard the current path, if (2) we discord the other path.
```
erikap: Can we maybe simplify by first sorting the `paths` array on string length? As such we only need to check (1) as (2) cannot occur then.

app/utils/minimal-include.js
```diff
@@ -0,0 +1,38 @@
...
+          i--;
+        }
+        if (!isPrefix) {
+          isPrefix = otherPath.startsWith(`${path}.`)
+        }
+      });
+      if (isPrefix) return acc;
+      return [...acc, path];
+    }, [])
+    .join(',');
```
erikap, line 31: Can we do an alphabetical sort of the resulting array in the end to 'stabilize' the outcome of the function?

sergiofenoll: I initially didn't want to depend on sorting, just in case the cache-warmup had non-sorted paths so that I didn't have to change them. Since it doesn't, it shouldn't matter whether we sort or not. Perhaps it's even better to just sort in the beginning on the paths (e.g. `agendaitem, agendaitem-type.kind, agendaitem.status` → `agendaitem, agendaitem.status, agendaitem-type.kind`), then the next step is just a single pass over the sorted list to check if a path is a prefix of the next path?

erikap (approved):

PR #1457, opened 2022-09-01
KAS-3570 case flow split

sergiofenoll:
https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/309
https://github.com/kanselarij-vlaanderen/yggdrasil/pull/42
https://github.com/kanselarij-vlaanderen/case-documents-sync-service/pull/3
https://github.com/kanselarij-vlaanderen/gov-field-sync-service/pull/3

Splits case and decisionmaking-flow. Where necessary (e.g. the case routes) the decisionmaking-flow model is now used in place of case. Functionality should be the same, except:
- `cases` route will only list decisionmaking-flows. This means that cases which are not linked to a MR (e.g., a case created when a publication is created manually in the `publications` route) are not shown here
- Publications will only display the linked government areas (beleidsvelden en -domeinen) if the publication is via MR (linked to a decisionmaking-flow). A manually created publication (from the `publications` route) will not have governmentAreas and cannot have governmentAreas added (because governmentAreas are linked to a decisionmaking-flow and no longer to a case). This also means that the governmentAreas panel is no longer visible for publications not via MR.

Jenkins:
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/280/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/281/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/282/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/283/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/284/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/285/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/286/~~
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/287/


app/routes/cases/case/subcases/subcase/overview.js
```diff
@@ -16,7 +16,7 @@ export default class CasesCaseSubcasesSubcaseOverviewRoute extends Route {
   };
 
   async beforeModel() {
```
erikap: `async` is not needed here

app/services/publication-service.js
```diff
@@ -56,23 +56,23 @@ export default class PublicationService extends Service {
   ) {
     const now = new Date();
 
-    let case_;
+    let _case;
```
erikap, line 59: Not a comment on this line, but on `getIsViaCouncilOfMinisters()` in the `publication-service`.

The logic to check whether a publication flow is via council of ministers becomes simpler. We just need to check whether the case related to the publication-flow has a related decisionmaking-flow. Might be that this helper function `getIsViaCouncilOfMinisters()` becomes void.

app/routes/publications/publication.js
```diff
@@ -9,8 +9,7 @@ export default class PublicationRoute extends Route {
     return this.store.queryOne('publication-flow', {
       'filter[:id:]': params.publication_id,
       include: [
-        'case',
-        'case.subcases',
+        'case.decisionmaking-flow.subcases',
```
erikap: Is this include still required here?  It used to be here for the check executed in `getIsViaCouncilOfMinisters()` in the publication-service, but it seems to me the subcases of the decisionmaking-flow are no longer needed in this route.

app/routes/publications/publication.js
```diff
@@ -9,8 +9,7 @@ export default class PublicationRoute extends Route {
...
         'decision-activity',
```
erikap, line 13: If the check in `getIsViaCouncilOfMinisters()` in the publication-service changes, this include might also be overkill.

app/templates/cases/case/subcases/overview.hbs
```diff
@@ -2,12 +2,13 @@
 
 <Subcases::SubcasesOverviewHeader
   @case={{this.case}}
+  @decisionmakingFlow={{this.decisionmakingFlow}}
```
erikap, line 4: Are both arguments `@case` and `@decisionmakingFlow` required? Feels a bit odd. Can't one of both be skipped and loaded deeper down in the component structure?

erikap:
> * Publications will only display the linked government areas (beleidsvelden en -domeinen) if the publication is via MR (linked to a decisionmaking-flow). A manually created publication (from the `publications` route) will not have governmentAreas and cannot have governmentAreas added (because governmentAreas are linked to a decisionmaking-flow and no longer to a case). This also means that the governmentAreas panel is no longer visible for publications not via MR.

@tdn @sergiofenoll Are we sure about this? Doesn't OVRB want to assign government-areas to publication-flows not via MR? 
I think we need to add government-areas on `publication-flow` as well. In case of a publication via MR this will contain a read-only copy of the government-areas of the `decisionmaking-flow`. In case of a publication not via MR the list can be freely edited by OVRB.

app/controllers/search/cases.js
```diff
@@ -58,7 +58,7 @@ export default class CasesSearchController extends Controller {
   }
 
   @action
-  navigateToCase(_case) {
-    this.router.transitionTo('cases.case.subcases', _case.id);
+  async navigateToCase(dmf) {
```
erikap: `async` not needed

erikap (changes requested):
In general, can we try to avoid abbreviated variable names like `dmf`. That's kind of cryptic if you're new to the codebase.

tdn:
> Publications will only display the linked government areas (beleidsvelden en -domeinen) if the publication is via MR (linked to a decisionmaking-flow). A manually created publication (from the publications route) will not have governmentAreas and cannot have governmentAreas added (because governmentAreas are linked to a decisionmaking-flow and no longer to a case). This also means that the governmentAreas panel is no longer visible for publications not via MR.
> @tdn @sergiofenoll Are we sure about this? Doesn't OVRB want to assign government-areas to publication-flows not via MR? I think we need to add government-areas on publication-flow as well. In case of a publication via MR this will contain a read-only copy of the government-areas of the decisionmaking-flow. In case of a publication not via MR the list can be freely edited by OVRB.

Just checked this with Johan, and it seems I was indeed wrong in the ticket. OVRB will add `governmentAreas` to their own cases (niet via MR). Misunderstood this when setting up the story, sorry about that.

So yes, we'll need to have a `has-many` for `publication-flow` as well, and it will need to be synced with the `governmentAreas` of the `decision-flow` if it exists. (And then maybe the governmentAreas need to be read-only?)

tdn:
@erikap @sergiofenoll TBD at 14:30: Apparently only the `governmentAreas` related to the most recent `subcase` of the `decision-flow` are relevant to a `publication-flow`, even if they're both in the same `case` and that `decision-flow` has other `governmentAreas` coming from earlier `subcases`. 

So that would mean that `governmentAreas` related to `publication-flow` are not necessarily equal to those of `decision-flow`. Is that something we can live with?

Workflow would then be (as I see it):
- Publication `case` without `decision-flow`: OVRB adds `governmentAreas` to `publication-flow`. If a `decision-flow` is added to the `case` later, the first `subcase` copies these  `governmentAreas`. (note that this is a scenario that shouldn't happen in practice)
- Publication `case` with `decision-flow`: automatically add `governmentAreas` of the latest `subcase` to `publication-flow`. This can remain editable by OVRB in case they change later (no need to automate this)

app/routes/publications/publication.js
sergiofenoll: :thought_balloon: might be missing something but it seems they were never used? `getIsViaCouncilOfMinisters` doesn't use the subcases either, only the `publicationFlow`'s `decisionActivity`. And grepping for `\.subcases` in the codebase doesn't turn up results for the publication routes/components  :thinking:

app/routes/publications/publication.js
sergiofenoll: Seems like `decisionAcitvity` is still used in the decision-related route and components. E.g. `DecisionsInfoPanel` uses it a bunch. Might be worth it to keep it in for that.

app/routes/publications/publication.js
erikap: `getIsViaCouncilOfMinisters()` previously contained logic based on subcases. So that include could actually already have been removed before.

tdn:
@sergiofenoll cleared up with @erika : go ahead and do this as mentioned in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1457#issuecomment-1240363984

We just need this:
- Publication `case` with `decision-flow`: automatically add `governmentAreas` of the latest `subcase` to `publication-flow`. This can remain editable by OVRB in case they change later (no need to automate this)

The scenario of a publication `case` without `decision-flow` getting one later will not be supported, so no need to do anything different there

erikap (approved):

erikap:
Executing a final Jenkins run before merging: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/289

sergiofenoll:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/37/

erikap:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/299/

ValenberghsSven:
replaced the case id's in tests with the new decisionmaking-flow id's.
testdata.zip in app-kaleidos contains db with migration already done
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/300

Edit: missed 1 test with 3 counts of searching on `/cases` in `subcase.spec.js`
All the other test are green
Running again with fixed test
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/301

ValenberghsSven:
During test fixes I noticed a bug:
There is a moment where adding governmentAreas on agenda detail view is possible with an empty list, even though the subcase already has some governmentAreas. Cypress is hitting that window and the areas are overridden with only the new selection.
It's possible to manually reproduce with throttled network.
As a fix, I was thinking of hiding the edit button while the existing list is still loading to counter this.

ValenberghsSven:
> During test fixes I noticed a bug: There is a moment where adding governmentAreas on agenda detail view is possible with an empty list, even though the subcase already has some governmentAreas. Cypress is hitting that window and the areas are overridden with only the new selection. It's possible to manually reproduce with throttled network. As a fix, I was thinking of hiding the edit button while the existing list is still loading to counter this.

To be specific:
When the `<governmentAreasPanel>` is rendered the relation is coming in (as a proxy) but not empty.
THe problem is that we do some processing to create some custom `Row` objects.
When we press the "edit" button, we grab information from those objects.
When pressing to fast, we have no information and render an empty fields/domains list in the edit mode.

Fixed it by hiding the edit button when the loading task is still running

ValenberghsSven:
run after fix
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/302/

ValenberghsSven:
first run after a dev merge with many selector changes
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/303/

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/304/
1 failing test is already fixed on dev, so this run can be considered as successful

PR #1458, opened 2022-09-02
KAS-3551 search relevance tuning

erikap:
Boost matches founds in specific search fields

For search on agendaitems: matches in title/short-title are more important than matches in documents and press-release title/text
For search on cases:  matches in case title/short-title are most important, next matches in subcase title/short-title.

User gets the option to sort on meeting date (as it is now) or on relevance. As such, users can evaluate whether sorting by relevance better fits their needs.

Backend: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/310


erikap:
Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/33/

PR #1461, opened 2022-09-02
KAS-3598: Editor grows too large

sergiofenoll:
When a line is too long and a user is typing in the middle of the
sentence, the editor doesn't always wrap long words at the end of the
line. To remedy this we apply overflow-wrap: anywhere to the editor,
which makes it break up words over multiple lines if they're too long.

### DIFF
--- a/app/styles/addon-overrides/_ember-rdfa-editor.scss
+++ b/app/styles/addon-overrides/_ember-rdfa-editor.scss
@@ -105,3 +105,7 @@ $say-editor-padding: 1rem;
     padding-top: 3rem;
   }
 }
+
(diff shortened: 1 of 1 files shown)


ValenberghsSven:
not tested yet.
`anywhere` is not supported by safari but this shouldn't be a problem?
The app accepts safari, the use of the rdfa-editor only accepts firefox and chrome

image

ValenberghsSven:
are we cherry-picking this to ACC branch when ok?

ValenberghsSven (approved):
Managed to reproduce the error, this PR seems to fix it.

PR #1475, opened 2022-09-15
KAS-3624 fix bug where invalid date is being used

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3624

Bugfix: The date for "plannedStart" for themis-publication and internal-document activities are not filled in for annex meetings.
The bug was caused by using an non-existing property on a model instead of a locally declared date.

Also made the date input field available for editing.


ValenberghsSven:
Hotfixed on PROD in separate commit and pulled back into dev.

PR #1476, opened 2022-09-19
Feature/kas 3490 rename confidential

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3490

backend PR https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/312

- subcase "confidential" is now "limited-access". only changed the translation string used in the relevant views.
  the `subcase.confidential` attribute name and relation `ext:vetrouwelijk` are not changed because we decided not to change this at the moment. Added a comment in the model to inform/remind others.
- access-level "ministerraad" changed to "vertrouwelijk" both in data and in frontend naming.
- copied styling of access-level pill from usertest: 
      - removed "small" size
      - changed color of "intern-regering"
      - changed icons of all levels except "publiek"


tdn (approved):
Couldn't test locally since I have a production db copy that I need for other work, which doesn't have the case split migration yet. But see no reason this shouldn't work.

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/306/
Flaky test on propagation failed, rest passed

PR #1478, opened 2022-09-19
KAS-3584 change file to replace with word file replacement

ValenberghsSven:
BASE: ACCEPTANCE :warning: 

Fix for flag in https://kanselarij.atlassian.net/browse/KAS-3584

- Fixed a typo in document-card.
- Changed file to show in viewer is word file when available, fallback to pdf or piece (profiles who can see piece but not file)
- Renamed some variables to indicate what file we are replacing  (future work may include also replacing the actual pdf file here)
- Added a translation. Did not find a design where there is no sourcefile, so was unsure how to name the button in that case so went for generic "Opladen"

### DIFF
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -48,7 +48,7 @@
                     href={{this.piece.file.primarySource.namedDownloadLink}}
                     download
                   >
-                    {{t (if (eq this.piece.file.primarySouce.extension "docx") "word-document" "source-file-lowercase")}}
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
no word doc yet:
image
image

ValenberghsSven:
word doc uploaded:
image
image

ValenberghsSven:
left some code in comment. The function seems like a copy but was never used in the document-viewer component.

PR #1483, opened 2022-09-23
Feature/kas 1652 retracted refactor

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-1652

App : https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/315

## What has changed:
Sidebar-item:
- added task to load the decisionActivity data in a tracked property
- use the property to determine postponed or retracted in a **getter**
- had to use `.get()` because of proxies

Agenda-overview-item:
- added task to load the decisionActivity data in a tracked property
- use the property to determine postponed or retracted in the **template**

Agendaitem-case-panel-view:
- added task to load the decisionActivity data in a tracked property
- use the property to determine postponed or retracted in the **template**
- added postponed/retracted to template separately because text is different

Agendaitem-controls:
- added task to load the decisionActivity data in a tracked property
- use the property to determine postponed or retracted in the **template**
- add retracted to template separately because text is different
- set the `decision-result-code` to postponed/retracted and reset to default when reverted
- made template logic to only show "x ongedaan maken" when status is x as stated in the jira ticket
(technically you could postpone a retracted item in 1 step, now you have to do it in 2 by reverting the retraction and then postponing) 
- Reused default logic from the `agenda-service` to be consistent.

item-group/item/content:
- added awaits to this template only component to show the resultcode status pull (maybe better to have loading tasks running in constructor?) 

Subcase-description-view:
- used new service methods to check if subcase is postponed or retracted by setting local properties.
- use the propertes to determine postponed or retracted in the **template**

Agendaitem-table:
- changed getter to get the full path from agendaitem because of sync method.
- added also filtering on retracted
- renamed to match the combined function of filtering on both statusses

Agendaitem:
- removed attribute from model

Decision-result-code:
- added `isRetracted` getter

Decisions/agendaitems:
- added relation to result code to `includes`

Agenda-service:
- removed default setting the `retracted` attribute

Subcase-is-approved:
- added methods to get the retracted and postponed status of a subcase
(should hold up with multiple decisions, if the last one is approved the status should be "approved" even if it was postponed. When 1 subcase is both postponed and retracted but not approved we should show both. Users may disagree though.)

Tests:
- Updated 1 test that was postponing an agendaitem
- updated selectors

nl-be.json:
- changed some translations
- added some translations.


app/services/subcase-is-approved.js
```diff
@@ -28,4 +28,40 @@ export default class SubcaseIsApprovedService extends Service {
...
+    if (meeting?.isFinal) {
+      const retractedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.INGETROKKEN
+      );
+
+      const nrDecisionActivities = await this.store.count('decision-activity', {
+        'filter[subcase][:id:]': subcase.id,
+        'filter[decision-result-code][:id:]': retractedDecisionResultCode.id,
+      });
```
tdn: Do we really need this extra call to get the `retractedDecisionResultCode` concept?

Wouldn't `'filter[decision-result-code][:uri:]': CONSTANTS.DECISION_RESULT_CODE_URIS.INGETROKKEN,` also work?
(see https://github.com/mu-semtech/mu-cl-resources#special-filters )

app/services/subcase-is-approved.js
```diff
@@ -28,4 +28,40 @@ export default class SubcaseIsApprovedService extends Service {
...
+
+    if (meeting?.isFinal) {
+      const postponedDecisionResultCode = await this.store.findRecordByUri(
+        'decision-result-code',
+        CONSTANTS.DECISION_RESULT_CODE_URIS.UITGESTELD
+      );
+
+      const nrDecisionActivities = await this.store.count('decision-activity', {
+        'filter[subcase][:id:]': subcase.id,
+        'filter[decision-result-code][:id:]': postponedDecisionResultCode.id,
```
tdn: Same thing here, I don't think we need 2 calls to the backend here. 
`'filter[decision-result-code][:uri:]': CONSTANTS.DECISION_RESULT_CODE_URIS.UITGESTELD` should work as well.

translations/nl-be.json
```diff
@@ -675,8 +675,8 @@
   "option-others": "Andere",
   "saving-agendaitem-title-message": "Aanpassen van agendapunten",
   "saving-change-message": "Uw aanpassing wordt opgeslagen.",
-  "saving-retracted-agendaitem-title": "Aanpassen van uitstelling",
-  "saving-retracted-agendaitem-message": "Uw aanpassing wordt opgeslagen.",
+  "saving-decision-result-title": "Aanpassen van resultaat beslissing",
+  "saving-decision-result-message": "Uw aanpassing wordt opgeslagen. Het resultaat van de beslissing wordt aangepast",
```
tdn, line 682: Possibly the people from Secretarie will have other suggestions for these labels, but we'll cross that bridge when we get to it.

tdn (changes requested):

app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -86,19 +96,18 @@ export default class AgendaitemControls extends Component {
   }
 
   @action
-  async postponeAgendaitem(agendaitem) {
-    this.isSavingRetracted = true;
-    agendaitem.set('retracted', true);
-    await agendaitem.save();
-    this.isSavingRetracted = false;
+  async postponeAgendaitem() {
+    this.isSavingDecisionResult = true;
```
erikap: Convert to a ember-concurrency task and use `isRunning` instead of manually managing this flag? Same comment for `retractAgendaitem` and `resetDecisionResultCode`.

app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -110,4 +119,26 @@ export default class AgendaitemControls extends Component {
...
+    const isAnnouncement =
+      agendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT;
+    const defaultDecisionResultCodeUri = isAnnouncement
+      ? CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME
+      : CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD;
+    this.setDecisionResultCode(defaultDecisionResultCodeUri);
+    this.isSavingDecisionResult = false;
+  }
+
+  async setDecisionResultCode(DRC_URI) {
```
erikap: We usually don't write function params in capital letters

app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -110,4 +119,26 @@ export default class AgendaitemControls extends Component {
...
+    // DRC_URI should come from CONSTANTS.DECISION_RESULT_CODE_URIS
```
erikap: Comment doesn't have a lot of value imo. It has become a commonly used pattern in the codebase so I don't think this needs to be explicated here.

app/components/agenda/agendaitem/agendaitem-controls.hbs
```diff
@@ -1,23 +1,32 @@
...
-            <Menu.Item {{on "click" (fn this.advanceAgendaitem @agendaitem)}} data-test-agendaitem-controls-action-advance>
-              {{t "not-postpone"}}
-            </Menu.Item>
-          {{else}}
-            <Menu.Item {{on "click" (fn this.postponeAgendaitem @agendaitem)}} data-test-agendaitem-controls-action-postpone>
-              {{t "postpone"}}
-            </Menu.Item>
-          {{/if}}
+        {{#if this.areDecisionActionsEnabled}}
+          {{#let this.decisionActivity.decisionResultCode as |drc| }}
```
erikap: Not a fan of abbreviated variable names as they make the code rather cryptic. Maybe use `resultCode` or just `code`?

app/components/utils/agendaitem-table.js
```diff
@@ -100,25 +100,28 @@ export default class AgendaitemTable extends Component {
...
-  setRowsPostponed(tableRows, model) {
+  setRowsNotApproved(tableRows, model) {
     const rowsCurrentlyInTable = tableRows;
-    const postponedItems = model.filter((agendaitem) => agendaitem.get('retracted'));
-    postponedItems.forEach((postponedItem) => {
-      const postponedRowInTable = rowsCurrentlyInTable.find(
-        (rowFromTable) => rowFromTable.content.get('id') === postponedItem.get('id')
+    // relations are included in route
+    const postponedItems = model.filter((agendaitem) => agendaitem.get('treatment.decisionActivity.decisionResultCode.isPostponed'));
+    const retractedItems = model.filter((agendaitem) => agendaitem.get('treatment.decisionActivity.decisionResultCode.isRetracted'));
```
erikap: Filtering an array on an async property will not work if the data isn't fetched already. But since this component is only used in the print-overviews, which are currently not actively used in the app and require a major rewrite, I would leave it as it is in this story.

app/services/subcase-is-approved.js
```diff
@@ -28,4 +28,40 @@ export default class SubcaseIsApprovedService extends Service {
...
+      });
+      return nrDecisionActivities > 0;
+    }
+    return false;
```
erikap: Is it correct to consider a subcase only being postponed if the meeting is finalized? (same question for retraction)

erikap (changes requested):

app/services/subcase-is-approved.js
ValenberghsSven: Ah yes, we can.
I copied this from a similar one that had 2 id's (since 2 URI's dont work)
But with only 1, uri is fine.

app/components/utils/agendaitem-table.js
ValenberghsSven: It works, the data is fetched with `includes` from  frontend-kaleidos/app/routes/print-overviews/decisions/agendaitems.js

app/services/subcase-is-approved.js
ValenberghsSven: Not sure, I would say it is correct but It also depends on the view.
A postponed subcase could be reverted right before finalizing meeting?
We only show approved when the meeting is finalized.

We might come back to this with feedback from the user.

app/components/agenda/agendaitem/agendaitem-controls.hbs
ValenberghsSven: Can change, no problem.
As to the approach of using `#let`, is that ok?
I could have just used the long path twice and then there would be no need to have an alias.

app/components/agenda/agendaitem/agendaitem-controls.js
ValenberghsSven: passing aliased constants as params in a method might not be so common, was unsure of capitals were better.
Will change and remove comment

app/components/agenda/agendaitem/agendaitem-controls.hbs
ValenberghsSven: `resultCode` sounds less vague down the line. using that

app/components/utils/agendaitem-table.js
erikap: Yes, I know. But even then we have the convention of still explicitely awaiting the relationship (even though we know that the promise will immediately resolve) because someone might remove the include in the future and not be aware that this will break things here.

But as said, since this component requires a major rewrite, you can just leave it as it is for now.

app/components/agenda/agendaitem/agendaitem-controls.hbs
erikap: No strong opinion about the `{{#let}}`. I get that it might sometimes be useful to make things more readable in templates.

erikap (approved):

translations/nl-be.json
ValenberghsSven: These translations in the popup won't be visible for long.
Could revert these back to be just vague saving messaging.

ValenberghsSven:
## :warning: Don't merge this branch yet! This branch is the base for the entire feature.
Unless all relevant tickets are done the full app might not work as expected.

tdn (approved):
Tested locally and all seems to work well

erikap:
Since this PR is approved, going to start merging branches that started from this feature branch into this PR so that we end up with 1 big PR that can be merged.

erikap:
Related PRs to be merged before merging this one in development
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1491
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1496
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1498
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1502
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1517
- [x] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1519

ValenberghsSven:
> Since this PR is approved, going to start merging branches that started from this feature branch into this PR so that we end up with 1 big PR that can be merged.

That was the idea, merge whats ready until we got the full feature in 1 branch but contained in separate PR's for easier review

ValenberghsSven:
http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/52/
final builds ongoing, attempting to merge today if nothing major pops up

so far: 
- 1 subcase test (code mistake, fixed)
- flaky agenda.spec with date issue today (wrong month)

PR #1484, opened 2022-09-26
KAS-3606 Use new authentication model

erikap:
Backend PR: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/314

Update login flow to use new authentication model. User role is fetched from backend. Frontend contains a mapping of roles to user-groups and accompanying permissions. E.g. `kabinetmedewerker` and `kabinetdossierbeheerder` roles both map to user-group `kabinet` in the current app. Permissions might become more finegrained in the future.

Renamed `roles` in frontend to `permissions` to avoid confusion with the user roles that are defined in the backend and ACM/IDM.

Update of the user management page to a basic list listing all users including the groups/organizations they belong to. Roles are no longer managed in the Kaleidos app.

Did an attempt to update the Cypress tests to make use of the new mock-accounts.

PR also contains some fixes related to authentication/login:
- correctly clear current-session on logout
- don't display tabs in main header if user is not authenticated
- removal of duplicate toolbar on mock-login page
- consistent permission checking in `beforeModel` hook of main routes
- cleanup of unused components
- moved preferred default route per user-group to config file

**Note: this PR may only be merged/deployed after ACM/IDM has upgraded the user roles on their side. Otherwise, all users will login as being 'Overheidsorganisatie' when using ACM/IDM.**


erikap:
Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/50

erikap:
> Jenkins: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/50

All tests succeeded. 

Waiting to merge this PR until ACM/IDM upgraded the user roles on their side.

PR #1486, opened 2022-09-26
KAS-3459 add alert for retracted and postponed

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3459

- Added the loading of relations needed to `afterModel`
- Added alert for retracted and postponed
- Already made headway for the `Auk:DropdowMenu` for next ticket, there is currently no option to pass a block in the `Auk::Alert` so made a semi copy. If we decide to change the `Auk::Alert` template we could but seemed out of scope right now.
- Reused some translations string but changed the vague "punt" to "agendapunt"

Questions.
- We show some pills in the same route in the titels panel.
Does this replace those pills? the screenshot in ticket does not have that pill.
- With the relations loaded in the route, we could pass the `DecisionActivity` to `Agenda::Agendaitem::AgendaitemControls` so we don't need a loading task there.

### DIFF
--- a/app/routes/agenda/agendaitems/agendaitem/index.js
+++ b/app/routes/agenda/agendaitems/agendaitem/index.js
@@ -31,6 +31,8 @@ export default class DetailAgendaitemAgendaitemsAgendaRoute extends Route {
     }
     const agendaItemTreatment = await model.treatment;
     this.newsletterInfo = await agendaItemTreatment?.newsletterInfo;
+    this.decisionActivity = await agendaItemTreatment?.decisionActivity;
+    await this.decisionActivity?.decisionResultCode;
     // When routing here from agenda overview with stale data, we need to reload several relations
     // The reload in model refreshes only the attributes and includes relations, makes saves with stale relation data possible
     await model.hasMany('mandatees').reload();
--- a/app/templates/agenda/agendaitems/agendaitem/index.hbs
+++ b/app/templates/agenda/agendaitems/agendaitem/index.hbs
@@ -1,3 +1,23 @@
+{{#let this.model.treatment.decisionActivity.decisionResultCode as |drc|}}
+  {{#if drc.isRetracted}}
+    <Auk::Alert @skin="warning" @message={{t "retracted-item"}}/>
+  {{/if}}
+  {{#if drc.isPostponed}}
+    <div class="auk-alert auk-alert--warning">
+      <div class="auk-alert__body">
+        <Auk::Badge @icon="alert-triangle" @skin={{"warning"}}/>
+        <div class="auk-alert__text">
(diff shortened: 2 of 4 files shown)


app/routes/agenda/agendaitems/agendaitem/index.js
```diff
@@ -31,6 +31,8 @@ export default class DetailAgendaitemAgendaitemsAgendaRoute extends Route {
     }
     const agendaItemTreatment = await model.treatment;
     this.newsletterInfo = await agendaItemTreatment?.newsletterInfo;
+    this.decisionActivity = await agendaItemTreatment?.decisionActivity;
+    await this.decisionActivity?.decisionResultCode;
```
erikap, line 35: Is it necessary to load the data here?  It will automatically be fetched by using `this.model.treatment.decisionActivity.decisionResultCode` in the template?

app/templates/agenda/agendaitems/agendaitem/index.hbs
```diff
@@ -1,3 +1,23 @@
+{{#let this.model.treatment.decisionActivity.decisionResultCode as |drc|}}
```
erikap: Use a less cryptic variable name `resultCode` or `code`?

app/templates/agenda/agendaitems/agendaitem/index.hbs
```diff
@@ -1,3 +1,23 @@
...
+  {{#if drc.isRetracted}}
+    <Auk::Alert @skin="warning" @message={{t "retracted-item"}}/>
+  {{/if}}
+  {{#if drc.isPostponed}}
```
erikap: `{{else if}}`?

erikap (changes requested):

app/templates/agenda/agendaitems/agendaitem/index.hbs
ValenberghsSven: since there is only 1 resultCode possible I didn't think there was a need for an `else if` here.
If neither are true we show nothing.

Will change it for clarity

app/templates/agenda/agendaitems/agendaitem/index.hbs
ValenberghsSven: yep

app/routes/agenda/agendaitems/agendaitem/index.js
ValenberghsSven: Only having it in the template would work, but that would be an "await in the template" even though we don't use `await` explicitly?. Thought it would be better off being loaded in the `afterModel`.
In later stories I am using the tracked `decisionActivity` in local methods.

Also, we could now pass down this as args to `agendaitemControls`, removing the need for a separate loading task in that component.

erikap (approved):

PR #1487, opened 2022-09-26
KAS-3520: government fields register

sergiofenoll:
https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/316

Small tweak to two URIs

### DIFF
--- a/app/components/utils/government-areas/area-selector/government-area-selector-form.js
+++ b/app/components/utils/government-areas/area-selector/government-area-selector-form.js
@@ -45,7 +45,7 @@ export default class GovernmentAreaSelectorForm extends Component {
     this.domainSelections = uniqueDomains.map((domain) => {
       const availableFieldsForDomain = availableFields.filter(
         (_, index) => domainsFromAvailableFields[index] === domain
-      );
+      ).sortBy('position');
       const selectedFieldsForDomain = selectedFields.filter(
         (_, index) => domainsFromSelectedFields[index] === domain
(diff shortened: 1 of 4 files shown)


tdn (commented):
Any particular reason we need to change this to the English variant?

Since https://github.com/kanselarij-vlaanderen/kaleidos-documentation/blob/master/data-model/uri-formatting.md specifies "The concept name must always be lowercase, singular, and in Dutch".

Edit: huh. I see we use 'scheme' for the mandatees, but we do use 'schema' in the resource base https://github.com/kanselarij-vlaanderen/app-kaleidos/blob/development/config/resources/concept-domain.json#L82

sergiofenoll:
Ah, I didn't realize that! I based myself off what seems to be used in Themis, e.g.: https://themis.vlaanderen.be/id/concept-scheme/e93481b2-342d-468b-8def-2d629232d3a5 (though this might also be the exception rather than the rule).

In any case this is a super minor change, I can just close this PR and revert the changes in the backend.

tdn (approved):
Apparently `concept-scheme` is what we're using most frequently, so if we want to change that, let's do it in a different story for all data. So this one can go through

erikap:
> Apparently `concept-scheme` is what we're using most frequently, so if we want to change that, let's do it in a different story for all data. So this one can go through

We need to take care for concept-schemes that have already been published on Themis. We cannot just change the URI for those ones.

PR #1489, opened 2022-09-28
KAS-3575: Legacy support report generation

sergiofenoll:
https://github.com/kanselarij-vlaanderen/publication-report-service/pull/6

### DIFF
--- a/app/components/publications/overview/reports/generate-report-modal.js
+++ b/app/components/publications/overview/reports/generate-report-modal.js
@@ -198,7 +198,10 @@ export default class GenerateReportModalComponent extends Component {
       'filter[top-concept-schemes][:uri:]':
         CONSTANTS.CONCEPT_SCHEMES.BELEIDSDOMEIN,
       'filter[:has-no:broader]': true, // only top-level government-domains
-      ...((this.publicationYear >= 2022) && { 'filter[deprecated]': false }), // Exclude deprecated government domains when generating modern reports
+      ...(this.publicationYear >=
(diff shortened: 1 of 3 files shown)


app/components/publications/overview/reports/generate-report-modal.js
```diff
@@ -198,7 +198,7 @@ export default class GenerateReportModalComponent extends Component {
       'filter[top-concept-schemes][:uri:]':
         CONSTANTS.CONCEPT_SCHEMES.BELEIDSDOMEIN,
       'filter[:has-no:broader]': true, // only top-level government-domains
-      'filter[deprecated]': false,
+      ...((this.publicationYear >= 2022) && { 'filter[deprecated]': false }), // Exclude deprecated government domains when generating modern reports
```
erikap: Can we define the magic value `2022` in the config file and import it here?

erikap: What is considered a 'modern' report?  One that is generated after the codelist of beleidsdomein has been finalized?  (trying to find a meaningful name for the variable in the config file)

sergiofenoll: 2022-03-02 is when OVRB started marking beleidsdomeinen using the current list (which only contains the non-deprecated beleidsdomeinen). Don't know if that makes coming up with a name easier.

erikap: `PUBLICATIONS_IN_KALEIDOS_START_DATE` ?  :woman_shrugging:

PR #1491, opened 2022-09-29
KAS-2008: Update the access levels of agendaitem pieces when retracting

sergiofenoll:


app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -47,6 +49,18 @@ export default class AgendaitemDecisionComponent extends Component {
     this.previousReport = yield this.report?.previousPiece;
   }
 
+  @action
+  async onSaveDecisionResultCode() {
+    const decisionResultCode = await this.args.decisionActivity.decisionResultCode;
+    if ([CONSTANTS.DECISION_RESULT_CODE_URIS.UITGESTELD, CONSTANTS.DECISION_RESULT_CODE_URIS.INGETROKKEN].includes(decisionResultCode.uri)) {
+      const pieces = await this.args.agendaitem.get('pieces');
```
erikap: I assume `get()` is required since `this.args.agendaitem` can be a promise? Can we try to avoid to pass promise objects as arguments?

app/services/piece-access-level-service.js
```diff
@@ -83,6 +83,24 @@ export default class PieceAccessLevelService extends Service {
     }
   }
 
+  /**
+   * Strengthen the access level of the piece to intern regering (unless the
+   * access level was already higher) and then update all the previous access
+   * levels.
+   */
+  async strengthenAccessLevel(piece) {
```
erikap: Since the target access level is fixed in the method maybe better reflect it in the name? `strengthenAccessLevelToInternRegering`?

app/services/piece-access-level-service.js
```diff
@@ -83,6 +83,24 @@ export default class PieceAccessLevelService extends Service {
...
+    const accessLevel = await piece.accessLevel;
+    if ([CONSTANTS.ACCESS_LEVELS.INTERN_SECRETARIE, CONSTANTS.ACCESS_LEVELS.VERTROUWELIJK].includes(accessLevel.uri)) {
+      return;
```
erikap: Can you inverse the if-condition to avoid preliminary returns?

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -47,6 +49,18 @@ export default class AgendaitemDecisionComponent extends Component {
     this.previousReport = yield this.report?.previousPiece;
   }
 
+  @action
+  async onSaveDecisionResultCode() {
```
erikap: The component hook is called `on...` but the action implementing the hook should not start with `on...` but reflect what will happen (eg. `updateDocumentAccessLevels`)

erikap (changes requested):

app/components/agenda/agendaitem/agendaitem-decision.js
sergiofenoll: Will change. In this case it doesn't really matter because we're just awaiting the relation. But does "avoid passing promise objects as arguments" also apply when we're e.g. using the relations in the template (there, it being a PromiseProxy is actually useful because we can then rely on it for reactivity purposes, at least that's how I understand Ember, but I could be wrong)?

app/templates/print-overviews/decisions/agendaitems.hbs
```diff
@@ -12,10 +12,15 @@
...
-    {{#let (await agendaitemRow.agendaActivity.subcase) (await agendaitemRow.treatment.decisionActivity) as |subcase decisionActivity|}}
+    {{#let
+      (await agendaitemRow.agendaActivity.subcase)
+      (await agendaitemRow.treatment.decisionActivity)
+      as |subcase decisionActivity|
+    }}
       {{#if (and subcase decisionActivity)}}
         <div class="auk-u-m-8">
           <Agenda::Agendaitem::AgendaitemDecision
+            @agendaitem={{agendaitemRow.content}}
```
erikap, line 23: Weird that the model hook returs a promise object instead of a resolved promise, but as already mentioned on other PRs as well, the print-overview pages need a major refactor so going to leave this as it is now.

erikap (approved):

app/templates/print-overviews/decisions/agendaitems.hbs
sergiofenoll: I *think* it's because the component we use here for the table, ember-light-table, wraps all the rows in `ProxyObjects`. So unrelated to the model hook, but the library's doing. https://github.com/adopted-ember-addons/ember-light-table/blob/master/addon/classes/Row.js#L12

PR #1494, opened 2022-09-30
Remove unexisting route from router.js

erikap:
Remove unexisting `print-overviews.newletter` route from `router.js`. 
Should be a no-brainer, but awaiting tests to be sure.


ValenberghsSven (changes requested):
one more reference:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/m-header.hbs#L30

PR #1495, opened 2022-09-30
KAS-3623: Perform checks before trying to send mail campaign

sergiofenoll:
Perform three checks before trying to send a mail campaign. If the checks fail, we don't even try to send the campaign and instead show a red error toast.

Toast message isn't the best, maybe we want to show a toast per failed condition instead.
image


app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +189,34 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+    for (const agendaitem of agendaitems.toArray()) {
+      const agendaitemTreatment = await agendaitem.treatment;
+      const newsletterItem = await agendaitemTreatment.newsletterInfo;
+      const themes = await newsletterItem?.themes;
+      if (themes?.length) {
+        hasThemes = true;
+      } else {
+        hasThemes = false;
+        break
+      }
```
ValenberghsSven: The ticket is not specific enough regarding themes.
The mailchimp campaign will not send if there are 0 themes in total. (no audience to send to since nothing matches the "interests" chosen by the subscribers)
So as long as there is 1 theme (on 1 or 100 agendaitem makes no difference) then it should work.
So can be simplified to just ensure there is at least 1.

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +189,34 @@ export default class NewsletterHeaderOverviewComponent extends Component {
     this.showConfirmPublishAll = false;
   }
 
+  async canSendMailCampaign() {
+    const documentPublicationActivity = await this.args.meeting.internalDocumentPublicationActivity;
+
+    let hasDocumentPublicationDate = isPresent(documentPublicationActivity.plannedDate);
```
ValenberghsSven: I think you need the themis publication activity.
That is what we use in the newsletter service: 
https://github.com/kanselarij-vlaanderen/newsletter-service/blob/master/util/query-helper.js#L212

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +189,34 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+      if (themes?.length) {
+        hasThemes = true;
+      } else {
+        hasThemes = false;
+        break
+      }
+
+      const type = await agendaitem.type;
+      if (type.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA) {
+        hasNotas = true;
```
ValenberghsSven: when there is 1 nota, you can break the loop.
But since the loop also checks the themes, you should break the for loop if both `hasThemes` and `hasNotas` are true at any stage.

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +189,34 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
   }
 
+  async canSendMailCampaign() {
+    const documentPublicationActivity = await this.args.meeting.internalDocumentPublicationActivity;
+
+    let hasDocumentPublicationDate = isPresent(documentPublicationActivity.plannedDate);
+    let hasNotas = false;
+    let hasThemes = false;
+
+    const agendaitems = await this.args.agenda.agendaitems;
```
ValenberghsSven: This is a wonky relation (easily stale) + this will only load up to 20 agendaitems due to default page size limits.
We typically never use it directly.

We could use the same query as used in the child route:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/routes/newsletter/print.js#L27

ValenberghsSven: That query is already setup to filter out announcements when it's a draft so you would only need to check if 1 theme exists.

ValenberghsSven (changes requested):

app/components/newsletter/newsletter-header-overview.js
ValenberghsSven: check the other comments too, if we query we could already filter on notas.

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +190,40 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+          ':id:': this.args.agenda.id,
+        },
+        'type': {
+          ':uri:': CONSTANTS.AGENDA_ITEM_TYPES.NOTA,
+        },
+      },
+      'page[size]': PAGE_SIZE.AGENDAITEMS,
+    });
+    for (const agendaitem of agendaitems.toArray()) {
+      hasNotas = true;
```
erikap: Can this be set outside the for-loop as `const hasNotas = agendaitems.length > 0`?

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +190,40 @@ export default class NewsletterHeaderOverviewComponent extends Component {
     this.showConfirmPublishAll = false;
   }
 
+  async canSendMailCampaign() {
+    let themisPublicationActivities = await this.args.meeting.themisPublicationActivities;
+    let themisPublicationActivity = themisPublicationActivities.find((activity) => activity.scope.includes('documents'));
```
erikap: Maybe better to use `this.store.query` to ensure we're not looking at stale data?

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +190,40 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+      hasNotas = true;
+
+      const agendaitemTreatment = await agendaitem.treatment;
+      const newsletterItem = await agendaitemTreatment.newsletterInfo;
+      const themes = await newsletterItem?.themes;
+      if (themes?.length) {
+        hasThemes = true;
+        break;
+      }
+    }
```
erikap: Execute a `this.store.count`-query on `newsletter-info` with a filter on agenda-id and `:has:themes`? Or am I missing something?

app/components/newsletter/newsletter-header-overview.js
erikap: Use a value imported from constants instead of hardcoded `documents` value (already used in other places in this file)

app/templates/newsletter.hbs
```diff
@@ -1,4 +1,5 @@
 <Newsletter::NewsletterHeaderOverview
-  @meeting={{this.model.meeting}}
+  @meeting={{@model.meeting}}
+  @agenda={{@model.agenda}}
```
erikap: Do we need both arguments?  Meeting can be uniquely derived from the agenda?
Argument documentation of `Newsletter::NewsletterHeaderOverview` needs to be updated

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -77,25 +79,32 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
-          } catch(e) {
-            console.log("error sending newsletter", e);
-            this.toaster.error(
-              this.intl.t('error-send-newsletter'),
-              this.intl.t('warning-title')
-            );
+      if (yield this.canSendMailCampaign()) {
+        yield this.ensureMailCampaign();
+
+        if (this.mailCampaign?.isSent) {
```
erikap: I know this was already here before but `?` is a bit silly right after doing `this.ensureMailCampaign()`

erikap (changes requested):

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -181,6 +190,41 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+        type: {
+          ':uri:': CONSTANTS.AGENDA_ITEM_TYPES.NOTA,
+        },
+        treatment: {
+          'newsletter-info': {
+            ':has:themes': true,
+          },
+        },
+      },
+      'page[size]': PAGE_SIZE.AGENDAITEMS,
```
erikap: Page size is not needed for `this.store.count()`

ValenberghsSven: I removed this now.

ValenberghsSven:
Took over a little bit.
The code technically worked with themes and nota in 1 query but was not picking up themes going from 0 to 1 because the count is cached for `agendaitem` so any saves on `newsletter-info` do not trigger that cache to be cleared.

There is 1 possible issue that was not mentioned in the ticket:
If there is only 1 nota and a `newsletter-info` with 1 theme it should be fine, but that one `newsletter-info` also needs `inNewsletter` to be true. The backend service covers this and will throw an error in that case.
The use case seems so small that we probably don't need to cover that...

erikap (approved):

erikap:
Failing test not related to this PR. Going to merge :+1:

PR #1502, opened 2022-10-11
Feature/kas 3612 subcase views with reproposing

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3612

The requirements of the ticket are done.
However there are still some features that need looking at:
- When removing a BIS on a postponed agendaitem (normal + when bis is linked to multiple agendas)
- other profile impact (checked mainly "overheid" because of the special rules, but need to make sure other profiles don't see too much or not enough.
- concurrency when editing(add doc to) a subcase that is being postponed/proposed by another user.

Will go into details about what has changed later


app/models/subcase.js
```diff
@@ -20,10 +20,12 @@ export default class Subcase extends ModelWithModifier {
   @belongsTo('concept') agendaItemType;
 
   // inverse: null or serialize: false is used for possible concurrency issues when saving without reloading possible stale relations.
-  @hasMany('agenda-activity', { inverse: null }) agendaActivities;
+  @hasMany('agenda-activity', { inverse: null, serialize: false })
```
erikap, line 23: Can we try to only keep `serialize: false` and remove `inverse: null`?  Same for `decision-activity`.

Since both are 1-to-many relations, I even assume that `inverse: null` and `serialize: false` can both be removed as ember-data will always save the relation at the `belongsTo` side. Can we give that a try?

app/templates/cases/case/subcases/subcase.hbs
```diff
@@ -9,6 +9,7 @@
   <Subcases::SubcaseHeader
     @subcase={{this.model}}
     @onMoveSubcase={{this.refreshSubcases}}
+    @onProposingSubcase={{this.refreshSubcases}}
```
erikap: In line with naming like `didDelete...`, can we rename to `@onProposedForAgenda` to indicate that the proposal has been done already?

app/models/subcase.js
ValenberghsSven: `serialize: false` is mainly for concurrent `subcase` saves while someone else is postponing.
We don't hit the "modified by" check in that case since postponing sets subcase on `agenda-activity` inversely.
maybe `inverse: null` can go, in some relations it has a use but thinking of a scenario where it is needed is difficult.
In some cases we don't want the ember store to update the inverse for reasons that are escaping my mind right now.
Some of those `inverse: null` are also legacy code, so I'm also in favor of trying to remove that one.

app/templates/cases/case/subcases/subcase.hbs
ValenberghsSven: true

app/models/subcase.js
ValenberghsSven: put this in a separate PR for now.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1519

PR #1510, opened 2022-10-18
KAS-3607: Filtering users

sergiofenoll:
# :warning: This PR merges into #1508 
Either merge this one before merging the parent PR, or change the base branch.

Adds filters for user and organizations overview, as well as the link from the organizations overview to the view with all users for said organization.

Related PRs
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/326

Custom Jenkins run: http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/51/


app/components/organization-filter.js
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      this.selected.push(organization);
+    }
+    this.args.onChange?.(this.selected);
+  }
+
+  @task
+  *search(query) {
+    if (query) {
+      yield timeout(500);
```
erikap: There is a `LIVE_SEARCH_DEBOUNCE_TIME` defined in `config.js`

app/components/organization-filter.hbs
```diff
@@ -0,0 +1,53 @@
+<Auk::Label>{{t "filter-on-organization"}}</Auk::Label>
+<div class="au-u-flex au-u-flex--spaced-tiny">
+  <div class="au-u-1-1">
+    <PowerSelectMultiple
```
erikap: Why do we use a multiselect if we only handle 1 selected value in `@onChange`?

app/components/user-role-filter.js
```diff
@@ -0,0 +1,62 @@
...
+      this.selected.splice(this.selected.indexOf(role), 1);
+    } else {
+      this.selected.push(role);
+    }
+    this.args.onChange?.(this.selected);
+  }
+
+  @task
+  *loadData() {
+    this.roles = yield Promise.all([
```
erikap: Why don't we just query all roles with `this.store.query()`? To limit number of requests to the backend? 

Maybe more readable to create a list of roles and map:
```javascript
[ 
  CONSTANTS.USER_ROLES.ADMIN, 
  CONSTANTS.USER_ROLES.SECRETARIE,
  ...
].map((role) => this.store.findRecordByUri('role', role))
```

app/controllers/settings/users/index.js
```diff
@@ -23,6 +25,50 @@ export default class UsersSettingsController extends Controller {
...
+
+  @action
+  setOrganizations(organizations) {
+    this.organizations = organizations.map((organization) => organization.id);
+    this.selectedOrganizations = organizations;
+  }
+
+  @task
+  *setRoles(roles) {
+    yield timeout(500);
```
erikap: Will users be able to check multiple checkboxes in this timeframe? If so, use value from `config.js` instead of hardcoded number.

app/routes/settings/organizations/index.js
```diff
@@ -54,12 +65,15 @@ export default class SettingsOrganizationsIndexRoute extends Route {
...
+  async setupController(controller) {
     super.setupController(...arguments);
 
     if (controller.page !== this.lastParams.committed.page) {
       controller.page = this.lastParams.committed.page;
     }
     controller.searchTextBuffer = this.lastParams.committed.filter;
+
+    const organizationIds = this.lastParams.committed.organizations;
+    controller.selectedOrganizations = await Promise.all(organizationIds.map((id) => this.store.findRecord('user-organization', id)));
```
erikap: Page rendering will not be blocked until `setupController` completed. Maybe better to load this in the `afterModel` hook? Or define as a task on the controller that is called from here, such that we can show a loading state in the template for the `OrganizationFilter` component?

app/routes/settings/users/index.js
```diff
@@ -54,12 +128,32 @@ export default class SettingsUsersIndexRoute extends Route {
...
     controller.searchTextBuffer = this.lastParams.committed.filter;
+
+    const organizationIds = this.lastParams.committed.organizations;
+    controller.selectedOrganizations = await Promise.all(organizationIds.map((id) => this.store.findRecord('user-organization', id)));
+
+    controller.dateFromBuffer = parseDate(this.lastParams.committed.dateFrom);
+    controller.dateToBuffer = parseDate(this.lastParams.committed.dateTo);
+
+    const roleIds = this.lastParams.committed.roles;
+    controller.selectedRoles = await Promise.all(roleIds.map((id) => this.store.findRecord('role', id)));
```
erikap: Page rendering will not be blocked until `setupController` completed. Maybe better to load this in the `afterModel` hook? Or define as a task on the controller that is called from here, such that we can show a loading state in the template for the `OrganizationFilter` and `UserRoleFilter` component?

app/routes/settings/users/index.js
```diff
@@ -54,12 +128,32 @@ export default class SettingsUsersIndexRoute extends Route {
...
+
+  @action
+  loading(transition) {
+    // eslint-disable-next-line ember/no-controller-access-in-routes
+    const controller = this.controllerFor(this.routeName);
+    controller.isLoadingModel = true;
+    transition.promise.finally(() => {
+      controller.isLoadingModel = false;
+    });
+    return true;
```
erikap, line 155: Nice catch that this was missing :+1: 
Should we use the same loading construct for the organizations table?

erikap (changes requested):

app/components/user-role-filter.js
sergiofenoll: The separate requests are purely to have a set order to the roles. I could also give the roles a `schema:position` in the backend, but I don't know if that holds for all uses.

erikap: `schema:position` in backend is fine :+1:

app/components/organization-filter.hbs
sergiofenoll: Remnant of the initial implementation where I was just using the Multi PowerSelect (but that looked quite ugly). I'll re-work it to use the single one since that makes the most sense.

erikap (approved):

PR #1513, opened 2022-10-21
KAS-3607: Filtering (bis)

sergiofenoll:
Some fixes for issues I missed in the first PR.

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/328


app/controllers/settings/users/index.js
```diff
@@ -83,16 +76,6 @@ export default class UsersSettingsController extends Controller {
...
 
-  @task
-  *loadSelectedOrganizations() {
-    this.selectedOrganizations = yield Promise.all(this.organizations.map((id) => this.store.findRecord('user-organization', id)));
-  }
-
-  @task
-  *loadSelectedRoles() {
-    this.selectedRoles = yield Promise.all(this.roles.map((id) => this.store.findRecord('role', id)));
-  }
```
erikap: Can we keep those tasks and just call them from the route's `setupController`? As such, we can still show a loading state in the template, without the entire page rendering being blocked (which is the case when data is being loaded in the `afterModel`-hook).

app/controllers/settings/users/index.js
```diff
@@ -145,4 +128,12 @@ export default class UsersSettingsController extends Controller {
     }
     await Promise.all(this.membershipsBeingBlocked.map((membership) => membership.save()));
   }
+
+  @action
+  filterMemberships(memberships) {
```
erikap, line 148: Can you add a comment for our future selves to clarify why we do the additional filtering in the frontend

app/controllers/settings/users/index.js
```diff
@@ -145,4 +128,12 @@ export default class UsersSettingsController extends Controller {
...
+    return memberships.filter((membership) => {
+      const shouldShow = !this.showBlockedMemberships || membership.get('isBlocked');
```
erikap: I got confused between `showBlockedMemberships` and `showBlockMemberships`. Can we rename the tracked properties to something more distinctive like `showBlockedMembershipsOnly` (for the filter checkbox) and `showBlockMembershipsConfirmationModal` (for the modal flag)?
Same for the other tracked properties in this controller.

erikap (changes requested):

erikap (approved):

PR #1520, opened 2022-10-27
KAS-3675: Themis URIs for beslissing resultaatcodes

sergiofenoll:
Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/332
- [x] https://github.com/kanselarij-vlaanderen/yggdrasil/pull/48
- [x] https://github.com/kanselarij-vlaanderen/app-themis/pull/10

Jenkins runs:
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/316/

PR #1521, opened 2022-10-28
KAS-3663: Warm concepts in the cache

sergiofenoll:
In draft until https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1520 and https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1518 get merged, because they replace some custom types with concepts and those are the calls we cache.

Jenkins runs:
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/322/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/327/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/56/~~
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/57/

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/cache-warmup-service/pull/9
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/334


app/services/store.js
```diff
@@ -60,4 +61,17 @@ export default class ExtendedStoreService extends Store {
       'filter[:uri:]': uri,
     });
   }
+
+  // TODO: Move this to a concept-service where we bundle logic for getting
+  // all concepts of a scheme/specific concepts? Might cut down some of the
+  // repetition across the codebase.
+  queryConceptsForConceptScheme(conceptSchemeUri) {
```
erikap: Probably a good idea to move this to a separate service, since it's becoming a bit too 'model-specific' to be an extension of the generic ember-data store. Another option is a util method that gets the store as argument, but not sure if I like that more.

Can you also use `this.queryAll` instead of `this.query`? This way we're sure we always load the complete codelist. I assume `PAGE_SIZE.CODE_LISTS` can be removed afterwards.

app/components/subcase/subcase-description-panel/subcase-description-edit.js
```diff
@@ -43,12 +44,15 @@ export default class SubcaseDescriptionEdit extends Component {
...
-      sort: '-label',
       filter: {
         'concept-schemes': {
           ':uri:': CONSTANTS.CONCEPT_SCHEMES.AGENDA_ITEM_TYPES,
         }
       },
+      page: {
+        size: PAGE_SIZE.CODE_LISTS,
+      },
+      sort: '-label',
```
erikap: Why not using the `queryConceptsForConceptScheme` method here? Because of the different sorting? Can we make the method accept a second `options` argument in which you can overwrite the `sort` param?

app/services/store.js
ValenberghsSven: I had some issues with `queryAll`, there is an option to pass in the page size but when you do, the outgoing query contains `page:size` twice (in different formats)

app/services/store.js
```diff
@@ -60,4 +61,17 @@ export default class ExtendedStoreService extends Store {
...
+    return this.query('concept', {
+      'filter[concept-schemes][:uri:]': conceptSchemeUri,
+      'filter[:has-no:narrower]': true,
+      include: 'broader,narrower',
+      'page[size]': PAGE_SIZE.CODE_LISTS,
+      sort: 'position',
+    });
+  }
```
ValenberghsSven: This query for concepts is quite specific (this exact query was only used for `VERGADERACTIVITEIT` before).
For generic use, the filter on `has-no:narrower`, the `includes` and `sort` could be dynamic.
```
'filter[:has-no:narrower]': true,
include: 'broader,narrower',
sort: 'position',
```
Right now, the only reason to even sort on `-label` is because of `AGENDA_ITEM_TYPES`. 
If we add `position` there as well, problem sorted? `Nota` first, `Mededeling` second

If we never have to reverse sort on `position` then you can just `sort: 'position'` by default

ValenberghsSven: of course, whatever we do in frontend has to match against the cache from the service.

sergiofenoll: I'm ok with adding this, just a bit concerned that it's going to be annoying down the line. Having the query set in stone behind a method means that there's no way to accidentally miss the cache. If we make it customizable, we risk missing the cache if changes happen in the frontend, but not in the cache-warmup-service.

app/services/concept-store.js
```diff
@@ -0,0 +1,26 @@
+import Service, { inject as service } from '@ember/service';
+
+export default class ConceptStoreService extends Service {
+  @service store;
+
+  allForConceptScheme(conceptSchemeUri, query) {
```
erikap: `queryAllByConceptScheme`?

app/routes/application.js
```diff
@@ -39,27 +38,11 @@ export default class ApplicationRoute extends Route {
...
-    await this.store.query('concept', {
-      filter: {
-        'concept-schemes': {
-          ':uri:': CONSTANTS.CONCEPT_SCHEMES.VERGADERACTIVITEIT,
-        },
-        ':has-no:narrower': true, // Only the most specific concepts, i.e. the actual meeting kinds (so no "Annex")
-      },
+    this.conceptStore.allForConceptScheme(CONSTANTS.CONCEPT_SCHEMES.VERGADERACTIVITEIT, {
+      'filter[:has-no:narrower]': true,
       include: 'broader,narrower',
```
erikap: Including `narrower` while filtering on `:has-no:narrower` looks a bit silly

erikap (changes requested):
Part of the story doesn't seem to be tackled:
> In de frontend wordt gezorgd dat de codelijsten pas ingeladen worden nadat de gebruiker volledig ingelogd is. Nu blokkeert het laden van de codelijsten soms het renderen van de login pagina. (Vb. door het laden van de codelijsten in de model-hook van de index-route te doen en de transition uit de beforeModel-hook te verplaatsen naar afterModel. Er kan dan ook een index-loading.hbs template gemaakt worden om een loading state weer te geven terwijl de codelijsten laden.)

Seems that there are still some places where data loading can be replaced with the new method from `conceptStore` (or is there a particular reason not to do so?):
- `Utils::DocumentTypeSelector`  (the search function in this component can also be refactored to a frontend search in that case)
- Codelists in `afterModel`-hook of `publications` route
- Codelists in `beforeModel`-hook of `publications.overview.search` route (`this.store.peekAll` can be used here since codelists are already loaded in `publications` route. Maybe add a note to clarify)

sergiofenoll:
> Part of the story doesn't seem to be tackled:
> 
> > In de frontend wordt gezorgd dat de codelijsten pas ingeladen worden nadat de gebruiker volledig ingelogd is. Nu blokkeert het laden van de codelijsten soms het renderen van de login pagina. (Vb. door het laden van de codelijsten in de model-hook van de index-route te doen en de transition uit de beforeModel-hook te verplaatsen naar afterModel. Er kan dan ook een index-loading.hbs template gemaakt worden om een loading state weer te geven terwijl de codelijsten laden.)

I attempted to tackle this [here](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/16acee3c1d77c478260c2104931d1c237af0cb44/app/routes/application.js#L41-L45) by removing the `await`s and just letting the promises resolve on their own. The reason for not moving to the index route as suggested in the ticket is that we will only reach that route if the user goes to that route explicitly, either as a redirect from the login or because they went to the bare URL. If for example a user has `/publicaties` open, refreshes, then navigates to the agendas (because there's no way to navigate to the index page itself), the concept loading calls won't have happened.

> Seems that there are still some places where data loading can be replaced with the new method from `conceptStore` (or is there a particular reason not to do so?):
> 
>     * `Utils::DocumentTypeSelector`  (the search function in this component can also be refactored to a frontend search in that case)

:+1:

>     * Codelists in `afterModel`-hook of `publications` route

[These](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/16acee3c1d77c478260c2104931d1c237af0cb44/app/routes/publications.js#L22-L37) should be okay, as in, the necessary calls were added to the warmup service. They're still using custom types which we might want to replace with concepts down the line, but I left them as is for now.

>     * Codelists in `beforeModel`-hook of `publications.overview.search` route (`this.store.peekAll` can be used here since codelists are already loaded in `publications` route. Maybe add a note to clarify)

:+1:

erikap:
> I attempted to tackle this [here](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/16acee3c1d77c478260c2104931d1c237af0cb44/app/routes/application.js#L41-L45) by removing the `await`s and just letting the promises resolve on their own.

Sorry, missed the removal of the `await`. Can you maybe add a comment that we fire-and-forget the requests on purpose?

> Codelists in `afterModel`-hook of `publications` route
[These](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/16acee3c1d77c478260c2104931d1c237af0cb44/app/routes/publications.js#L22-L37) should be okay, as in, the necessary calls were added to the warmup service. They're still using custom types which we might want to replace with concepts down the line, but I left them as is for now.

Oops, you're right :+1:

erikap (approved):

PR #1522, opened 2022-10-29
KAS-3678: Disable old print routes

sergiofenoll:
Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/338


app/models/agendaitem.js
```diff
@@ -20,7 +20,6 @@ export default ModelWithModifier.extend({
   modified: attr('datetime'),
   titlePress: attr('string'),
   textPress: attr('string'),
```
erikap: Can we remove these properties. As far is I can see they are currently only used:
- on creation of an agendaitem
- in agendaitem search

Both occurrences can be removed. 

Can you also open a PR on the backend to put the properties in comments there with a note that data is still available in the triplestore.

app/templates/agenda/print.hbs
```diff
@@ -1,7 +1,55 @@
...
+            "dddd DD MMMM YYYY"
+            "dddd DD MMMM YYYY \o\m HH:mm"
+          )
+        }}
+        -
+        {{kind-print-label @model.meeting.kind}}
+      </h2>
+
+      <div class="auk-content">
+        {{#if (gt this.notaGroups.length 0)}}
```
erikap: Just use `{{#if this.notaGroups}}`

app/templates/agenda/print.hbs
```diff
@@ -1,7 +1,55 @@
...
+          <Agenda::PrintableAgenda::ListSection
```
erikap: I am not so convinced of the usefulness of this template-only component. Can we just embed the HTML here?

app/templates/agenda/print.hbs
```diff
@@ -1,7 +1,55 @@
...
+          >
+            {{#each this.notaGroups as |notaGroup|}}
+              <Agenda::PrintableAgenda::ListSection::ItemGroup
+                @items={{notaGroup}}
+              />
+            {{/each}}
+          </Agenda::PrintableAgenda::ListSection>
+        {{/if}}
+
+        {{#if (gt @model.announcements.length 0)}}
```
erikap: Just use `{{#if @model.announcements}}`

app/controllers/agenda/print.js
```diff
@@ -9,13 +8,10 @@ function equalContentArrays(array1, array2) {
...
 
-// TODO: octane-refactor
-/* eslint-disable ember/no-get */
-// eslint-disable-next-line ember/no-classic-classes
-export default Controller.extend({
+export default class AgendaPrintController extends Controller {
 
-  notaGroups: computed('model.notas.@each.sortedMandatees', function() {
-    const agendaitems = this.get('model.notas');
+  get notaGroups() {
```
erikap, line 13: Can you have a look whether we can use `groupAgendaitemsOnGroupName` of the `agenda-service` here?

app/components/agenda/printable-agenda/list-section/item-group.hbs
```diff
@@ -6,9 +6,23 @@
...
+      {{#let this.items.first.sortedMandatees as |mandatees|}}
+        <h3
+          class="auk-u-mb-3 auk-u-text-uppercase l-printable-agenda__item-group-header
+            {{if (not mandatees) "l-printable-agenda__item-group-header--muted auk-u-keep-colors@print"}}"
+        >{{! auk-u-text-decoration-underline }}
+          {{#if mandatees}}
+            {{#each mandatees as |mandatee|}}
+              {{mandatee.title}}
+              {{#if (not-eq mandatee mandatees.lastObject)}}
+                <br />&
```
erikap: Remove `&`

app/components/agenda/printable-agenda/list-section/item-group.hbs
```diff
@@ -6,9 +6,23 @@
 <div class="l-printable-agenda__item-group">
   <div class="l-printable-agenda__faux-item">
     <div class="auk-u-non-breakable-group@print">
-      <Agenda::PrintableAgenda::ListSection::ItemGroup::ItemGroupHeader
-        @mandatees={{@items.firstObject.sortedMandatees}}
-      />
+      {{#let this.items.first.sortedMandatees as |mandatees|}}
```
erikap: Can we replace the `this.items.first` and `this.items.remainder` that is defined on the component with template-helpers like for example [drop](https://github.com/DockYard/ember-composable-helpers#drop)?

erikap (changes requested):
If possible I would like to cleanup the components under `Agenda::PrintableAgenda` a bit in this PR as well. 

Added some comments on specific lines, but in general I would like to simplify the deep nested component structure and only keep components that do have an added value. Imo the components are going to be easier to understand if we just have a flat structure under `Agenda::PrintableAgenda` 

Possible improvements that could not be related to a diff line in this PR:
- `Agenda::PrintableAgenda::ListSection::ItemGroup::Item` is a template-only component that is used in only 1 place hence it can just be embedded
- `Agenda::PrintableAgenda::ListSection::ItemGroup::Item::Content` contains some unnecessary `await`s in the template

app/components/agenda/printable-agenda/list-section/item-group.hbs
ValenberghsSven: Should we? unsure myself.
image

ValenberghsSven:
> Agenda::PrintableAgenda::ListSection::ItemGroup::Item is a template-only component that is used in only 1 place hence it can just be embedded

It is being used in 2 templates. 
templates/agenda/print.hbs
components/agenda/printable-agenda/list-section/item-group.hbs

app/controllers/agenda/print.js
sergiofenoll: That would require a bit fiddling with the rest of the components + given that `groupAgendaitemsOnGroupsName` is an async function, getting that to output correctly through the `notaGroups` getter doesn't really work. You just get a promise and it looks like Ember doesn't pick it up as an array after it resolves (so it doesn't render anything).
Might be that I'm doing something wrong though.

erikap: Will leave it as it is here, but create a follow up story. Diving deeper in it, it seems there is a lot of overlap in the logic of `agenda.agendaitems` and `agenda.print`. But refactoring the structure of the `agenda` subroutes will lead us to far in this PR.

erikap:
Failing test not related to this PR. Going to merge.

PR #1527, opened 2022-11-04
Unauthorized error message on 403 Forbidden on mock-login

erikap:
More clear error message on mock-login when user doesn't have access to the application (for example because account is blocked)


erikap:
Failing tests not related to this PR.

PR #1528, opened 2022-11-07
Bugfix/kas 3717 new subcase documents invisible

ValenberghsSven:
## :information_source: started from ACCEPTANCE

# bugfix for ACCEPTANCE

https://kanselarij.atlassian.net/browse/KAS-3717

subcase-documents:
- added a query for any submissions not on agendas (always happens)
- query for those from latest meeting get added to the "set" of submissions (both queries should be mutually exclusive so no explicit "`Set` array" needed)
- added a tooltip to inform the users this view is filtered. (QUESTION: should we show this to all users or just editors? This is a last minute addition and is not validated by designers)

agendaitem-postponed:
- added a query for any submissions "not on agenda" to add to the previous query of submissions of the current meeting. 

nl-be:
- translation string added

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-postponed.js
+++ b/app/components/agenda/agendaitem/agendaitem-postponed.js
@@ -68,12 +68,20 @@ export default class AgendaitemPostponed extends Component {
 
   @task
   *reProposeForAgenda(meeting) {
-    const submissionActivities = yield this.store.query('submission-activity', {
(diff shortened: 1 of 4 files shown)


tdn:
> added a tooltip to inform the users this view is filtered. (QUESTION: should we show this to all users or just editors? This is a last minute addition and is not validated by designers)

I'd leave it in for all users.

tdn (approved):
Haven't tested this, but looks like it would mitigate the edge case just fine

PR #1530, opened 2022-11-10
Replace ALL non-breaking spaces with normal spaces

sergiofenoll:
The editor inserts `&nbsp;` when making selected text bold. These insertions are erroneous, and will get fixed at some point. In the meantime, we will replace all non-breaking spaces with normal spaces, to ensure that text flows as expected when shown in the Kaleidos app and in the Mailchimp mails.

The editor should not insert non-breaking spaces anymore, except when multiple spaces are typed one after another, so it's OK to replace all non-breaking spaces with normal spaces.

Jenkins run:
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/53/


erikap (approved):
Very well documented :+1:

PR #1532, opened 2022-11-14
KAS-3469 refactor all computed uses and model

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3469

Need to test the "setPreviousPieces" method.
rest looked fine.

### DIFF
--- a/app/components/documents/linked-document-link.js
+++ b/app/components/documents/linked-document-link.js
@@ -22,12 +22,13 @@ export default class LinkedDocumentLink extends Component {
 
   @task
   *loadData() {
-    const containerPieces = yield this.args.documentContainer.sortedPieces;
+    const containerPieces = yield this.args.documentContainer.pieces;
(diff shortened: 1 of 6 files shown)


erikap (approved):

erikap:
Tests succeeded. Can we merge @ValenberghsSven or is there still something you want to test manually?

ValenberghsSven:
Not sure if the replace previous document method for subcase is tested by the suite.
Might only be from the agendaitem views.

ValenberghsSven:
@erikap Tested locally with a normal flow (only 1 agenda-activity) 

Added some more comments about this block of code + removed the check on `this.subcase`, we are on a subcase route so should always exists.
I think the code could be better with some queries.
I think this feature is barely used (using batch edit on subcase documents to delete a BIS/TER piece) so not really urgent.
Will create a follow-up refactor ticket later.

PR #1534, opened 2022-11-15
KAS-3470 refactor mandatee model and computeds, cleanup person model deprecations

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3470

- removed `inverse: null` from `hasMany` relations. Don't think we need it.
- changed computed `fullDisplayName` to a getter.
It is called in 2 places without `await`.
In both cases, the mandatee model is being called via the mandatee service / getMandateesActiveOn method.
If called this way, `mandatee.person` and `mandatee.role` was already included.
- changed computed `priorityAlpha` to local fix in `agendaitem-utils`.

- Also solved the deprecations in `person` model


ValenberghsSven:
Tests were failing after I removed the `inverse:null` because there are 2 connections between mandatee and subcase.
To fix it I can either: define the inverses correctly or keep the `inverse:null`.
In any case, the mandatee model is static and can't be saved so option 1 should be fine.
If we ever decide to make this model saveable, all those `hasMany` relations should have `serialize: false´.

erikap (approved):

ValenberghsSven:
ah those awaits. I just went in for a quick name replace and didn't bother sidetracking :D

PR #1535, opened 2022-11-16
Change base Docker image to a static file host

erikap:
Since frontend has moved behind the dispatcher service in the stack a simple Nginx static file service is sufficient to host the frontend. No redirection logic needs to be applied by the frontend since dispatcher is responsible for that now.

See also https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/336


erikap:
Failing tests not related to this PR. Going to merge. Thanks for the help @sergiofenoll  :pray:

PR #1529, opened 2022-11-09
KAS-3658: Source file as main file

sergiofenoll:
- Rename usage of `primarySource` to `source` and `primarySourceOf` to `derived`
- Consider `source-file` as main file, so use checks for `derived` instead of `source`
- Fix bug with replacing files. We should first replace the file, then delete. Because if you delete it first and then try to replace a link to the file, `mu-authorization` will throw errors (because it can't find type info about the resource you're overwriting, I think)
- ~~Don't show file extension unless it's explicit which file you're talking about. In `document-card` it's somewhat vague, so leave out the extension there.~~ Always show the extension of the "main file" (derived file if it exists, otherwise source file)

Jenkins run(s):
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/323/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/324/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/325/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/61/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/62/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/65/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/66/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/67/~~ Flaky test? Can't immediately repro the failing test locally
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination_2/68/ :heavy_check_mark: all green :)

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/335
- [x] https://github.com/kanselarij-vlaanderen/yggdrasil/pull/49
- [x] https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/3
- [x] https://github.com/kanselarij-vlaanderen/themis-export-service/pull/11

PR #1536, opened 2022-11-17
KAS-3659: Automatic file conversion

sergiofenoll:
# :warning: This merges into #1529! Make sure to double check the base branch before merging :waffle: 

Adds automatic file conversion for uploaded files to the following routes:
- Agenda > Documenten
- Agenda > Agendapunt (detail) > Documenten
- Agenda > Agendapunt (detail) > Beslissing
- Dossier > Procedurestap > Documenten
- Document

When creating a new piece, before saving, files are sent to the `docx-conversion-service` to generate a derived file. Although the service ensures it only tries to convert files it's meant to handle (DOCX files), we also do this check in the frontend to cut out on superfluous requests.

Additionally, the DocumentCard component is now optionally editable, this way the Dossier > Procedurestap > Beslissing route works as expected.

Jenkins runs:
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/326/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/328/~~
- ~~http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/329/~~
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/331/

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/337
- [x] https://github.com/kanselarij-vlaanderen/file-conversion-poc/pull/1


app/config/config.js
```diff
@@ -27,3 +27,9 @@ export const ESTIMATED_PUBLICATION_DURATION = 30 * 60 * 1000;
 export const PUBLICATION_ACTIVITY_REFRESH_INTERVAL_MS = 60 * 1000;
 
 export const DOCUMENT_DELETE_UNDO_TIME_MS = 15000;
+
+export const SUPPORTED_MIME_TYPES = [
```
erikap: Integrate `DOCUMENT_CONVERSION` or something like that in the variable name. It's not clear now what it's used for.

app/services/file-service.js
```diff
@@ -77,6 +81,31 @@ export default class FileService extends Service {
     record.aboutToDelete = false;
   }
 
+  async convertSourceFile(sourceFile) {
+    if (!SUPPORTED_MIME_TYPES.some((mimeType) => sourceFile.format.includes(mimeType))) {
+      // The source file MIME type is unsupported, don't try to send it to the backend
+      return
```
erikap: Missing ;   (also more a fan of positive if-clause instead of prelimenary return)

app/services/file-service.js
```diff
@@ -77,6 +81,31 @@ export default class FileService extends Service {
...
+    }
+
+    const response = await fetch(`/files/${sourceFile.id}/convert`, {
+      method: 'POST',
+      headers: {
+        'Accept': 'application/vnd.api+json',
+        'Content-Type': 'application/vnd.api+json',
+      },
+    });
+    const result = await response.json();
```
erikap: Are we sure `.json()` will always work in case of `!response.ok`?

app/services/file-service.js
```diff
@@ -77,6 +81,31 @@ export default class FileService extends Service {
...
+    if (!response.ok) {
```
erikap: Since there is an if- and else-branch, inverse to positive if-clause to minimize brain overload?

erikap (commented):
Some minor comments. Looks good overall :+1:

app/services/file-service.js
sergiofenoll: Just wondering, how come? I prefer the inverse to prevent the extra level of nesting :smile: 
I'm okay with changing it, just curious.

erikap: A more natural/logical way of reading the code. For a more elaborate argumentation :arrow_right: https://www.youtube.com/watch?v=6wDoopbtEqk (referenced to me by @madnificent)

app/services/file-service.js
sergiofenoll: Goh... good point. In theory a response from the service should always return JSON(:API), but other breakages (if the backend can't be reached or it's overloaded or ...) won't return JSON. I guess we should handle errors separately (i.e., if there's JSON, use it to throw the error, other wise throw a generic error containing the status code).

sergiofenoll:
Failing test here is (I believe) due to the DOCX conversion service not having access to the files, that's the only time I've seen 400's returned like that locally. Will try and debug that.

sergiofenoll:
I don't understand why the `new-document-viewer` test fails. The commands executed by Cypress on Jenkins don't match the commands executed for me locally.

This is from Cypress running locally, command 67 is checking `contains new name test.docx`, which is what I'd expect to see. But on Jenkins that same command seems to check `contains new name test.pdf`.
image

erikap:
> I don't understand why the `new-document-viewer` test fails. The commands executed by Cypress on Jenkins don't match the commands executed for me locally.
> 
> This is from Cypress running locally, command 67 is checking `contains new name test.docx`, which is what I'd expect to see. But on Jenkins that same command seems to check `contains new name test.pdf`. 

Are you sure you're running the same version as Jenkins? Looking at [the test file on this PR branch](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3659-automatic-file-conversion/cypress/integration/unit/new-document-viewer.spec.js#L194-L213) it seems to me that all tests on `document.previewDetailsTab.sourceFile` check for `new name test.pdf`.

sergiofenoll:
> > I don't understand why the `new-document-viewer` test fails. The commands executed by Cypress on Jenkins don't match the commands executed for me locally.
> > This is from Cypress running locally, command 67 is checking `contains new name test.docx`, which is what I'd expect to see. But on Jenkins that same command seems to check `contains new name test.pdf`.
> 
> Are you sure you're running the same version as Jenkins? Looking at [the test file on this PR branch](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-3659-automatic-file-conversion/cypress/integration/unit/new-document-viewer.spec.js#L194-L213) it seems to me that all tests on `document.previewDetailsTab.sourceFile` check for `new name test.pdf`.

:facepalm: I was running the tests locally on the 2742 branch, where the conditions are correct

erikap:
Failing test in latest Jenkin run is not related to this PR. Going to merge :tada:

PR #1541, opened 2022-11-21
KAS-3695: Date helpers

sergiofenoll:
Adds the following helpers based on https://date-fns.org/ to replace `moment-format`:
- `time`: Displays hours and minutes, equivalent to `HH:mm` of moment formatting
- `date`: Displays day-month-year, equivalent to `DD-MM-YYYY` of moment formatting
- `datetime`: Displays date and time, equivalent to `DD-MM-YYYY HH:mm` of moment formatting
- `datetime-at`: Displays date at time, equivalent to `DD-MM-YYY [om] HH:mm` of moment formatting. `om` can be optionally replaced with whatever string you want (e.g. in the codebase we sometimes use `-`) by passing in a second argument
- `date-phrase`: Displays date with full month written out, equivalent to `DD-MMMM-YYYY` of moment formatting.
- `datetime-at-phrase`: Displays date at time with full month written out, equivalent to `DD-MMMM-YYYY [om] HH:mm`, `om` can also optionally be replaced
- `date-format`: Takes in a date and a [date-fns format string](https://date-fns.org/v2.29.3/docs/format) and spits out a formatted date

`date-format` is only used 6 times for 2 format strings:
1. Prepend a datetime with the first 2 letters of the day of the week (used 5 times over 2 files)
2. Prepend a date with the day of the week written out in full (used once)
Given that these usages are rare, I decided to add the `date-format` helper and not move the format strings to separate helpers. If desired we can do the opposite though (remove `date-format` and add two extra helpers for these two cases)

We might also want to check if `datetime-at` should take a `om` parameter is necessary, or if we want to settle on a single format.


app/helpers/date-format.js
```diff
@@ -0,0 +1,12 @@
...
+import { format } from 'date-fns';
+import { dateHelper } from 'frontend-kaleidos/utils/date-helper';
+
+export default helper(function dateFormat([dateOrString, formatString]) {
+  if (formatString) {
+    const dateObject = dateHelper(dateOrString);
+    if (dateObject) {
+      return format(dateObject, formatString);
+    }
+  }
```
erikap: It's maybe a bit more clean (and less code duplication) if we extract this snippet into a util instead of only the `dateHelper`. As such it can be imported in the other helpers and used like
```javascript
export default helper(function datePhrase([dateOrString]) {
  return dateFormat(dateOrString, 'd MMMM yyyy');
});
```

erikap (approved):

PR #1547, opened 2022-11-23
KAS-3736 fix: archiveCase use decisionmakingFlow, proxies and router …

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3736

## :warning: started from ACCEPTANCE

- fixed `archiveCase` not working with case > decisionmakingFlow > subcases
- fixed `archiveCase` using `then` promise when not really needed.
- using `get('id')` because our row data is a proxy and gave errors. (comments might be overkill though, just for safety ..)
- used `findRecord` in `unarchiveCase` to avoid getting decisionmakingFlow from a proxy (also saving proxy errors).
- used `router.refresh` in both instead of `send('...')` , removed action from route
- added refresh to unarchiving for consistency.

Will add this feature to the testing TODO pile for regression.

### DIFF
--- a/app/controllers/cases/overview.js
+++ b/app/controllers/cases/overview.js
@@ -68,31 +68,32 @@ export default class CasesCaseSubcasesOverview extends Controller {
 
   @action
   async archiveCase() {
-    const caseModel = await this.store.findRecord('case', this.selectedCase.id);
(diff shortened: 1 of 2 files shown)


ValenberghsSven:
only flaky failed tests

PR #1549, opened 2022-11-25
KAS-2491: User impersonation

sergiofenoll:
Adds the ability to impersonate users to the app.

- The dropdown in the top right corner has an extra option to select which user (profile) to impersonate
- When impersonating a user the data received is the same as what that user would see
- The user's role is also inherited, so any UI elements that are conditionally rendered based on user role also update accordingly
- Both in the frontend and backend, impersonation is only available for admin users. In the frontend that's done by hiding the impersonation selection for non-admins
- Impersonating a user refreshes the current page you're on so you can immediately see what said user would see on that route. This also means that certain features like default homepage require an extra step to simulate (e.g. OVRB profiles navigate to `/publicaties` by default after log in, to see this with impersonation you would have to impersonate OVRB and then navigate to the root of the app manually)
- Impersonation exists behind the `ENABLE_IMPERSONATION` feature flag. It's enabled by default in development mode as well as on ACC. It should remain toggled off on PROD, as this feature is not desired in that environment.

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/343
- [x] https://github.com/kanselarij-vlaanderen/impersonation-service/pull/1

Jenkins runs:
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/333/


config/environment.js
```diff
@@ -40,6 +40,7 @@ module.exports = function (environment) {
     APP: {
       // eslint-disable-next-line quotes
       ENABLE_SIGNATURES: '{{ENABLE_SIGNATURES}}',
+      ENABLE_IMPERSONATION: '{{ENABLE_IMPERSONATION}}',
```
erikap, line 43: In `Dockerfile` best nog een default zetten voor `EMBER_ENABLE_IMPERSONATION` zodat feature disabled is in builded images by default  (zie ook `EMBER_ENABLE_SIGNATURES` in `Dockerfile`)

app/services/session.js
```diff
@@ -25,4 +26,9 @@ export default class ExtendedSessionService extends SessionService {
       this.router.transitionTo('login');
     }
   }
+
+  async invalidate() {
+    await this.impersonation.stopImpersonation();
```
erikap: Can't this be added to `currentSession.clear()` which is called in `handleInvalidation()` in this service? As such all impersonation-related data loading/clearing is managed by the `current-session` service.

app/services/impersonation.js
```diff
@@ -0,0 +1,82 @@
...
+        if (impersonatedRoleId) {
+          this.role = await this.store.findRecord('role', impersonatedRoleId);
+        }
+      } else {
+        throw new Error('An exception occurred while loading impersonation data: ' + JSON.stringify(result.errors));
+      }
+    }
+  }
+
+  clear() {
```
erikap: What's the difference between `clear()` and `stopImpersonation()`?  Shouldn't the logic of both always be executed to keep frontend and backend impersonation state in sync?

app/services/current-session.js
```diff
@@ -38,21 +44,34 @@ export default class CurrentSessionService extends Service {
...
     return this.role && findGroupByRole(this.role.uri);
   }
 
   get isAdmin() {
     return this.userGroup.name == 'ADMIN';
   }
+
+  get isReallyAdmin() {
+    return this.ownUserGroup.name == 'ADMIN';
+  }
```
erikap: If not used in this PR, I wouldn't introduce this one since we want to move away from this kind of methods and consistently use permissions instead.

app/services/current-session.js
```diff
@@ -38,21 +44,34 @@ export default class CurrentSessionService extends Service {
...
+    this.impersonation.clear();
   }
 
   may(permission) {
     return this.userGroup?.permissions.includes(permission);
   }
 
+  reallyMay(permission) {
+    return this.ownUserGroup?.permissions.includes(permission);
+  }
```
erikap: Isn't it more elegant to add an extra option to `may(permission)` which is `false` by default to validate on the impersonator role/group instead? Something like `may(permission, checkImpersonator = false)`

app/services/current-session.js
```diff
@@ -38,21 +44,34 @@ export default class CurrentSessionService extends Service {
...
+
   hasAccessToApplication() {
     return isPresent(this.role);
   }
 
   get userGroup() {
+    return this.impersonatedRole && findGroupByRole(this.impersonatedRole.uri);
```
erikap: Shouldn't we fallback to `this.role` here in case the user is currently not impersonated?  Or am I missing something?

app/services/current-session.js
```diff
@@ -7,11 +7,16 @@ import { findGroupByRole } from 'frontend-kaleidos/config/permissions';
 export default class CurrentSessionService extends Service {
   @service session;
   @service store;
+  @service impersonation;
 
   @tracked user;
-  @tracked role;
   @tracked organization;
   @tracked membership;
+  @tracked ownRole;
```
erikap: Doesn't seem to be used?  Shouldn't this still be `@tracked role`?

app/services/current-session.js
```diff
@@ -7,11 +7,16 @@ import { findGroupByRole } from 'frontend-kaleidos/config/permissions';
...
+
+  get impersonatedRole() {
+    return this.impersonation.role;
```
erikap: Is this 'alias' necessary?  I would like to avoid exposing properties of other services through this service.

app/helpers/user-really-may.js
```diff
@@ -0,0 +1,10 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+export default class UserReallyMay extends Helper {
```
erikap: Provide a named argument `checkImpersonator` on the `user-may` helper instead of introducing a separate helper?

app/components/m-header.hbs
```diff
@@ -58,6 +58,16 @@
       {{/if}}
     </Auk::Toolbar::Group>
     <Auk::Toolbar::Group @position="right" class="auk-u-mr-3">
+      {{#if
+           (and this.currentSession.impersonatedRole
+           (not-eq this.currentSession.role this.currentSession.impersonatedRole))
```
erikap: Provide a `get isImpersonated()` on the `currentSession` service containing this logic to 'hide' service internals

app/components/impersonate-user-modal.js
```diff
@@ -0,0 +1,30 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
+import { task } from 'ember-concurrency';
+
+export default class ImpersonateUserModalComponent extends Component {
```
erikap: Is this component used?

app/components/m-header.hbs
```diff
@@ -58,6 +58,16 @@
...
+      }}
+        <Auk::Toolbar::Item class="auk-u-mr-3">
```
erikap, line 62: Can the user stop impersonation without the need to logout?

app/services/impersonation.js
```diff
@@ -0,0 +1,82 @@
...
+          'Accept': 'application/vnd.api+json',
+          'Content-Type': 'application/vnd.api+json',
+        },
+        body: JSON.stringify({
+          data: {
+            type: 'sessions',
+            relationships: {
+              role: {
+                data: {
+                  type: 'accounts',
```
erikap: Must be `roles`?

app/services/impersonation.js
```diff
@@ -0,0 +1,82 @@
...
+                  id: role.id,
+                  attributes: {
+                    uri: role.uri,
+                  }
```
erikap: Not in line with JSON:API. Is this URI a requirement?

erikap (changes requested):

app/services/session.js
sergiofenoll: My understanding of the auth addon we use is that `handleInvalidation` only gets called if the underlying session is no longer valid (maybe it expires while the user is logged in). If the user manually logs out, we never go through `handleInvalidation`. I will double check though, because that was last week and perhaps I'm just mistaken.

app/components/m-header.hbs
sergiofenoll: No, though in essence choosing Admin again serves the same purpose. To make it more complete I'll add a proper button though

app/services/impersonation.js
sergiofenoll: Just added it to save on a query in the backend. Probably better not to, then we can check if the passed ID actually exists and if it doesn't the error is a bit more specific.

Aside: can you not pass data about relationships inline in JSON:API? Does relationship data have to go in the `includes` array?

erikap: Afaik, `includes` can only be used in responses, not in requests. You can only pass `type` and `id` for relationships inline in requests.

erikap:
Jenkins http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/334

erikap:
Failing tests not related to this PR. Going to merge this one.

PR #1556, opened 2022-11-30
KAS-3738 implement warning for uploading on agendaitem with approved …

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3738

## :warning: started from acceptance

This ticket was requested to be implemented before the final agenda's of 2022 so branched from acceptance.

- reworked the current warning/confirmation modal to be more generic (was only usable to open 1 specific upload modal)
- Added checking the callback of a  `document-card` `this.args.method` instead of also implementing the very specific logic there
- Changed the modal so it would appear:
    - when adding new documents on approved agendas
    - when finalizing adding new documents on approved agendas if you started uploading on design agenda but status changed
    - when adding new versions to existing documents on approved agendas
    - when finalizing adding new versions to existing documents on approved agendas if you started uploading on design agenda but status changed
    - does not appear on legacy or design agendas.  
    
how it works:
When trying any of the actions above we verify if uploading should be allowed.
if not, a manual confirmation is needed.
We show the confirmation modal.
- If not pressed within 60 seconds or cancelled, we just stop doing anything and close the confirmation modal.
     - if the action was finalizing doc upload, we delete the uploaded docs without extra confirmation (ok?)
- if confirm is pressed, we assume the user knows what they are doing and continue as normal. The window opens in less then 0.5 seconds with the for loop timeout strategy (we need the modal to stay up for awhile + our main code waits for that task to finish before its continues.

Comments:
- Not fully tested yet, enough to know that deletes happen after cancelling or not responding
- the for loop is quite basic, I've tried `waitForProperty` on the `@tracked` property, but doesn't seem to work properly. Github issues hinted at avoiding it. Room for improvement.
- Added more code to already complex `updateRelatedAgendaitemsAndSubcase`. May be better off separating some of that code in well named "private" methods
- Noticed now that I used `mayDocsBeEdited`, will have to rename that to avoid permission confusions.

Will continue with this on monday if it can wait. Feedback welcome

### DIFF
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -140,7 +140,11 @@ export default class DocumentsDocumentCardComponent extends Component {
   @action
   async openUploadModal() {
     if (this.args.onOpenUploadModal) {
-      await this.args.onOpenUploadModal();
+      // opening model depending on calculations made in parent
+      const shouldOpen = await this.args.onOpenUploadModal();
+      if (shouldOpen === false) { // explicit checking on boolean
+        return;
+      }
     }
     this.isOpenUploadModal = true;
   }
--- a/app/controllers/agenda/agendaitems/agendaitem/documents.js
+++ b/app/controllers/agenda/agendaitems/agendaitem/documents.js
@@ -56,8 +56,8 @@ export default class DocumentsAgendaitemsAgendaController extends Controller {
   @action
   async openPieceUploadModal() {
     await this.ensureFreshData.perform();
-    const mayDocsBeEdited = await this.checkIfDocumentsMayBeEdited();
-    if (mayDocsBeEdited) {
+    const canDocsBeEdited = await this.checkIfDocumentsCanBeEdited();
+    if (canDocsBeEdited) {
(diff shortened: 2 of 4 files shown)


erikap (commented):
Looks good. Just in doubt about the new timeout pattern that is introduced to close the modal automatically after 60s. Sounds more like a design decision we cannot make on our own. Or is it implemented like that in the usertest prototype as well?

ValenberghsSven:
> Looks good. Just in doubt about the new timeout pattern that is introduced to close the modal automatically after 60s. Sounds more like a design decision we cannot make on our own. Or is it implemented like that in the usertest prototype as well?

The 60 second thing was my own input since I am holding the pop-up modal open while the task is running.
Since we are waiting for this active task to end, I didn't want to keep it running forever.

ValenberghsSven:
Will update the `may` to `can` to remove any confusion with permissions.

Since we only have 1 window left to release this year I will go ahead and merge this ticket and make a minor version bump with just this ticket.
We can still decide after testing on ACC if this is acceptable to release on PROD.

PR #1557, opened 2022-12-01
cypress/KAS-3725-tests-for-users-organizations-management

MathiasZabot:


ValenberghsSven (approved):
4 tests need to be looked at later.
Scoping this PR for now.
If this spec passes (and no side-effects to other tests), it's ok to merge

PR #1560, opened 2022-12-05
cypress/KAS-3766-new-tests-remove-restore-case

MathiasZabot:


ValenberghsSven (approved):
Should be fine, merged in dev and run jenkins again for good measure

PR #1561, opened 2022-12-07
KAS-2742: Updated newsletter model

sergiofenoll:
Updates the NewsletterInfo model and renames it to NewsItem

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/345
- [x] https://github.com/kanselarij-vlaanderen/themis-export-service/pull/13 (Requires a bump in the export stack)
- [x] https://github.com/kanselarij-vlaanderen/newsletter-service/pull/49
- [x] https://github.com/kanselarij-vlaanderen/yggdrasil/pull/51
- [x] https://github.com/kanselarij-vlaanderen/agenda-approve-service/pull/32
- [x] ~~https://github.com/kanselarij-vlaanderen/nieuwsberichten-migration-service~~ (This can also be deleted, no need to keep it around)


app/models/agenda-item-treatment.js
```diff
@@ -5,8 +5,8 @@ export default class AgendaItemTreatment extends Model {
   @attr('datetime') created;
 
   @belongsTo('decision-activity') decisionActivity;
+  @belongsTo('news-item') newsItem;
```
sergiofenoll: Before this relationship had `serialize: false`. I'm not sure why. I changed it here because the inverse is now a `hasMany`. Ember doesn't serialize many-to-one relationships by default (you need to save on the one-to-many side to persist the relationship), so I made it serializable on this side.

We could also make the many-to-one side serializable explicitly by adding `serialize: true`, but I'm not sure why it was disabled on this side to begin with, and if something could break from serializing the many-to-one side.

ValenberghsSven: It was added as a concurrency fix in https://kanselarij.atlassian.net/browse/KAS-3453.
What happened was that user 1 was creating a newsletter while user 2 uploaded a report (then attached to treatment, now somewhere else) so user 2  saved `newsletter` as `null` on treatment, disconnecting the `newsletter` created by user 1.
You can try the steps to reproduce to check if it still occurs. instead of uploading a report, try also creating a conflicting newsletter or just saving the model

I haven't really looked at the other code just yet but I'm lost as to why a newsitem would have many treatments. We create 1 treatment, use that as a reference to connect to many agendaitem, 1 newsitem and 1 decisionActivity.
If the `has-many` of `treatments` is defined in the themis model I would argue that this is incorrect.
Since we currently implement one-to-one, when will the has-many ever come in to play?

erikap: Sorry for the confusion. Relation between treatment and newsitem is indeed 1-on-1 as @ValenberghsSven says. The story description hasn't been updated after the data migration where we split treatment and decision-activity. Before that migration the relation used to be has-many.

erikap: Removal of unnecessary `serialize: false` and `inverse: null` options will be tackled in future stories, so maybe best to keep them here.

ValenberghsSven:
:warning: this PR also needs a change to `agenda-approve-service`

ValenberghsSven:
looking at the code, there are multiple names now for this newsItem.
newsletterInfo (some components and routes still), newsItem (replaced many newsletterInfo), newsletter (more often in tests)
Perhaps we could take the opportunity to reach a consensus and use 1 term as much as possible

When referencing to the "nieuwsbericht": `newsItem`. (google suggests this as translation)
when referencing to the complete "nieuwsbrief" when sending a mailcampaign: `newsletter`   (uses `newsItems` to create it)
when referencing "kort-bestek", no idea, google translate has nothing, we mainly use `newsletter` in translations and routes). 


I will help remove any remaining `newsletterInfo` names and use`newsItem` instead (frontend tests and backend services)
Other renames could happen later if it complicates things

ValenberghsSven:
in the migrations service there is a large `DELETE` `INSERT` query with many `OPTIONAL` from the `SELECT`.
I thought that deleting with `OPTIONAL` could lead to issues?
In recent migrations we changed it to single predicate queries, sometimes in single migration files.
https://github.com/kanselarij-vlaanderen/nieuwsberichten-migration-service/blob/master/app.js#L70

sergiofenoll:
> in the migrations service there is a large `DELETE` `INSERT` query with many `OPTIONAL` from the `SELECT`.
> I thought that deleting with `OPTIONAL` could lead to issues?
> In recent migrations we changed it to single predicate queries.
> https://github.com/kanselarij-vlaanderen/nieuwsberichten-migration-service/blob/master/app.js#L70

I also thought the delets with optionals didn't work, but I tried running the migration service on an ACC copy and looking at the count of nieuwsberichten, and the migration seemed to work. Out of caution, I'm okay with rewriting it away from optionals, I'm just confused as to why it seemed to be okay here.

ValenberghsSven:
Let me check the frontend before merging. Will prioritize so we can merge this week.

ValenberghsSven:
> Let me check the frontend before merging. Will prioritize so we can merge this week.

Currently creating a fix for some newsletter-info issues on accept (pre refactor).
Will have to update this PR when the fix is stable and pulled into development

ValenberghsSven (approved):
Seems good for merge.
Last jenkins run had 1 subcase spec fail which I fixed. wait for 1 more jenkins run (run 13)

PR #1564, opened 2022-12-08
KAS-3782: Big attachments

sergiofenoll:
Adds warnings and filtering to the modals for creating publication requests (vertaling, drukproef, publicatie).
Files larger than 30 MB are outright rejected. If the sum of all attached files exceeds 30 MB, the request can't be saved. If it exceeds 10 MB, a warning will be shown about the files potentially not making it to the receiver.

Related PRs:
- [x] https://github.com/redpencilio/deliver-email-service/pull/10


erikap (approved):

app/components/publications/publication/proofs/proof-request-modal.js
```diff
@@ -32,6 +33,13 @@ export default class PublicationsPublicationProofsProofRequestModalComponent ext
     this.initValidators();
   }
 
+  get sumOfUploadedPiecesIsTooLarge() {
+    const sizeSum = this.uploadedPieces
+                        .map((piece) => piece.get('file.size'))
+                        .reduce((total, size) => total + size, 0);
+    return sizeSum > EMAIL_ATTACHMENT_MAX_SIZE;
+  }
+
```
ValenberghsSven, line 42: Used more than once, not sure if we want to turn a part of this into a util?

app/components/auk/file-uploader.js
```diff
@@ -10,8 +10,10 @@ import { enqueueTask } from 'ember-concurrency';
  * @argument {Boolean} multiple (optional) allow uploading multiple files
  * @argument {Boolean} reusable (optional, False by default) allow reusing the uploader to upload even more files, or add new files after previous uploads were deleted.
  * @argument {String} fileQueueName (optional) Name of the file queue to use.
- *   Setting this name will allow you to access this queue from the file-queue service outside of this component.
+ *  Setting this name will allow you to access this queue from the file-queue service outside of this component.
  * @argument {Function} onUpload: action fired for each file that gets uploaded. Passes a semantic.works File as an argument,
+ * @argument {Fuctnion} validateFile: function run to validate files before sending them to the backend.
```
ValenberghsSven: typo

app/components/publications/publication/publication-documents-list.js
```diff
@@ -0,0 +1,47 @@
...
+import { trackedFunction } from 'ember-resources/util/function';
+
+export default class PublicationsPublicationPublicationDocumentsList extends Component {
+  fileSizePillSkin = (size) => {
+    if (this.shouldWarnFileSize(size)) {
+      return 'warning';
+    }
+    return 'default';
+  }
+
```
ValenberghsSven, line 15: the methods in this component are getters without the explicit `get`?

sergiofenoll: It's actually a helper (because it needs an argument, so it wouldn't work as a getter). Starting in v4.5 (I think?) of Ember standalone function can be used helpers, you don't need to write them in a separate file, which is especially useful when they're one-offs.

In the meantime there's a [polyfill](https://github.com/ember-polyfills/ember-functions-as-helper-polyfill) for the feature which we have as a [dependency](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blame/development/package.json#L102).

ValenberghsSven: Right my bad, you pass in arguments. missed that.
The `trackedFunction` is new. helpers in a component we don't use that often?
is there like a `@helper` to clarify that?

sergiofenoll: `trackedFunction` comes from https://github.com/NullVoxPopuli/ember-resources

Basically, it works like getters and old computed properties in that it's rerun whenever a tracked value changes. E.g. if the args or a `@tracked` member change, the function runs again. It also lets you write async functions. In this specific case I needed it because the `Publications::Publication::PublicationDocumentsList` component is used in two places: the request creation modals & the email cards where we display the mail sent to BS.

The problem is that in the latter, when a user deletes the email, we destroy the pieces while still rendering the document list. With a regular getter (I first wrote `sumOfPiecesShouldWarn/Block` as getters) we just use the PromiseProxy we get from ember for the arguments and iterate over those. But that breaks because as the pieces get destroyed, we're still trying to access them and Ember throws exceptions about that.

With the trackedFunction we can write an async function and explicitly await the pieces argument, that way if the pieces are getting destroyed we get the settled array of pieces instead (or an empty array) instead of iterating over pieces that are getting deleted.

As for a `@helper` decorator, as far as I know there isn't one. The idea is that any function/method can be a helper (so helpers defined in `app/helpers` are also just going to export a plain function, not functions wrapped in a `helper`). The tricky thing about them is that binding can work in an unexpected way. If you don't use an arrow function to define a method, `this` won't be bound to the controller/component in the helper. You can read about that here if you're interested: https://blog.emberjs.com/plain-old-functions-as-helpers/

ValenberghsSven (approved):
fixed the typo.
Tested translation request / proof request / publication request.
Frontend part seems to be working fine.
Actually sending to be tested on ACC

PR #1565, opened 2022-12-08
KAS-3702: Use of index routes

sergiofenoll:
Replaces `overview` routes with the built-in `index` pages that Ember supports. Also removes any unnecessary definitions of `index` and `loading` routes from the router.


erikap:
Cypress tests need updates because of the changed route paths

ValenberghsSven:
Ik zal dit even oppakken?

ValenberghsSven:
> unnecessary definitions of index and loading routes from the router.

Define "unnecessary"
It seems that certain '/overzicht' routes now have changed. that is part of the reason tests are failing.
like a subcase overview
before:
dossiers/6374F30CD9A98BD0A2288557/deeldossiers/6374F312D9A98BD0A2288558/overzicht
after:
dossiers/6374F30CD9A98BD0A2288557/deeldossiers/6374F312D9A98BD0A2288558

So I guess in cases like this we **do** need to have an explicit index route?
Same for "instellingen/overzicht", which is now just "instellingen"

It means if people had links to those addresses saved locally they will no longer work...

erikap:
> So I guess in cases like this we **do** need to have an explicit index route? Same for "instellingen/overzicht", which is now just "instellingen"
> 
> It means if people had links to those addresses saved locally they will no longer work...

If we want to keep the `/overzicht` path, an explicit index route is indeed required. But I don't see the added value of it tbh. 
Using implicit index routes will indeed 'break' existing bookmarks for users, but I don't think that will be a major issue. 
By sticking to the Ember defaults all index routes in the app are at least consistent.

ValenberghsSven:
> > So I guess in cases like this we **do** need to have an explicit index route? Same for "instellingen/overzicht", which is now just "instellingen"
> > It means if people had links to those addresses saved locally they will no longer work...
> 
> If we want to keep the `/overzicht` path, an explicit index route is indeed required. But I don't see the added value of it tbh. Using implicit index routes will indeed 'break' existing bookmarks for users, but I don't think that will be a major issue. By sticking to the Ember defaults all index routes in the app are at least consistent.

Sounds good, will adapt the tests accordingly

erikap (approved):
Failing test not related to this PR

PR #1566, opened 2022-12-08
KAS-3703: Octane refactor ModelWithModifier and add some comments

sergiofenoll:
Updates the model-with-modifier model to use Octane syntax and adds some comments to the methods it has.

As to finding a more elegant solution, I'm not entirely sure there is one given the constraints of the app. Concurrent editing requires some form of either locking, or conflict merging. In either case we need extraneous data about the record (what was the original state, how does it compare to the current state, etc.) which AFAIK we can't get in Ember directly without going through the hoops we already do.

The `modified` and `modifiedBy` data aren't strictly necessary, we could just compare the records' fields directly to know if changes occurred, but I imagine they are useful for users to track down who changed what.

May be of interest: http://jim-mcbeath.blogspot.com/2009/02/concurrent-editing-without-locking.html

### DIFF
--- a/app/models/model-with-modifier.js
+++ b/app/models/model-with-modifier.js
@@ -9,7 +9,7 @@ import ModifiedOldDataError from 'frontend-kaleidos/errors/modified-old-data-err
  * preventing the user from persisting edits of the record data if another user
  * persisted their edits before us.
  */
-export default class ModelWithModified extends Model {
+export default class ModelWithModifier extends Model {
   @service currentSession;
   @service intl;
   @service store;
--- a/app/models/model-with-modifier.js
+++ b/app/models/model-with-modifier.js
@@ -1,25 +1,39 @@
+import Model, { belongsTo, attr } from '@ember-data/model';
 import { inject as service } from '@ember/service';
 import moment from 'moment';
-import Model, { belongsTo, attr } from '@ember-data/model';
 import fetch from 'fetch';
-import ModifiedOldDataError from '../errors/modified-old-data-error';
+import ModifiedOldDataError from 'frontend-kaleidos/errors/modified-old-data-error';
 
-// TODO: octane-refactor
-/* eslint-disable ember/no-get */
-// eslint-disable-next-line ember/no-classic-classes
-export default Model.extend({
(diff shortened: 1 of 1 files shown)


ValenberghsSven:
This one needs some manual testing as cypress can't simulate concurrent edits easily

ValenberghsSven (approved):
Tested it locally, concurrency seems to be working.

Trying to upload docs (new or BIS) on agendaitem when someone else edited agendaitem > error, upload modal doesn't open untill refresh
Trying to upload docs (with upload modal already open)...  (same scenario)  > upload can't proceed, brought back to upload modal and need to refresh  (cancel should still happen or pieces keep existing, just refresh does not do deletes of piece)
Trying to edit titles...  (same scenario) > save does not work, need to refresh.
Editing agendaitem titles and then trying to concurrently edit subcase titles > save does not work, need to refresh.
Trying to edit the text of newsitem > save does not work, need to refresh.

**out of scope** (this was already happening before I reckon)
not everything seems to trigger the error.
I could overwrite the list of `governmentAreas` by concurrent edits on agendaitem and subcase.
I could edits newsitems themes and thereby overwrite the other users text.
Both these issues > (model is clean, relations are dirty and all stale attributes are saved again even though none have been changed.)

I could create a newsitem with both users, last save wins and first newsitem is disconnected.
The second CreateRecord should not be allowed.

erikap:
> The `modified` and `modifiedBy` data aren't strictly necessary, we could just compare the records' fields directly to know if changes occurred, but I imagine they are useful for users to track down who changed what.

Comparing fields is getting complex when we need to compare async relations of a record. I assume that's why this approach with `modified` and `modifiedBy` has been implemented in the past.

> May be of interest: http://jim-mcbeath.blogspot.com/2009/02/concurrent-editing-without-locking.html

Thanks :pray:  Will have a look at that one. 

For me, this PR can be merged since the main purpose of this story was to convert to Octane syntax to unblock the Ember v4 upgrade. Nevertheless, I think we need to work out a better/thoughtful/consistent way to address concurrency in the entire application in the future. That blog post may be an inspiration :+1:

PR #1567, opened 2022-12-08
Feature/kas 3468 refactor agendaitem model

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3468

Refactor agendaitem model:
`sortedMandatees` > sort locally where possible
`sortedPieces` > only used in print view, loaded data properly and sort like we do in other agenda routes (cached calls)
`formallyOkToShow` > looks fine as a getter.
`nota` > made new service for `notaOrVisie`, was only used by some components that were no longer used after recent cleanup.
`notaOrVisie` > made new service, return latest version of piece instead of document-container. Kept in local tracked.


erikap (approved):
Changes look :+1: 
Not sure whether failing tests are caused by this PR.

ValenberghsSven:
> Changes look +1 Not sure whether failing tests are caused by this PR.

They are.
It seems that this branch has the issues that were being caused by `ember-in-viewport` not working properly.
The sidebar icons and the overview "formally ok" are not showing when they should.
Not sure why this is happening here

Could also be the formally ok getter, will test more locally

app/models/agendaitem.js
```diff
@@ -1,128 +1,49 @@
...
+    return this.constructor.modelName;
+  }
 
-  // TODO this computed property is used in:
-  // - Agenda::AgendaOverview::AgendaOverviewItem
-  // - Agenda::AgendaDetail::SidebarItem
-  // - Agenda::Agendaheader::AgendaActionPopupAgendaitems
-  // - Agenda::Agendaitem::AgendaitemCasePanel::AgendaitemCasePanelView
-  // Refactor these usages and remove this computed property
-  formallyOkToShow: computed('formallyOk', function() {
```
ValenberghsSven, line 123: Guessing this is the problem, `formallyOkToShow` computed vs `formallyOk` getter

ValenberghsSven:
That fixed it, merging

PR #1570, opened 2022-12-09
KAS-3690 add plausible package and config and setup in application route

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3690

- implementing plausible (done I think? custom events and goals out of scope)
- removing matomo (not done, don't we want to run both analytics in parallel for accuracy?)
- testing (I don't have a plausible login that I know of, how does one connect with the api host)

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -14,14 +14,12 @@ export default class ApplicationRoute extends Route {
   @service currentSession;
   @service fileService;
   @service router;
-  @service metrics;
   @service userAgent;
   @service plausible;
 
   constructor() {
     super(...arguments);
     this.setupPlausible();
-    this.setupTracking();
   }
 
   async beforeModel() {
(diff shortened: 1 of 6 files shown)


erikap (commented):
Plausible setup looks good :+1: 

Env vars to be used (for production) are
```
      EMBER_ANALYTICS_APP_DOMAIN: "kaleidos.vlaanderen.be"
      EMBER_ANALYTICS_API_HOST: "https://plausible.kanselarij.org"
```

Business has decided to go for Plausible and not Matomo. No need to run both in parallel. The Matomo config/setup can be removed from the app.

ValenberghsSven:
> Business has decided to go for Plausible and not Matomo. No need to run both in parallel. The Matomo config/setup can be removed from the app

That is everything regarding "matomo" or "metrics" so ember-metrics, ember-metrics-matomo-adapter and all `@service metrics`?
Will do tomorrow.

ValenberghsSven:
1 question regarding the setup. 
I put the plausible setup in the `constructor` because that was were the matomo setup was.
I saw other apps put it in the `beforeModel`.
I guess that as long as it gets called, it should be fine

ValenberghsSven:
only failing test is a flaky one.

erikap (approved):

erikap:
> 1 question regarding the setup. I put the plausible setup in the `constructor` because that was were the matomo setup was. I saw other apps put it in the `beforeModel`. I guess that as long as it gets called, it should be fine

Constructor is OK.  To do it the Ember-way it should be done in [an application instance initializer](https://guides.emberjs.com/release/applications/initializers/#toc_application-instance-initializers) (same holds for `moment` and `intl` setup) but let's consider that out of scope of this story.

PR #1573, opened 2022-12-13
Use application instance initializer to setup application

erikap:
Inspired by @ValenberghsSven's question on #1570 moving the application setup from the application route to application instance initializers. 

2 initializers have been added:
- `plausible`: enabling/configuring Plausible
- `locale`: setting default locale for `date-fns`, `moment` and `intl`


ValenberghsSven (approved):
looks good :+1:

PR #1575, opened 2022-12-15
Feature/kas 3803 basic plausible metrics

sergiofenoll:
- Extends the Plausible service to facilitate tracking user role types as well as "first action of session" tracking
- Adds a `plausible-click` modifier that sends a custom event whenever the attached DOM element is clicked
- Adds "action tracking" for opening the Decisions and Kort bestek views in the agendaitem

Note: I also wrote code to track opening the Nota of an agendaitem, snippet:

```js
{{(if
  this.documentContainer.isNota
  (modifier
    "plausible-click" "Actie" category="Nota van agendapunt opvragen"
  )
)}}
```

`ember-template-lint` complains about `{{(`, but that should be valid handlebars syntax (at least, it compiles and I can see the result in my browser). The syntax is also how you're supposed to have conditional modifiers, see [this blog post](https://v5.chriskrycho.com/journal/conditional-modifiers-and-helpers-in-emberjs/#the-basics).

I suspect it's because we're using an outdated version of `ember-template-lint`, but I didn't want to start upgrading packages in a production "hotfix".

In any case, once we've upgraded Ember, we should be able to add that back in.

To add the action tracking in Plausible, the following goals need to be added:
- Actie
- Eerste actie
- Aanmelding (per rol)
- Gebruikssessie (per rol)
- Pageview (per rol)

Before implementing all the tracking actions defined in the spreadsheet, I would recommend going over what we want to track and how. The goal tracking in Plausible's UI may not be as flexible as desired (e.g., it doesn't seem to be possible to select multiple goals as a filter), so we'll want to check if we can get what we want from it. Otherwise, we should explore if [Plausible's Stats API](https://plausible.io/docs/stats-api) can get us the data we want, or if we need to play around with which events we send to Plausible.

Plausible screensots:

This contains most of the basic session-related data we want, like number of unique visitors, number of page views, session duration:
image

Country of origin, device, OS, browser, ... are also included by default:
image

Action tracking is done using Plausible's goals:
image
image
image

It is possible to select a goal and filter on it. E.g., select `Admin` in `Pageviews (per rol)`, then we can see the pages which were viewed by admin users:
image

Custom Jenkins run:
- http://kal-kastaar.s.redpencil.io:8080/job/kaleidos/job/custom_frontend_backend_combination/336/

### DIFF
--- a/app/modifiers/plausible-click.js
+++ b/app/modifiers/plausible-click.js
@@ -1,0 +1,26 @@
+import Modifier from 'ember-modifier';
+import { inject as service } from '@ember/service';
+
+export default class PlausibleClickModifier extends Modifier {
+  @service plausible;
+
+  get eventName() {
+    return this.args.positional[0];
+  }
+
+  get props() {
+    return this.args.named;
+  }
+
+  onClick = () => {
+    this.plausible.trackEvent(this.eventName, this.props);
+  }
+
+  didInstall() {
+    this.element.addEventListener('click', this.onClick);
+  }
+
+  willRemove() {
+    this.element.removeEventListener('click', this.onClick);
+  }
+}
--- a/app/services/plausible.js
+++ b/app/services/plausible.js
@@ -1,0 +1,37 @@
+import PlausibleService from 'ember-plausible/services/plausible';
+import { inject as service } from '@ember/service';
+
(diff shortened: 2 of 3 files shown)


erikap (approved):
Elegant solution :+1:

PR #1583, opened 2022-12-17
KAS-3754 New user role kanselarij

erikap:
Add new user role kanselarij with same permissions as admin, except for the admin features.

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/352


sergiofenoll (approved):
Looks good, failing tests are related to addition of new user, some tests check for number of results in user overview, should be a tiny change.

We may want to change the blocked users from testdata.zip,which were added when the blocking feature was added for testing purposes. Their role is Admin, so now they appear at the top of the mock login page. Maybe we want to set them to Vlaams Parlement role instead, so they appear at the bottom?
image

cypress/integration/unit/user-organisation-management.spec.js
```diff
@@ -144,7 +144,7 @@ context('testing user and organisation management', () => {
     });
 
     it('check filter by group', () => {
-      checkRoleFilterSingle('Admin', 8);
+      checkRoleFilterSingle('Admin', 9);
       checkRoleFilterSingle('Secretarie', 11);
```
ValenberghsSven: I think all of these should be "van 12", since we are filtering out one by one
I'll check locally to avoid multiple jenkins runs

app/config/permissions.js
```diff
@@ -67,6 +68,30 @@ const groups = [
...
+      'manage-meetings',
+      'manage-documents',
+      'manage-document-access-levels',
+      'manage-publication-flows',
+      'search-publication-flows',
+      'manage-news-items',
+      'manage-decision-publications',
+      'manage-document-publications',
+      'manage-themis-publications',
+      'view-document-version-info',
```
ValenberghsSven, line 90: manage settings?

ValenberghsSven (approved):
looks fine, permissions can be finetuned later if needed.

PR #1586, opened 2022-12-19
Feature/kas 3771 dependencies

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3771

Draft.
Checking when or if test are failing because of version bumps.
1st run:
bumped `ember-promise-helpers` to most recent
removed `@lblod/ember-rdfa-editor-plugin-system-dispatcher`, archived, should be fine with newer version of editor
removed `ember-maybe-import-regenerator` was pointless since it does nothing if we have `includePolyfill: true`
replaced `ember-cli-sanitize-html` with `sanitize-html` and custom octane component (the package only had 1 component)

note: it is possible jenkins does not play nice with small bumps in npm and gets previous cached npm cached. Unsure of how all that works precisely...


ValenberghsSven:
our own fork https://github.com/erikap/ember-attacher
Can be removed if no other repo uses this.
Removed dependency in this PR

ValenberghsSven:
Deployed this feature branch on ACC to test acmidm and seems to work.
No more babel warnings on startup.

erikap (approved):

ValenberghsSven:
some tests are failing, trying to fix them.
It's a bit weird that I had a green run at one point, then I changed the jquery flag and the acmidm package.
Unsure what is happening

ValenberghsSven:
made small changes to jenkins to only run a few specs for faster results.
Reverted the jquery integration flag back to true and it seems to work better.
Is Jquery being used somewhere without explicit injection?

Going to remove the flag again and add the jquery package to see if it that makes a difference.

ValenberghsSven:
update, looking at the logs and screenshots, it seems that mandatees are not always loading.
Looking deeper, it seems that the migrations service is building the DB from scratch (so all migrations)
Unsure why the zip is not being used.

ValenberghsSven:
finally a green PR without the jquery thing, merging this one.

PR #1587, opened 2022-12-20
feat: add explicit inverses and asyncs to models

sergiofenoll:
Adds explicit inverses and async to all ED relations.

Where possible, I've tried to add inverses even when we nulled them. For relations that shouldn't be serialized (keeping concurrency in mind), I have moved that to the serializer (using `serializeBelongsTo` & `serializeHasMany` to force certain keys to be skip serialization, as well as `shouldSerializeHasMany` to force certain many-to-one relations to be serialized).

This still requires thorough testing to ensure all serialization assumptions still hold w.r.t. concurrency, as well as checking that the added inverses don't break anything (I'm not sure why they'd break anything in the first place, but some comments seem to indicate that the inverses are absent to prevent breakage?). WIP until that is done, but I already want to see how the tests behave.

Where possible inverses are added. I.e., if both sides of a relation are already defined in the resources config, the relation is properly linked in the frontend models too. E.g., the link between piece and documentContainer is now made in the frontend, this required some extra code change in the document card component.

Below follows a list of relations with `inverse: null` so we can either change them, or figure out why they should remain nulled:
- [x] All relations to `concept`s and concept-like resources (e.g. `case-type`). Listing all inverses on the concept side would be cumbersome and would probably offer very little benefit, as we seldom have a concept model from which we want to get related models (instead, we would query using the concept URI for filtering)
- [x] All relations to `file`s. Adding inverses on the file side makes working with them hard. Since file deletion is handled by the file-service, removal of the extra relations requires manual bookkeeping in the frontend or addition of a reactive service to clean up triples not handled by the service.
- [x] `piece:document-container`: The two-way relation is defined in the backend, we just weren't defining it in the frontend. I'm not entirely sure *why* we did that, but given some small tweaks to the code it seems to work.
- [ ] Modified and generated relations to `user`: we could add the inverses to the user resource for e.g. `modified-by` on agendaitem/subcase/etc. and `generated-by` on `publications-metrics-export-jobs`, but I'm not sure how useful that would be in practice.
- [x] `submission-activity:piece`: On the piece side, this relation is not defined in the backend. For that reason, we don't serialize the relation when sending it in PATCH and POST requests to the backend, and we also don't link the relationship in the model. It's unclear to me why we do the latter, and why we don't just define the relationship in the backend and just serialize it.
- [x] `meeting:agenda`: Meeting has a one-to-none relationship to agenda that points to the final agenda of the meeting. Agenda does not store a backlink to the meeting. Unsure why.
- [x] `submission-activity:mandatee`: Mandatee has a link to submissionActivity, but the inverse isn't true. We can probably just define it in the backend and add it to the frontend, I don't immediately see why it's absent.
- [x] `agendaitem:piece`: Agendaitem has a link to piece via predicate `ext:bevatReedsBezorgdAgendapuntDocumentversie` called `linked-pieces`. The inverse isn't defined in the backend. I believe it's an analogous relation to `ext:bevatReedsBezorgdeDocumentversie` between subcase and piece. Unclear why it's undefined.
- [x] `cancellation-activity:piece`: This relation is simply not defined in the backend, on either side. But it appears we also don't use this model in the frontend at all, so it might just be a bug that we never noticed because we never use the codepath.


erikap:
@sergiofenoll  About `meeting:agenda` relation, will probably be cleared out in https://kanselarij.atlassian.net/browse/KAS-3733 where that relation will be used more actively. Can you try to define the inverse here and check whether it has impact on tests?

erikap:
About undefined inverse relations on `piece`, if I remember correctly, this was done to limit cache invalidations on `piece`-related requests in mu-cache since cache clearing is very naive/agressive at the moment. Any change on any piece will invalidate all piece-related requests in the cache. 
Nevertheless, I am inclined to try to make the inverse relationships explicit, both in frontend and backend and see what the impact is.

erikap:
`cancellation-activity:piece` can be safely removed afaik. They were introduced to support cancellation flows for OVRB, but this feature is deprioritized. Relation can be easily readded in the future once the feature will be implemented.

erikap (approved):

app/models/agenda.js
```diff
@@ -13,25 +13,18 @@ export default class Agenda extends Model.extend(LoadableModel) {
   @attr('datetime') created;
   @attr('datetime') modified;
 
-  @belongsTo('meeting') createdFor;
+  @belongsTo('meeting', { inverse: 'agenda', async: true }) meeting;
+  @belongsTo('meeting', { inverse: 'agendas', async: true }) createdFor;
   @belongsTo('agendastatus', { inverse: null }) status;
```
ValenberghsSven, line 18: `agendastatus` not changed? we have the relation in the backend resources and a frontend model so we could set the inverse properly.

app/serializers/publication-flow.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  shouldSerializeHasMany(snapshot, key, relationshipType) {
+    if (key === 'mandatees') {
+      return true;
+    } else {
+      return super.shouldSerializeHasMany(snapshot, key, relationshipType);
+    }
+  }
+}
```
ValenberghsSven: Don't fully get this.
I think all hasMany's that get loaded get saved unless you skip them.
Why would it be different in this case but not other hasMany's that do get serialized?

ValenberghsSven (commented):
I think it looks quite ok. Was expecting something different with the inverses and serialize false.
the serializers are a better place to define what to not serialize.

app/models/agenda.js
erikap: I think `agendastatus` will be changed to `concept` in [a story](https://kanselarij.atlassian.net/browse/KAS-3775) @StuyckRuben is working on, so I would leave it as it is here.

app/serializers/publication-flow.js
erikap: Not 100% sure, but I think an `hasMany` that is loaded only gets serialized if `inverse: null` is set on that relation. As far as I understood, a 1-to-many relation with inverses propertly defined at both sides, is always serialized at the `belongsTo`-side, not at the `hasMany`-side.

ValenberghsSven: Is this a many-to-many specific thing?
Shouldn't this then also be true for `mandatees` on `agendaitems` and `subcases`?

sergiofenoll: I defined it explicitly like this because the old code also explicitly defined the serialization of this relationship: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1587/files/2565ed9f40ee286ff0503774283c0f4c7180ce7c#diff-b51f28464e3d15b4cbec6fc95c877c9d917aff07d4e8c6f2e72c0544a31b17c9L37-L39

*But* I didn't check to see if it was actually necessary, I was just working on auto pilot for most of these mechanical changes. Thinking about it, this is probably completely unnecessary. I know for `one-to-many` you have to explicitly serialize on the `hasMany` side, but I don't think it's necessary for `many-to-many` relations.

erikap: According to [the docs](https://api.emberjs.com/ember-data/4.9/classes/JSONAPISerializer/methods/shouldSerializeHasMany?anchor=shouldSerializeHasMany), it's indeed not necessary for `many-to-many`.

> By default only many-to-many and many-to-none relationships are serialized.

app/models/agenda.js
ValenberghsSven: Ah I see.

erikap:
Can be merged @ValenberghsSven @sergiofenoll ?

app/serializers/publication-flow.js
sergiofenoll: I'm removing this `shouldSerializeHasMany` here since it's not necessary, afterwards the PR should be okay change-wise (at least AFAIK)

ValenberghsSven:
Perhaps we should merge.
Sergio fixed some things in this that are also happening on DEV, but I don't see any failing test that would be caused by this PR.

sergiofenoll:
> Perhaps we should merge. Sergio fixed some things in this that are also happening on DEV, but I don't see any failing test that would be caused by this PR.

Alright, merging this and the backend branch :)

PR #1595, opened 2023-01-11
KAS-3829 Fully reload application after impersonation

erikap:
Data in the ember-data store also needs to be reset on impersonation, because it may contain data the alias-profile is not allowed to see (yet).

(@tdn You were right that it was an easy story :wink: )


ValenberghsSven (approved):
The issue with document rights after switching is gone now.

PR #1596, opened 2023-01-12
Kas 3733 remove isfinal boolean

StuyckRuben:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/355
- [ ] https://github.com/kanselarij-vlaanderen/agenda-approve-service/pull/33


app/controllers/agenda/agendaitems/agendaitem/index.js
```diff
@@ -24,6 +24,7 @@ export default class IndexAgendaitemAgendaitemsAgendaController extends Controll
   @tracked decisionActivity;
 
   @tracked isEditingAgendaItemTitles = false;
+  //TODO: add isClosed
```
tdn: Any reason why this TODO is here? Don't see any use for `isFinal`/`isClosed` in this controller. I'll remove this

app/components/subcase/subcase-titles-panel/subcase-titles-panel-view.js
```diff
@@ -14,30 +14,35 @@ export default class SubcaseTitlesPanelView extends Component {
...
+    }
     if (this.canShowDecisionStatus) {
       this.approved = yield this.subcaseIsApproved.isApproved(this.args.subcase);
     } else {
       this.approved = false;
     }
   }
 
-  get pillSkin(){
+  get pillSkin() {
```
tdn, line 45: I've just condensed this getter a bit to make it a bit cleaner

app/controllers/agendas.js
```diff
@@ -58,7 +58,6 @@ export default class AgendasController extends Controller {
     const plannedStart = new Date(now.getFullYear(), now.getMonth(), now.getDate(), 10, 0, 0);
     this.newMeeting = this.store.createRecord('meeting', {
       plannedStart,
```
tdn: Removing trailing comma here (and in some other places)

tdn (commented):
Looks good, works locally. I'm just committing a few nitpicking changes here and there and then this one is fine to merge for me

tdn (approved):

app/controllers/agenda/agendaitems/agendaitem.js
```diff
@@ -19,11 +19,11 @@ export default class AgendaAgendaitemsAgendaitemController extends Controller {
 
   get canShowDecisionTab() {
     return this.currentSession.may('manage-decisions')
-      || (this.meeting.isFinal && this.hasDecision);
+      || (this.meeting.agenda && this.hasDecision);
```
erikap: Since `this.meeting.agenda` returns a promise, `this.meeting.agenda` will always be thruthy, even if the promise resolves to `null`. You should check whether it resolves to an actual agenda, for example by checking the existance of `agenda.id`. In practice it might be that the promise immediately resolves if the data has already been loaded, but that's not something you may rely on.

Since `this.meeting.agenda` is a promise, you will need to use `get()` to retrieve the `id` property, like `this.meeting.agenda.get('id')`.

To stay in line with the remainder of this file it's probably best to define a `get isFinalAgenda() { ... }` similar to `get hasDecision()` and `get hasNewsItem()`.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -57,7 +60,7 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
   get canPublishDecisions() {
     return (
       this.currentSession.may('manage-decision-publications') &&
-        this.args.meeting.isFinal &&
+        this.isClosed &&
```
erikap: Similar to what you do in `app/controllers/agenda/agendaitems/agendaitem.js` in this PR, you can define a `get isFinalMeeting() { ... }` and use that instead of introducing the tracked property `isClosed`. 

Advantage of this approach is that the template will automatically be updated if the relation `this.meeting.agenda` is set somewhere in the app, while this will not be the case with the tracked property `isClosed`.

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -50,13 +50,14 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
   @tracked designAgenda = null;
   @tracked lastApprovedAgenda = null;
 
+  @tracked isClosed = false;
```
erikap: Similar to previous comments, you can define a `get isFinalMeeting() { ... }` and use that instead of introducing the tracked property `isClosed`.

app/components/subcase/subcase-description-panel/subcase-description-view.js
```diff
@@ -21,6 +21,8 @@ export default class SubcaseDescriptionView extends Component {
   @tracked approved = null;
   @tracked modelsOfMeetings = [];
 
+  @tracked isClosed = false;
```
erikap: Similar to previous comments, you can define a `get isFinalMeeting() { ... }` and use that instead of introducing the tracked property `isClosed`.

app/components/subcase/subcase-titles-panel/subcase-titles-panel-view.js
```diff
@@ -14,33 +14,35 @@ export default class SubcaseTitlesPanelView extends Component {
   @service currentSession;
 
   @tracked approved; // or acknowledged
+  @tracked isClosed;
```
erikap: Similar to previous comments, you can define a `get isFinalMeeting() { ... }` and use that instead of introducing the tracked property `isClosed`.

app/services/subcase-is-approved.js
```diff
@@ -13,7 +13,7 @@ export default class SubcaseIsApprovedService extends Service {
       sort: '-planned-start',
     });
 
-    if (latestMeeting?.isFinal) {
+    if (latestMeeting?.agenda) {
```
erikap: Similar comment as on `app/controllers/agenda/agendaitems/agendaitem.js`. `latestMeeting?.agenda` returns a promise. You need to explicitely `await` the relation and check on the resolved value.

erikap (changes requested):

tdn:
@erikap , in the ticket, I forgot to think about the search routes and sort options that use `created-for.is-final`, causing all kinds of havoc. Fixed that in my latest commit. Also improved the sorting on agendas while I was at it. This was already working poorly on dev.

For the sorting on 'status', which we don't have anymore, I had to use a workaround since we can't really sort on the existence of a relation or not. Sorted on `created-for.agenda.status.label` instead, which has the desired effect since open meetings won't have that relation set.

app/components/agenda/agenda-header/agenda-version-actions.hbs
```diff
@@ -24,7 +24,7 @@
     </AuButton>
     <AuHr />
   {{/if}}
-  {{#if @meeting.isFinal}}
+  {{#if this.isClosed}}
```
erikap: Must be replaced with `isFinalMeeting`

app/components/agenda/agenda-header/agenda-version-actions.hbs
```diff
@@ -70,7 +70,7 @@
       {{t "agenda-reopen-previous-version"}}
     </AuButton>
   {{/if}}
-  {{#if (and (not this.designAgenda) (not @meeting.isFinal))}}
+  {{#if (and (not this.designAgenda) (not this.isClosed))}}
```
erikap: Must be replaced with `isFinalMeeting`

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -44,7 +45,7 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
   constructor() {
     super(...arguments);
     this.loadPublicationActivities.perform();
-  }
+  }
```
erikap: Remove trailing whitespaces

app/components/subcase/subcase-titles-panel/subcase-titles-panel-view.js
```diff
@@ -17,30 +18,31 @@ export default class SubcaseTitlesPanelView extends Component {
...
   constructor() {
     super(...arguments);
-    this.loadApproved.perform();
+    this.loadAgendaData.perform();
   }
 
   get canShowDecisionStatus() {
     return (
-      this.args.meeting?.isFinal &&
+      this.isClosed &&
```
erikap: Must be replaced with `isFinalMeeting`

erikap (changes requested):

ValenberghsSven:
Guessing tests are failing because of the `agenda-approve` service.
You could add a temporary feature branch bump in the app-kaleidos `docker-compose.yml` to see if it works

ValenberghsSven:
some agenda-header test are failing, even with the correct agenda-approve image.

ValenberghsSven:
Ik zal eens kijken waarom die 3 testen falen hier.

ValenberghsSven:
think I found it:
-when we have a meeting with a final agenda, you could delete the latest version (rarely used in PROD)
In that case we need to move the final agenda to the latest existing one.
The meeting is still final, but a different agenda is the last one. I'll fix this.

ValenberghsSven:
Pushed changes to agenda-approve service and manually started jenkins run.

ValenberghsSven:
the first failed test of `agenda-header.spec` is working flawlessly locally.
image

I made some of those agenda-header.spec tests to be loopable and ran a loop of 5 on the failing test.

PR #1599, opened 2023-01-13
replaced send() with router.refresh()

StuyckRuben:


app/controllers/agenda.js
```diff
@@ -30,7 +30,7 @@ export default class AgendaController extends Controller {
 
   @action
   refresh() {
-    this.send('reloadAgendaModel');
+    this.router.refresh();
   }
```
tdn: @StuyckRuben I would be inclined to remove this action altogether and use `this.router.refresh()` directly. Would declutter the code. Do you agree @erikap ?

tdn (changes requested):
Would remove the `refresh` actions everywhere where they're only doing this one thing (`this.router.refresh()`), and use `this.router.refresh()` directly in the code.

app/controllers/agenda.js
erikap: Yes, occurs in a couple of others files as well. Was also my single remark on this PR :+1:

app/controllers/agenda/agendaitems/agendaitem/documents.js
```diff
@@ -338,9 +338,4 @@ export default class DocumentsAgendaitemsAgendaController extends Controller {
   async unmarkForSignature(piece) {
     await this.signatureService.unmarkDocumentForSignature(piece);
   }
-
-  @action
-  refresh() {
-    this.send('reloadModel');
-  }
```
erikap: In some places the `refresh` action has been removed while it's still used in the template to pass as an argument to a component. Can you double check all removals of the `refresh` actions?

erikap (changes requested):

app/controllers/agenda/documents.js
```diff
@@ -126,12 +127,7 @@ export default class AgendaDocumentsController extends Controller {
...
-    this.refresh();
+    this.router.refresh();
     this.isOpenBatchDetailsModal = false;
   }
-
-  @action
-  refresh() {
-    this.send('reloadModel');
-  }
 }
```
ValenberghsSven, line 138: What @erikap said.
The `refresh` action in the controller has been removed but it still used in the template.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/KAS-3809-replace-send-by-router-refresh/app/templates/agenda/documents.hbs#L34
Perhaps it made more sense to keep the action with `this.router.refresh()` and use `this.refresh()` in the 2 methods in this controller that need to trigger a refresh.

There are more cases.
Perhaps best to check all templates of the controllers where the action was removed.

ValenberghsSven (changes requested):

ValenberghsSven:
Taking over to keep this moving

app/controllers/agenda.js
ValenberghsSven: using `{{this.router.refresh}}` in template does not seem to work (console errors), so in those cases we will need a controller action.

app/controllers/agenda/agendaitems/agendaitem/documents.js
ValenberghsSven: fixing this one

ValenberghsSven: also `router` service was missing here.

app/controllers/agenda/agendaitems/agendaitem/index.js
```diff
@@ -64,7 +64,7 @@ export default class IndexAgendaitemAgendaitemsAgendaController extends Controll
     await this.navigateToNeighbouringItem(this.model);
     // reload the agenda route, detail tab should no longer show if we deleted the last and only agendaitem
     // Also, if we deleted the first agendaitem, we should also reload the main route to reload <Agenda::agendaTabs>
-    return this.agendaController.send('reloadAgendaModel');
```
ValenberghsSven, line 67: this was a parent route refresh for route `agenda`, changing to `this.router.refresh('agenda');`

app/routes/newsletter/nota-updates.js
```diff
@@ -20,7 +20,7 @@ export default class NewsletterNotaUpdatesRoute extends Route {
   *pollModel() {
     while (true) {
       yield timeout(3 * 60000);
-      this.refresh();
```
ValenberghsSven: calling `router.refresh()` in a route refreshes the parent route it seems.
just using local `this.refresh()` only reloads our own model which is what we need in this case.

app/controllers/publications/overview/_base.js
```diff
@@ -42,6 +46,6 @@ export default class PublicationsOverviewBaseController extends Controller {
 
   @action
   reload() {
-    this.send('reloadModel');
```
ValenberghsSven, line 45: this is too aggresive, it also reloads the publications navbar.
replacing with a `this.router.refresh('publications.overview._base');`
It looks similar to a local `router.refresh()` since we are in `_base`, but it seems to be reloading route `publication.overview` instead.
It makes sense in that `_base` is not a real route but all routes do extend it.
Since we did not declare this route in `router.js`, the closest real route from `_base` is the parent `publication.overview`.

ValenberghsSven: also using in place rather than an action

app/controllers/publications/publication/decisions/index.js
```diff
@@ -27,13 +28,13 @@ export default class PublicationsPublicationDecisionsIndexController extends Con
...
 
-    this.send('refresh');
+    this.router.refresh();
     this.closeReferenceDocumentUploadModal();
   }
 
   @dropTask
   *deletePiece(piece) {
     yield this.publicationService.deletePiece(piece);
-    this.send('refresh');
```
ValenberghsSven: both of these are too aggressive.
Before we only reloaded the documents list on adding or deleting.
Now the parent route is refreshed.
using `this.router.refresh('publications.publication.decisions.index');` instead has the same effect as pre refactor.

ValenberghsSven:
I think we missed a point of the `router.refresh()`.
Refreshing from router only refreshes that route and the nested child routes (pre refactor)
Not passing any route refreshes all active routes, all the way from application. (refresh from controller)
Excerpt from ember docs.

>Refreshes all currently active routes, doing a full transition. If a route name is provided and refers to a currently active route, it will refresh only that route and its descendents

I am going to add specific routes to each of those refreshes, to avoid a full route reload unless we need it.

ValenberghsSven:
Only a few test fail, mainly in publication.
Nearly there! I'll look at these tomorrow

ValenberghsSven (approved):
Not fully convinced this is the way to go.
But it works and you know exactly what route is going be reloaded.
No bubbling to parent, but possible errors if you pass a route that is note active.

Merging for now, could take different approach in different ticket if desired

PR #1606, opened 2023-01-19
refactor: replace all usages of moment.js with date-fns

sergiofenoll:
Removes usages of Moment.js with fitting equivalents from date-fns. Some helpers & utils where not used at all or used only once, so where possible they were deleted or inlined. The `date` and `datetime` transforms overriding the ember-mu-transforms were deleted, as we no longer need to cast dates into moments.


erikap (approved):

ValenberghsSven (approved):
Merging this, only flaky tests failing at the time

PR #1608, opened 2023-01-19
KAS-3757-indicate-closed-or-opened-on-agenda

StuyckRuben:


app/components/agenda/agenda-header.hbs
```diff
@@ -20,6 +20,11 @@
                 data-test-agenda-header-kind
                 class="auk-u-text-muted auk-body-1"
               >
+                {{#if this.isFinalAgenda}}
```
sergiofenoll, line 23: You can rewrite this condition to `{{#if @meeting.agenda}}` and skip the task & get automatic reactivity to boot. Right now, it won't matter in practice because if the condition changes it always happens after a refresh (so this components gets re-rendered and the `task` that loads this data is recalculated) but I think it'd be better to have it work reactively from the get-go. But it's a small detail and isn't really blocking.

app/components/agenda/agenda-header.js
```diff
@@ -19,6 +21,20 @@ export default class AgendaHeader extends Component {
...
+
+    this.loadAgendaIsFinal.perform();
+  }
+
+  @task
+  *loadAgendaIsFinal() {
+    const meeting = yield this.args.meeting;
+    const agenda = yield meeting.agenda;
+    this.isFinalAgenda = isPresent(agenda?.get('id'));
+  }
```
sergiofenoll: Since you're awaiting in the previous line via the `yield` keyword, Ember returns you an actual object instead of a proxy, so using `.get()` is no longer necessary. You could write this as `agenda?.id`.

sergiofenoll (changes requested):
You need to add `await this.args.meeting.belongsTo('agenda').reload();` to [this](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/b04942459206cb030a96ca742d861ea482391a7d/app/components/agenda/agenda-header/agenda-version-actions.js#L537-L542) method. Otherwise the `meeting` <=> `agenda` relation doesn't get reloaded. If a user closes/reopens a meeting, the pill won't update (without a manual page refresh).

app/components/agenda/agenda-header.hbs
tdn: I actually asked Ruben to do it like this because he got feedback on his last PR to be careful with async relationships from Erika...

I'd be inclined to leave it like this, since it's also semantically more clear what this check does this way.

But like you said, either will work.

sergiofenoll: Ah, I see. I think in the other PR it's about a slightly different situation.
My (possibly faulty!) understanding is that you can't rely on the async relationship's value or truthiness in JS-code (so for example inside an action you can't do `if (this.args.meeting.agenda) {}` because that will always be a truthy condition due to it being a promise proxy object). But it's completely valid in the template, because there Ember picks up on it being a promise and a proxy and it will re-render the template as the promise settles.
But yeah, just a tiny detail that doesn't need to be changed, but might be good to keep in mind (personally, it took me quite a while to start to understand how the async nature of relationships impacts both template-land and JS-land).

sergiofenoll (approved):

app/components/agenda/agenda-header.js
```diff
@@ -19,6 +21,20 @@ export default class AgendaHeader extends Component {
...
+
+  constructor() {
+    super(...arguments);
+
+    this.loadAgendaIsFinal.perform();
+  }
+
+  @task
+  *loadAgendaIsFinal() {
+    const meeting = yield this.args.meeting;
```
erikap: This `yield` doesn't seem necessary since `@meeting` is not a promise.

app/components/agenda/agenda-header.hbs
tdn: Didn't know Ember was smart that way :-) But makes sense!

erikap: Your understanding is correct @sergiofenoll :+1:  
I agree `{{#if @meeting.agenda}}` is better here, since the template will indeed automatically update if it changes.

tdn (approved):

sergiofenoll:
@ValenberghsSven failing tests seem to be flaky unrelated tests (publication stuff). I think we can merge this one.

ValenberghsSven:
@sergiofenoll Agree

PR #1619, opened 2023-01-26
KAS-3816: Stop using Observer and Evented APIs for jobs

sergiofenoll:
Now, the job monitor invokes the callback function that we previously ran when the didEnd event occurred. Events are no longer necessary because we can just check the job status inside the job monitor polling mechanism.


sergiofenoll:
Only failing test the concept scheme failing, unrelated to the changes here. Should still be tested locally, but if it works I think we can just merge this one.

ValenberghsSven (approved):
Locally tests and downloads seem fine.

PR #1621, opened 2023-01-26
Feature/kas 3847 extra checks on sending newsletter

sergiofenoll:
Also unifies the request to one request on news-item to ensure that
we're checking correctly. Otherwise if we check for notas and the
existence of themes separately, we'd allow notas in newsletter without
themes as long as there exists a nota with a theme.

### DIFF
--- a/app/components/newsletter/newsletter-header-overview.js
+++ b/app/components/newsletter/newsletter-header-overview.js
@@ -206,7 +206,6 @@ export default class NewsletterHeaderOverviewComponent extends Component {
       'filter[in-newsletter]': true,
     })) > 0;
 
-    // return hasDocumentPublicationPlanned && hasThemes && hasNotas;
     return hasDocumentPublicationPlanned && hasNotasWithThemes;
   }
 
--- a/app/components/newsletter/newsletter-header-overview.js
+++ b/app/components/newsletter/newsletter-header-overview.js
@@ -199,17 +199,15 @@ export default class NewsletterHeaderOverviewComponent extends Component {
     const themisPublicationActivity = themisPublicationActivities.find((activity) => activity.scope.includes(CONSTANTS.THEMIS_PUBLICATION_SCOPES.DOCUMENTS));
 
     const hasDocumentPublicationPlanned = isPresent(themisPublicationActivity?.plannedDate);
-    const hasNotas = (await this.store.count('agendaitem', {
-      'filter[agenda][:id:]': agenda.id,
-      'filter[type][:uri:]': CONSTANTS.AGENDA_ITEM_TYPES.NOTA,
-    })) > 0;
-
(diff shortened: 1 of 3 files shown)


app/components/web-components/vl-modal-verify.js
```diff
@@ -28,7 +28,7 @@ export default class WebComponentsVlModalVerify extends Component {
   }
 
   get verifyButtonText() {
-    return this.intl.t(this.args.buttonText ?? 'delete');
+    return this.args.verifyButtonText ?? this.intl.t('delete');
   }
```
ValenberghsSven, line 32: why this here?
Seems like a commit for a different PR

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -199,17 +199,15 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
-    const hasThemes = (await this.store.count('news-item', {
+    const hasNotasWithThemes = (await this.store.count('news-item', {
       'filter[agenda-item-treatment][agendaitems][agenda][:id:]': agenda.id,
+      'filter[agenda-item-treatment][agendaitems][type][:uri:]': CONSTANTS.AGENDA_ITEM_TYPES.NOTA,
       'filter[:has:themes]': true,
+      'filter[in-newsletter]': true,
     })) > 0;
 
-    return hasDocumentPublicationPlanned && hasThemes && hasNotas;
+    // return hasDocumentPublicationPlanned && hasThemes && hasNotas;
```
ValenberghsSven: this comment can be removed

app/components/web-components/vl-modal-verify.js
ValenberghsSven: Found a hardcoded dutch string because of this.
`@verifyButtonText="Verwijderen"` in `subcase-header.hbs` line 86
Removing it in favor of default translation of 'delete'

ValenberghsSven (approved):
The check should be fine.
Existing tests on it keep working so logic should be good.

app/components/web-components/vl-modal-verify.js
ValenberghsSven: fixed it

app/components/newsletter/newsletter-header-overview.js
ValenberghsSven: fixed it

app/components/web-components/vl-modal-verify.js
sergiofenoll: This was a mistake from the Octane refactor PR. Noticed it here(because the modals for publication in Kort Bestek had a blue "Verwijderen" button) and decided to fix it here because it was such a small change to begin with.

PR #1622, opened 2023-01-26
Kas 3789

kobemertens:
https://kanselarij.atlassian.net/browse/KAS-3789

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/362


app/config/permissions.js
```diff
@@ -65,6 +65,7 @@ const groups = [
       'view-document-version-info',
       'view-documents-before-release',
       'view-decisions-before-release',
+      'view-confidential-cases',
```
ValenberghsSven: This is what was requested in the ticket, but is somewhat confusing.
A `case` or `decisionmaking-flow` itself is never confidential, subcases are.
Is a case with 5 subcases where 1 subcase is confidential a confidential case?
According to current logic, it is.
In that light the permission name does make sense.

I suggest:
- in any case, add the permission to the commented list above
- keep the name but mention how a case can be considered confidential when 1 of its subcase is confidential.

app/controllers/cases/index.js
```diff
@@ -25,18 +25,12 @@ export default class CasesIndexController extends Controller {
...
-      },
-    }
   ];
   sizeOptions = Object.freeze([5, 10, 20, 50, 100, 200]);
   page = 0;
   size = 20;
 
   sort = '-opened';
-  showArchived = false;
   @tracked isLoadingModel;
```
ValenberghsSven: This change was requested, but now the view is showing both unarchived and archived cases without a clear difference.
To the user, it seems like "delete case" action does nothing.
It would make more sense to never show archived cases or to show archived cases with a different css class (greyed out)
Because the feature "restore case" is still a thing that users use (recent support issue when it was broken) we still need to offer it in some way and you can't access it from search
Could we instead toggle showing all normal cases only / all archived cases only.
Ideally, I was thinking of an action menu inside the case route on individual cases but seems out of scope.

app/templates/search/cases.hbs
```diff
@@ -6,17 +6,34 @@
...
+        @label={{t "search-include-archived"}}
+        @checked={{this.includeArchived}}
+        @onChange={{set this "includeArchived"}}
+      />
+      <AuControlCheckbox
+        @label={{t "search-archived-only"}}
+        @checked={{this.archivedOnly}}
+        @onChange={{set this "archivedOnly"}}
+      />
+    </div>
```
ValenberghsSven: both options seem somewhat mutually exclusive, radiobuttons?

ValenberghsSven (changes requested):
Mainly changes needed in the cases main route.
archiving/restoring case feature needs to be handled in some different way.
Now both are in the same view with no visual differences

ValenberghsSven:
FYI, the change in the backend can be merged already since that is an addition to the index and should not break anything.
So already merging that one. I don't expect any more index changes from this PR so removing that backend branch.
Continue on development branch for app-kaleidos.

app/templates/search/cases.hbs
kobemertens: They are not mutually exclusive but `archivedonly` implies `includearchived`, otherwise there will always be no results. So using a radiobutton seems like a good solution to me.

app/controllers/cases/index.js
kobemertens: In the search results rework, there is a visual indication that a case is archived. I'm not sure about the usefulness of having the ability to see archived and non-archived cases in the same view but I don't see why it should not be possible.
image

app/templates/search/cases.hbs
kobemertens: Implemented in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1622/commits/0dc59ed5e8559112a0795a206578cdd8a17728b9

app/controllers/cases/index.js
ValenberghsSven: I was referring to the datatable in the general case route (not search)
That table now shows both "deleted" and "not deleted" cases, while previously it showed only all "not deleted" cases.
You then had to trigger the "show deleted" button and the list would contain both (but not ordered or visually different).
So finding any specific archived case was just a no go.

With this suggested change, now you always see all cases no matter if you archive them or not which may lead users to conclude that "delete case" or "restore case" feature does nothing as the list stays the same.
Something needs to change there, but reworking it is out of scope.
So my suggestion would be to change the previous behavior "show deleted" to "show only deleted" so we can toggle between both lists (deleted and not deleted)
It would make it easier to find deleted cases (used but only seldom by users) and restore them.

ValenberghsSven:
@Kobe I will take over this branch and implement those final things (comments in permissions file and cases route datatable correction)
If the "show deleted" feature needs to really be removed from that one view then we will make a follow up ticket.

ValenberghsSven (approved):
implemented the requested changes (by me).
Tested it locally with the toggling of archived and the changed test spec.
rest of the changes are good.

PR #1623, opened 2023-01-26
KAS-3771 more dependencies bumped

ValenberghsSven:
draft, bumped many dependencies


ValenberghsSven:
only failed test is flaky (conceptshemes)

app/components/auk/tab.js
```diff
@@ -41,4 +41,12 @@ export default class Tab extends Component {
...
     return null;
   }
+
+  get query() {
+    if (isPresent(this.args.query)) {
+      return this.args.query;
+    }
+    // return empty object
+    return [];
+  }
```
ValenberghsSven, line 51: The reason for this and removal of `@query` elsewhere is that with recent dependency version, passing this as undefined throws errors.
Also `@query` is only used when used in `<auk::tab>`

ValenberghsSven:
jenkins green run.

package.json
```diff
@@ -49,106 +49,107 @@
...
     "ember-cli-string-helpers": "^6.1.0",
     "ember-cli-terser": "^4.0.2",
-    "ember-cli-update": "^1.0.0",
-    "ember-click-outside": "^2.0.0",
+    "ember-cli-update": "^1.0.1",
+    "ember-click-outside": "^5.0.1",
     "ember-code-snippet": "^3.0.0",
-    "ember-composable-helpers": "^4.4.1",
-    "ember-concurrency": "^2.2.1",
+    "ember-composable-helpers": "^5.0.0",
```
sergiofenoll, line 91: ~~New version of ember-composable-helpers has renamed the `contains` helper to `includes`. This was deprecated for a while but now it outright doesn't exist anymore. We don't use it in the app, but it's used in `ember-vo-webuniservum-data-table`. I *think* that if you merge development again into this branch it should just work, because we don't use DataTable anywhere then. Currently checking locally if that's the case.~~ I hadn't pulled your latest commit, you have already merged development :D

sergiofenoll (approved):

PR #1624, opened 2023-01-27
Feature/kas 3793 kort bestek search

sergiofenoll:
Adds `search/news-items` route that displays news items in the same way they're currently displayed in the `newsletters/search` route.

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/363

KAS ticket: https://kanselarij.atlassian.net/browse/KAS-3793


app/templates/search/news-items.hbs
```diff
@@ -0,0 +1,110 @@
...
+      data-test-route-search-news-items-data-table={{true}}
+      @content={{this.model}}
+      @page={{this.page}}
+      @noDataMessage={{t "no-results-found"}}
+      @size={{this.size}}
+      @sort={{this.sort}}
+      @class="auk-table auk-table--striped auk-table--vertical-align-top"
+      @isLoading={{this.isLoadingModel}}
+      @onClickRow={{this.navigateToNewsletter}}
+      as |table|
```
ValenberghsSven: use `<KdataTable`

ValenberghsSven (changes requested):

app/templates/search/news-items.hbs
ValenberghsSven: Did I just miss the commit by a minute :D :man_facepalming:

app/routes/search/news-items.js
```diff
@@ -0,0 +1,132 @@
...
+      refreshModel: true,
+      as: 'aantal',
+    },
+    sort: {
+      refreshModel: true,
+      as: 'sorteer',
+    },
+  };
+
+  textSearchFields = ['title^3', 'htmlContent'];
```
ValenberghsSven: not searching on `subtitle` ?
we transpose the titles on creation between agendaitem and newsItem
agendaitem.title == newsItem.subtitle, agendaitem.shortTitle == newsItem.title

Since title (also named long title) is optional on agendaitems, not searching for subtitles on newsitem will filter some results when the titles contain different keywords.

ValenberghsSven (changes requested):

app/routes/search/news-items.js
sergiofenoll: Hmm, good point. I just applied the ticket literally, where the fields are stated as:

- titel (boost ^3)
- HTML inhoud

In other tickets title & subtitle distinction is explicitly stated, so I assumed it was intentional, but I think you may be right here.

ValenberghsSven: I marked a change request, but I did see that you are following the jira ticket.
If we persist on only using `title` and `htmlContent` then this is fine and the request can be ignored.
Wanted to bring it up

ValenberghsSven (approved):

PR #1631, opened 2023-02-02
test: update the profile names for Kabinet roles

sergiofenoll:
They're now Kabinet*s*dossierbeheerder and
Kabinet*s*medewerker.

Should fix one of the failing tests on dev branch.


sergiofenoll:
Failing user management test no longer fails. Merging this one.

PR #1635, opened 2023-02-03
Feature/kas 3462 search QoL

sergiofenoll:
- Stops resetting search query params when navigating to the search route
- Resets agendaitem filter param when navigating away from the route
- Make hierarchical back button in agenda view act like a brower back button
- Store scroll position in search route so that you can scroll back to the right position when going back

This PR adds a new dependency, `ember-router-scroll`, which tracks the scroll position when navigating away from a route so that it can be put back when going back/forward in history. Due to limitations of the addon, it only works for the search routes.
The search routes render the content inside a div with id `#content`, which the addon uses to determine the scroll position. Because other routes don't render in this way, we won't track their scroll position. The addon doesn't support using other selectors to determine which container's scroll position to store (it has to be an id). We could potentially extend/fork the addon to let us set a selector per route.
For the other nice to have, about updating the URL (and history state) quicker (before all the model hooks are done): we could rework our main routes from the ground up by making their model hooks resolve very quickly (and thus load all of our data through other means), but that seems like a lot of work. Alternatively there's thid addon: https://github.com/raido/ember-refined-route-substate-url which seems to aim to solve our own problem, but I haven't gotten around to testing it. **EDIT:** I installed the package locally and it seems to work as expected, but it doesn't seem to be a very mature addon, so we should spend some time figuring out how it works before we try to add it to the project.


sergiofenoll:
This can be check (or merged immediately, up to the merger).

Only weird thing immediately coming to mind is that the "Filters verwijderen" knop does not clear the documents' tab document type filters due to how they're stored. If desired I can try fixing it quickly though.

sergiofenoll:
- Nice-to-haves removed, as they weren't working (anymore)
- Always remembering filters removed, brought back old behaviour of clearing filters when clicking on the search tab
- Just like before, when going back in history to the search tab the filters are still there (this was always the case)
- Agendaitem filter gets cleared after navigating away, as requested
- HierarchicalBack button in agenda route is now an extra history back button

Extra:
- Re-worked the minister filter component a little bit to not require records, instead working on IDs.

@ValenberghsSven feel free to wait for this to pass the tests or the check locally, but this should be mergeable.

EDIT: I think you could also change the base branch here to ACC and merge it there (and then merge to development) if you want to make sure you don't take any unexpected changes to ACC. It shouldn't really matter because the last time DEV was merged in here was quite a while ago, but not sure.

ValenberghsSven (approved):
looks ok

PR #1636, opened 2023-02-03
KAS-3898: Upgrade van lblod/ember-rdfa-editor package & component

brenner-company:
**Tickets binnen Jira:** 
- https://kanselarij.atlassian.net/browse/KAS-3898


ValenberghsSven:
for tests, it seems `cy.get(dependency.rdfa.editorInner).should('be.empty');` is just no longer true on an empty editor.
It always seems to contain a paragraph with a trailingbreak
image

Need to figure out how to assert 0 content

ValenberghsSven:
branch only fails on a known flaky test (search-other-data-spec)

.woodpecker/.feature.yml
```diff
@@ -3,8 +3,7 @@ pipeline:
     image: plugins/docker
     settings:
       repo: "${CI_REPO_OWNER%%-vlaanderen}/${CI_REPO_NAME}"
-      tags: "feature-${CI_COMMIT_BRANCH##feature/}"
+      tags: "feature-${CI_COMMIT_BRANCH##DES/enhancement/}"
```
ValenberghsSven: - [x] roll this back and merge?

ValenberghsSven (approved):

.woodpecker/.feature.yml
ValenberghsSven: We could add a separate woodpecker config for this if we use it frequently

PR #1642, opened 2023-02-10
Hotfix/consistent pagination options

sergiofenoll:


ValenberghsSven:
- via MR spec was failing because page size 25 is needed and was 20 (need to adapt test eventually to be less flaky, our publication is in the bottom of the list) 
- delete contact person flaky also solved here, the random promise list of deletes is a bad strategy since we should delete in reverse order of how we added in most cases to avoid ember store update issues.
- improved search spec pagination test to not expect ("aantal=") when testing the default option

PR #1646, opened 2023-02-15
fix warning message shows when bronbestand was docx file before conve…

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3828

I added a workaround using a timeout that shows the warning after 0.5s if there's a docx file

### DIFF
--- a/app/components/documents/document-view.hbs
+++ b/app/components/documents/document-view.hbs
@@ -6,7 +6,7 @@
       class="pdf-view"
     >
     </object>
-  {{else}}
+  {{else if this.showWarning}}
     <div class="auk-u-p-4">
       <Auk::Alert
         @skin="warning"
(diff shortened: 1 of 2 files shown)


ValenberghsSven (changes requested):
using the timeout workaround is not really solving the issue.
I think the main issue is that we are silently awaiting in the parent template:
`{{#let (or this.selectedVersion.file.derived this.selectedVersion.file) as |file|}}` in `document-preview-modal.hbs`.

Perhaps a more reliable fix would be to properly load the data in that component with a task in the constructor.
The IF in the template is fine as long as we loaded the data from `file` and `file.derived`
We could wrap that with a `{{#if this.task.isRunning}} > loader`

Choosing a different version ẁith `setSelectedVersion` would then need to rerun the loading task

ValenberghsSven:
Alternative with data loading = https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1679

ValenberghsSven:
Closing for now in favor of better data loading PR

PR #1614, opened 2023-01-24
KAS-3819-hide-toon-aangepaste-punten-toggle-in-agenda

StuyckRuben:

PR #1649, opened 2023-02-17
Feature/kas 3913 revert toon aangepaste punten toggle in agenda

tdn:
This is a revert of https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1614

Automatic revert on github failed because of conflicts. Checked out the current `development` branch, and then reverted the 3 commits in the original PR (not the merge) 1 by 1. Then resolved the conflicts (by always choosing `development`) and added the translation keys back in.

I left the tests to the newer ones that check for the green markers, especially in case the users decide later on that this function can be removed after all... @ValenberghsSven you can decide to add the `toggleShowChanges` action back in the tests if you prefer to have both of them, but for me this is not required.

### DIFF
--- a/app/components/agenda/agenda-detail/sidebar.hbs
+++ b/app/components/agenda/agenda-detail/sidebar.hbs
@@ -7,6 +7,17 @@
         </h4>
       </Auk::Toolbar::Item>
     </Auk::Toolbar::Group>
+    <Auk::Toolbar::Group @position="right">
+      <Auk::Toolbar::Item>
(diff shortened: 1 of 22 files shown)


ValenberghsSven:
If the feature stays, testing it would be advised so we catch it should it stop working. (also good to test if it works the same way it did before remove/revert)
Will do this + check the icons

tdn:
Ok, thanks! Then it'd probably be best to do both the old and new tests so at least that effort wasn't for nothing?

ValenberghsSven (approved):

PR #1651, opened 2023-02-21
Upgrade/ember v4.4

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3845

Up to date with DEV.
Jenkins was green at build 31.
Every merge on DEV could break something so need to verify every test if it's flaky or broken.
Known flaky tests = new-document-viewer.spec, new-document-widget.spec, first test in newsitem-spec.js

Tried to separate commits to make the change log easier to look at and give a reason why I did it.


ValenberghsSven:
Only 1 or 2 tests are flaky (new-document-viewer.spec was the last one with actual issues).
almost good to go

app/models/document-type.js
```diff
@@ -1,5 +1,5 @@
 import { hasMany } from '@ember-data/model';
 import Concept from './concept';
 export default class DocumentType extends Concept {
-  @hasMany('document', { async: true }) documents;
+  @hasMany('document', { inverse: null, async: true }) documents;
```
tdn, line 4: what does this `inverse: null` do?

app/modifiers/plausible-click.js
tdn, line 1: I assume these are the result of upgrading plausible as well? 
Do the clicks still register with these changes? (I don't have a local plausible instance running to test)

tdn (approved):
Discovered a few minor bugs/peculiarities, but nothing that didn't exist before the upgrade.

Looks good and tests well locally as far as I can tell.

Didn't review cypress tests, but assuming these are OK since they all pass on jenkins.

app/models/document-type.js
ValenberghsSven: Was getting a deprecation warning (for ember v5), you have to mention the inverse of a relation.
It was one of the only relation that did not have this.
ex: 
in `document-container` model:
`@hasMany('piece', { inverse: 'documentContainer', async: true }) pieces;`
in `piece` model:
`@belongsTo('document-container', { inverse: 'pieces', async: true }) documentContainer;`
The explicit mentioning helps the ember store update the list of pieces on documentContainer when a new piece sets the belongsTo to that document-container.

For concepts however, we often don't specify the inverse.
It should not have any negative effects.

Also, from looking at this code. We wouldn't be able to get `documentType.documents` in ember because it does not match the definition in domain models.
image

The relation should be
`@hasMany('document-container', { inverse: null, async: true }) documentContainers;`
and maybe `inverse: 'type'` but normally we don't do that (since concept does not really know us inversely in `concept.js` model)
Also unsure how that works with polymorphism.

app/modifiers/plausible-click.js
ValenberghsSven: I tested it before the latest Edit by Sergio.
Before my changes, the clicks did not trigger (`didInstall` no longer works). After my changes to `modify`, they did.
Sergio's changes just streamline the refactor. My local variables were not needed.
I can help setup a local plausible.

ValenberghsSven:
Might merge this tomorrow. running some jenkins tonight

PR #1652, opened 2023-02-22
added 2 buttons to move agendaitems up and down in the list

StuyckRuben:


app/components/agenda/agenda-overview.hbs
```diff
@@ -77,6 +77,9 @@
                   @showFormallyOkStatus={{@currentAgenda.status.isDesignAgenda}}
                   @isEditingFormallyOkStatus={{@isEditingOverview}}
                   @showDragHandle={{this.canDragAgendaitems}}
+                  @onMove={{perform this.move}}
+                  @isFirst={{and (eq index 0) (eq indexItems 0)}}
+                  @isLast={{and (not (has-next agendaitemGroup @notaGroups)) (not (has-next agendaitem agendaitemGroup.agendaitems ))}}
```
sergiofenoll: Whitespace

```suggestion
                  @isLast={{and (not (has-next agendaitemGroup @notaGroups)) (not (has-next agendaitem agendaitemGroup.agendaitems))}}
```

app/components/agenda/agenda-overview.js
```diff
@@ -26,4 +27,34 @@ export default class AgendaOverview extends Component {
   get canDragAgendaitems() {
     return this.canEdit && this.args.isEditingOverview;
   }
+
+  @task
```
sergiofenoll: Why is this a `task` instead of an `action`?

app/components/agenda/agenda-overview.js
```diff
@@ -26,4 +27,34 @@ export default class AgendaOverview extends Component {
...
+    let itemArray = [];
+    if (changedAgendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA) {
+      for (const notaGroup of this.args.notaGroups) {
+        if (notaGroup.agendaitems) {
+          for (const nota of notaGroup.agendaitems) {
+            itemArray.push(nota);
+          }
+        }
+      }
+    } else if ( changedAgendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT) {
```
sergiofenoll: Whitespace

```suggestion
    } else if (changedAgendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT) {
```

app/components/agenda/agenda-overview.js
```diff
@@ -26,4 +27,34 @@ export default class AgendaOverview extends Component {
...
+      itemArray = [...this.args.announcements];
+    }
+    for (let i = 0; i < itemArray.length; i++) {
+      if (itemArray[i].id === agendaitem.id) {
+        agendaitemIndex = i;
+        break;
+      }
+    }
+    if(agendaitemIndex > -1 &&
```
sergiofenoll: Whitespace

```suggestion
    if (agendaitemIndex > -1 && 
```

ValenberghsSven:
Wanted to approve this, but during test I found out you can use the feature in combination with "toon aangepaste punten".
I don't think we want this?
It has very strange side effects:
ex: 1 and 4 are changed. Moved 1 down > 4 became 1 and 1 became 2
When using "toon alle punten" I got 1, 2, 2, 3

I'm guessing this problem also exists for the current dragging feature..

Perhaps we leave it as-is but change the text above?
image

Add something like "Opgelet!. Slepen of verplaatsen van agendapunten in combinatie met "toon aangepaste punten" kan onverwachte resultaten geven." ?

tdn:
Oh I think we definitely want to disable this when the 'toon aangepaste agendapunten' is toggled! Didn't think it would be enabled even with the current dragging function 

@StuyckRuben can you add this to the PR please? Thanks!

ValenberghsSven (approved):
Changed it so that the drag or move buttons are not available when we are showing the modified items to avoid items being numbered incorrectly.

PR #1658, opened 2023-02-28
Feature/kas 3927 all type search

kobemertens:


app/templates/search/all-types.hbs
```diff
@@ -0,0 +1,134 @@
...
+      @size={{this.size}}
+      @numberOfItems={{this.model.results.length}}
+      @enablePagination={{false}}
+      @onClickRow={{this.navigateToResult}}
+    >
+      <:body as |row|>
+        <td>
+          {{#if (eq row.name "cases")}}
+            <Search::ResultCards::Case
+              @created={{row.data.attributes.created}}
```
sergiofenoll: Should be `@sessionDate={{row.data.attributes.sessionDates}}`

app/controllers/search/all-types.js
```diff
@@ -0,0 +1,126 @@
...
+    {
+      sort: {
+        type: 'string',
+      },
+    },
+  ];
+
+  sortOptions = [{ value: '-session-dates', label: this.intl.t('date') }];
+
+  @tracked sort;
```
sergiofenoll, line 22: Do we still need this? I don't think we're sorting in this route.

app/controllers/search/all-types.js
```diff
@@ -0,0 +1,126 @@
...
+
+  constructor() {
+    super(...arguments);
+    this.sort = this.sortOptions[0].value;
+  }
```
sergiofenoll: Do we still need this? I don't think we're sorting in this route.

app/controllers/search/all-types.js
```diff
@@ -0,0 +1,126 @@
...
+
+  @action
+  selectSize(size) {
+    this.size = size;
+  }
+
+  @action
+  selectSort(event) {
+    this.sort = event.target.value;
+  }
```
sergiofenoll: Can also go, I think.

sergiofenoll (changes requested):
Looks good :+1:
Just needs the change to the case card in all types.

PR #1661, opened 2023-03-02
KAS-3883: Bundle agenda documents per mandatees

sergiofenoll:
Allows users to select the ministers for which to download agenda documents.

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/373
- [ ] https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/5

JIRA: https://kanselarij.atlassian.net/browse/KAS-3883


sergiofenoll:
Deze nog niet mergen, tekstje bij modal moet nog aangepast worden.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -36,6 +36,9 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
   @tracked showPlanDocumentPublicationModal = false;
   @tracked showConfirmPublishThemis = false;
   @tracked showConfirmUnpublishThemis = false;
+  @tracked showDownloadDocuments = false;
+  @tracked showDownloadDocuments = false;
```
ValenberghsSven, line 40: This double has been removed in follow up PR. Can ignore here.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -256,6 +259,7 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
     debug('Checking if archive exists ...');
     const jobPromise = fetchArchivingJobForAgenda(
       this.args.currentAgenda,
+      this.selectedMandatees,
       this.store
     );
     const [name, job] = await all([namePromise, jobPromise]);
```
ValenberghsSven, line 265: Saw this in next PR, the `namePromise` is always the same.
Might cause file overwrites on local system if someone decided to download all documents 1 mandatee at a time or something. I think in downloads a number will be appended by the browser (1) (2) ?

translations/nl-be.json
```diff
@@ -1159,6 +1159,7 @@
   "also-found-in": "Ook gevonden in",
   "uploaded-in-thing": "Geüpload in {thing}",
   "creation-date": "Aanmaakdatum",
+  "download-agenda-documents-info": "Vink de ministers aan wier documenten je wilt downloaden. Documenten zonder minister worden altijd toegevoegd aan het bestand.",
```
ValenberghsSven: We always grab the docs without mandatees (which should be all the announcements and the "verslag").
Hypothetically, if someone downloads all documents per mandatee they will always have these so results in many duplicates across zips (there also will be duplicates when multiple mandatees are on 1 agendaitem but that is unavoidable).

Do we want a select box for all documents without mandatees that is checked by default. Could be a follow up PR
Perhaps even a filter on the `agendaItemType`

sergiofenoll: I don't disagree, but this comes verbatim from the FP: https://www.notion.so/Feature-Passports-Kaleidos-4e5ca9a91ae24b609ee2f56b2166f370?p=645bcc0f0d544e0c961d6eb3a64d815e&pm=c

You could ask there/ask Hannes directly and we can add it as a follow up ticket?

sergiofenoll:
Added this explanatory text from Johan and a short message when an agenda has no ministers.

image

Not sure how clear it is but I'll ask for Hannes' input.

translations/nl-be.json
ValenberghsSven: Follow up PR is fine. Just thinking ahead. We could have waited for user feedback on the issue.

ValenberghsSven (approved):

PR #1672, opened 2023-03-06
added selection of document-types in use, and expanded labeling options for RadioDropdownComboSelect

tdn:
Backend PR to be merged and deployed with this: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/374

It's possible we might get some different intructions from the users/design here on where to show the full or abbreviated labels. However, this is much easier to verify for them when this is merged into accept, so this user input is not blocking.

Update: users approved of current method, so this can be merged after review


sergiofenoll:
This is already broken in DEV so not caused by changes here, but perhaps we want to fix it while here.

The search box when selecting "Andere" as document type of newly uploaded pieces doesn't show typed text:
[Screencast from 2023-03-07 16-57-50.webm](https://user-images.githubusercontent.com/22411874/223477030-9e7f5081-1cef-4b24-9ff2-ea883de0e475.webm)

EDIT: If we add `@renderInPlace` to the `PowerSelect` in `Utils::RadioDropdownComboSelect`, the input works as expected but then the dropdown gets rendered inside the modal, so we get a scrollbar when it's rendered. We could enable `@overflow` on the modals but that would break the view when uploading multiple files. I think it's best to just enable `@renderInPlace`.

tdn:
Ok... so we add @renderInPlace so the component works and then contact @brenner-company to fix the UX in a separate PR?

ValenberghsSven (approved):
1 test failed = fixed on DEV already

PR #1673, opened 2023-03-06
fix: add an override for transitive ember-auto-focus dep

sergiofenoll:
It bumped ember-modifier to v4 but that breaks our app until we upgrade.


ValenberghsSven:
Maintainer fixed their borkings

PR #1675, opened 2023-03-07
cypress/KAS-3925-run-up-to-profile-tests-newsletter-route

MathiasZabot:


cypress/integration/all-flaky-tests/profile-minister.spec.js
```diff
@@ -320,4 +321,19 @@ context('Testing the application as Minister user', () => {
...
+      cy.get(agenda.agendaActions.optionsDropdown)
+        .children(appuniversum.button)
+        .click();
+      cy.get(agenda.agendaActions.navigateToNewsletter).forceClick();
+      cy.url().should('not.include', 'vergadering/6374F696D9A98BD0A2288559/kort-bestek');
+    });
+
+    // no other tests here because there is no acces to kort bestek
+  });
 });
```
ValenberghsSven: `.only` not needed.
Minister profile can see kort-bestek zebra view but will get errors if not released yet.

ValenberghsSven: image

Pretty sure all profiles can go to that page

ValenberghsSven (changes requested):
Think all profiles can see the zebra.
The error that would show on non-closed agendas I fixed in this PR.
We need to check what the profiles see on
- open agenda / closed agenda (pre release of decisions) should be same (test 1?)
- closed agenda (post release of decisions)

ValenberghsSven (approved):

PR #1678, opened 2023-03-08
Feature/kas 3884 beslissingen downloaden

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3884

- [ ] merge https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/6
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/384


app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -244,7 +244,7 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
   }
 
   @action
-  async downloadDocuments() {
+  async downloadDocuments(decisions = false) {
```
ValenberghsSven, line 247: fyi, I hinted at reusing the current logic to prevent duplication and to get a first draft working
Perhaps we could pass a type 'documents' or 'decisions' to make it more generic and expandable (even though there aren't really any other types at the moment)

app/utils/zip-agenda-files.js
```diff
@@ -21,11 +21,14 @@ async function constructArchiveName(agenda) {
...
 }
 
-async function fetchArchivingJob(agenda, mandateeIds) {
+async function fetchArchivingJob(agenda, mandateeIds, decisions) {
   let url = `/agendas/${agenda.id}/agendaitems/pieces/files/archive`;
   if (mandateeIds.length) {
     url += '?' + (new URLSearchParams({ mandateeIds }).toString());
   }
+  if (decisions) {
+    url += '?decisions' + (new URLSearchParams({ decisions }).toString());
```
sergiofenoll: I should check before asking, but does this work if you also select ministers? Query params in a HTTP request should be separated by an `&`, only the first query param gets a `?` prefixed. I suspect that this will turn the last mandatee ID in the list into `mandatee-uuid?` instead of `mandatee-uuid`.

sergiofenoll: Yeah just checked if you select ministers/pass `mandateeIds` to this function the service won't actually recognize that you passed the `decisions` query param because it's part of the mandateeIds array `?decisions`.

Try something like:
```js
url += '?' + new URLSearchParams( { mandateeIds, decisions }).toString();
```

This will spit out something like `?mandateeIds=id1,id2,id3&decisions=true`. You can try it in your browser console too. In case `mandateeIds` is an empty list it will be something like `?mandateeIds=&decisions=true` but the service should be able to cope with that.

sergiofenoll (changes requested):
Needs some small changes, see comments

app/utils/zip-agenda-files.js
ValenberghsSven: Fixed it some other way.

ValenberghsSven (approved):

PR #1679, opened 2023-03-08
KAS-3818 implement data loading to prevent banner showing

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3828

Alternative implementation with a data loading solution instead of workaround
Swapping selected version also needed to trigger a reload of the tracked file.


app/components/documents/document-preview/document-preview-modal.hbs
```diff
@@ -1,43 +1,45 @@
...
+    </span>
+  </:title>
+  <:body>
+    <div class="auk-panel-layout">
+      <div class="auk-panel-layout__main-content">
+        {{#if this.loadFile.isRunning}}
+          <WebComponents::LoadingPage />
+        {{else}}
+          <Documents::DocumentView @file={{this.file}} />
         {{/if}}
```
ValenberghsSven, line 28: This is the main part that changed in template 
`+` removing the `#let` which marked the entire file as changed
`+` renaming `file` to `this.file`

ValenberghsSven:
To test this, I proxied to DEV and uploaded a docx.
Without this fix, the banner showed for a split second.
With fix, it doesn't.
Added a timeout and we just show a loader now instead of showing the banner.
Could make a video later

ValenberghsSven:
Added a video of DEV vs local in the JIRA ticket

sergiofenoll (changes requested):
Because `file` is now a property stored in the component class instead of being derived from tracked objects, when a user uploads a new file the document view component won't update its content. Users will have to manually refresh to see the newly uploaded file.

Simplest fix is to pass a callback down the sidebar component to the details panel that gets called whenever a new file is uploaded and saved on the piece. This way we can reload the file(s) in the preview modal component.
"Proper" fix is probably to rework the details panel component a bit to respect DDAU and not mutate its arguments directly nor save it. That would also prevent other unwanted changes (e.g. if you're typing a new document title it also changes the title in the modal header before saving). But that's a bigger change that doesn't need to happen here per se.

Happy to implement the simple fix if you're busy btw.

ValenberghsSven:
Added the callback. Is this what you meant?
If not feel free to step in

sergiofenoll (approved):

ValenberghsSven:
Only failed test is flaky pagination on search

PR #1685, opened 2023-03-10
fix: back button in cases and subcases routes work as history back

sergiofenoll:


ValenberghsSven (approved):
Looks good, already cherry-picked in to ACC

PR #1687, opened 2023-03-10
fix: replace count meta in all types with one liner

sergiofenoll:
image


ValenberghsSven:
Branch was not up to date. Used the code with copy paste instead of merge

PR #1694, opened 2023-03-21
cypress/KAS-3985-new-tests-buttons-to-change-agendaitems-order

MathiasZabot:


ValenberghsSven (approved):
Looks good, checking if jenkins succeeds at least twice to avoid new flaky

ValenberghsSven:
only fail is flaky concepts

PR #1698, opened 2023-03-24
Feature/transition history

ValenberghsSven:
Browser history is somewhat out of our control.
Using history.back in our app may bring users outside our app if they used a link or from a new tab.

To counter this we could keep a "history" of our own.
This was too much for a hotfix to moved to new branch.
When using this we have to make sure we always keep our history when transitioning.
I have found instances where we lose this history in our own code.
some causes/ places:
- any table with `@onClickRow` and a `<LinkTo>` in it. We nagivate twice with the `transitionTo` from the first and the `<a href=..>` from the second. This seems to cause a full page reload (application route gets rebuild) and history is lost
- Using the document link on the overview of an agendaitem. This seems like a common `<a>` element and not a `<LinkTo>`. Other links to the same page work without losing history.

When using this method, we can't just fall back on browser history.
Because when out app transitions back, the browser history keeps going forward.
Falling back to it would mean we backtrack twice which is very confusing.

In theory it makes perfect sense as long as we have a perfect history of our own so we don't have to fallback to browser history.
Browser history should not be our concern

We would have to have a fallback on those "hierarchical back" buttons. Was experimenting with that until I decided to stash this code for higher priority work


app/routes/application.js
```diff
@@ -48,5 +49,6 @@ export default class ApplicationRoute extends Route {
     if (this.session.isAuthenticated && !this.currentSession.hasAccessToApplication) {
       this.router.transitionTo('accountless-users');
     }
+    this.transitionHistory.registerTransition();
```
sergiofenoll, line 52: Ember has built-in events for doing actions on route changes. Check [this for an example in our code](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/plausible.js#L14-L27) and [this guide from Ember](https://guides.emberjs.com/release/routing/preventing-and-retrying-transitions/)

ValenberghsSven:
Going to close this for now. History.back seems to be working ok, no support issues logged recently.

PR #1701, opened 2023-03-30
KAS-3893: Bump mu-cl-resources to feature branch

sergiofenoll:
Before we sorted agendaitems on `type,number`, which does not do what we expect, it sorts on the `type` relationship, which currently mu-cl-resources interprets as "sort on each property of this relationship". Additionally this results in faulty responses from the new construct-based fetches branch. What we really want is to sort on the `type`'s `position`, so let's just do that instead.

Related PR:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/372
- [x] https://github.com/kanselarij-vlaanderen/cache-warmup-service/pull/12

### DIFF
--- a/app/routes/agenda/agendaitems.js
+++ b/app/routes/agenda/agendaitems.js
@@ -36,7 +36,7 @@ export default class AgendaAgendaitemsRoute extends Route {
       'filter[agenda][:id:]': agenda.id,
       include: 'type',
       'page[size]': PAGE_SIZE.AGENDAITEMS,
-      sort: 'type,number',
+      sort: 'type.position,number',
     });
 
     // Ensure mandatee data for each agendaitem is loaded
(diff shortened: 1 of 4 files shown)


app/routes/agenda/agendaitems.js
```diff
@@ -36,7 +36,7 @@ export default class AgendaAgendaitemsRoute extends Route {
       'filter[agenda][:id:]': agenda.id,
       include: 'type',
       'page[size]': PAGE_SIZE.AGENDAITEMS,
-      sort: 'type,number',
+      sort: 'type.position,number',
     });
```
ValenberghsSven, line 40: Cache-warmup caches this query

ValenberghsSven (approved):
This PR looks ok, cache-warmup needs same changes

PR #1702, opened 2023-03-31
Feature/kas 3933 link user and organisations with mandataris

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3933
Add organization detail page
Add functionality to link organization to mandatee

- [ ] Merge https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/386


app/components/mandatees/mandatee-person-selector.js
sergiofenoll, line 1: This component is a bit strange, because it doesn't seem to take the `dateRange` into account *except* when searching. So if the dateRange is set to e.g. 2016 the list still shows the ministers from the current government.

But this behaviour is the same as when it was implemented in the `GenerateReportModal`, so it can stay as is here and we can have a look at it later.

app/routes/settings/users/user.js
sergiofenoll, line 1: In the user detail screen, we show the linked mandatee in the dropdown. We use `selectedPerson` to display the link/unlink button. You need to set the value of `selectedPerson` with `model.person` so that works on a new page load, otherwise we will show the selector with the linked mandatee but we won't show an unlink button.

You can do this by expanding the route class like this:

```diff
 import Route from '@ember/routing/route';
 import { inject as service } from '@ember/service';

 export default class SettingsUsersUserRoute extends Route {
   @service store;

+  person;

   model(params) {
     return this.store.findRecord('user', params.id, {
       include: 'memberships.organization,memberships.role,person'
     });
   }
+
+  async afterModel(model) {
+    this.person = await model.person;
+  }
+
+  setupController(controller, model) {
+    super.setupController(controller, model);
+    controller.selectedPerson = this.person;
+  }
 }
```

This will load in the `person` relationship after the model has been loaded, and it will then assign it to a `this.person` member of the route. Afterwards, during the `setupController` hook, we assign the person to the controller's `selectedPerson` member. This way it's pre-loaded when we display the route.

app/templates/settings/organizations/index.hbs
```diff
@@ -123,7 +123,22 @@
                   {{t "block-organization"}}
                 </AuButton>
               {{/if}}
+              <AuButton
+              @skin="naked"
+              role="menuitem"
+              {{on "click" (queue (toggle "showMandateeSelectorModal" this) (set this "organizationBeingLinked" organization))}}
+              >
+                {{t "connect to mandatee"}}
+              </AuButton>
```
sergiofenoll: This seems unused.

app/templates/settings/users/user.hbs
```diff
@@ -130,6 +130,45 @@
...
+                      {{on "click" this.linkPerson}}
+                      >
+                      {{t "connect"}}
+                    </AuButton>
+                    {{/if}}
+                {{/if}}
+              </Group>
+              </AuToolbar>
+          </Auk::Panel::Header>
+          <Auk::Panel::Body class="au-u-padding-none">
```
sergiofenoll: It looks a little better if you leave the default padding in.

sergiofenoll (changes requested):

ValenberghsSven (approved):
Made some changes, mostly prettier and some styling / cleanup
Added a serializer for user to avoid saving the `login-activity` (done in microservice) and `account` (should not be able to be updated or saved in frontend)

app/templates/settings/users/user.hbs
ValenberghsSven: hmm, somewhat disagree. No padding means the borders match up. 
image


Default padding means floating lines

image


But styling without a prototype/design is just exchanging opinions.
This can be fine-tuned later.

sergiofenoll: That's a different view. I'm talking about the view in the user detail. You're showing the organization detail.

No padding:
image

With padding:
image

ValenberghsSven:
Merged this before jenkins was done.
42 of 58 already run with 0 failed tests.
The other tests are mainly publication/search.
Want to cut this process short to get a dev yggie rebuild/reindex going with other merged PR before the weekend.

PR #1705, opened 2023-04-11
KAS-3992: Add shortlist for publication-flows

sergiofenoll:
JIRA: https://kanselarij.atlassian.net/browse/KAS-3992

Adds a new route to the publications page where users can find a list of pieces that need to have their publications started.
image

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/390

Related service:
- [x] https://github.com/kanselarij-vlaanderen/shortlist-service


app/controllers/publications/overview/shortlist.js
```diff
@@ -0,0 +1,127 @@
...
+    for (let maybeAgendaitem of agendaitems) {
+      const agenda = await maybeAgendaitem.agenda;
+      const nextVersion = await agenda.nextVersion;
+      if (!nextVersion) {
+        agendaitem = maybeAgendaitem;
+        break;
+      }
+    }
+    return agendaitem;
+  }
```
ValenberghsSven, line 31: Prefer a query for this one.
We are loading a lot of data that we eventually don't need.

```
await this.store.queryOne('agendaitem', {
      'filter[pieces][:id:]': this.args.piece.id,
      'filter[:has-no:next-version]': 't',
      sort: '-created',
```
This will yield the latest agendaitem linked to that piece.

app/controllers/publications/overview/shortlist.js
```diff
@@ -0,0 +1,127 @@
...
+          'regulation-type',
+          CONSTANTS.REGULATION_TYPES.BVR
+        );
+        break;
+      case CONSTANTS.DOCUMENT_TYPES.MB:
+        regulationType = await this.store.findRecordByUri(
+          'regulation-type',
+          CONSTANTS.REGULATION_TYPES.MB
+        );
+        break;
```
ValenberghsSven: MB doc type should not exist here? 
Shortlist service only returns Decreet and BVR
If any other doc type is returned, does that mean we should abort creating the publication instead of setting it undefined?

app/routes/publications/overview/shortlist.js
```diff
@@ -0,0 +1,33 @@
...
+          'agendaitems.agenda.next-version',
+          'agendaitems.mandatees.person',
+          'agendaitems.treatment.decision-activity',
+          'document-container.type',
+        ].join(','),
+        sort: '-created',
+        'page[size]': result.data.length,
+        'filter[:id:]': result.data.map((record) => record.id).join(','),
+      });
+    }
```
ValenberghsSven, line 29: Potentially loads too much data (all mandatees for all agendaitem versions linked to 1 piece)
But we only need the latest agendaitem.
Perhaps this one should be simpler with just including `document-container.type`.
We could then just rely on the getter (waiting in the template) or load some data in the afterModel instead

app/controllers/publications/overview/shortlist.js
```diff
@@ -0,0 +1,127 @@
...
+        publicationProperties,
+        {
+          case: this.caseForPublication,
+          decisionActivity: this.decisionActivityForPublication,
+          mandatees: this.mandateesForPublication,
+          regulationType: regulationType,
+        }
+      );
+    this.pieceForPublication.publicationFlow = publicationFlow;
+    await this.pieceForPublication.save();
```
ValenberghsSven, line 91: Saving the piece here.
We do this for publications I see (this is also in older code).
This could lead to concurrency issues.
Perhaps user 1 is starting a pub-flow while user 2 is updating the access right to public.
Also, with the default save override we also change `modified` and `accessLevelLastModified` properties by doing this.
Could we just create `pub-flow` with a list (of 1) instead so we set the relation on the inverse?

app/controllers/publications/overview/shortlist.js
```diff
@@ -0,0 +1,127 @@
...
+  @service publicationService;
+  @service router;
+  @service store;
+
+  @tracked isOpenNewPublicationModal;
+  pieceForPublication;
+  agendaitemForPublication;
+  decisionActivityForPublication;
+  caseForPublication;
+  mandateesForPublication;
```
ValenberghsSven, line 17: should these be tracked? we are updating these constantly.
Or not needed if we don't use them in the template?

ValenberghsSven (changes requested):

app/controllers/publications/overview/shortlist.js
sergiofenoll: Because all the data is included, the only calls we actually do is getting `agendaitem.nextVersion` in the case that the nextVersion is null (and we could probably even work around this and have 0 calls instead).

Using a query means that we're doing this call over and over again, because `store.query(One)` isn't cached.
I'm okay with re-writing this but this will turn this route into one of many routes where we're making 10s or 100s of calls just to show a list.

app/controllers/publications/overview/shortlist.js
sergiofenoll: I copy-pasted the logic for creating publications, this case can just be removed. Will do.

app/routes/publications/overview/shortlist.js
sergiofenoll: Same remark as above. We can remove the includes (but with new mu-cl-resources this call is relatively fast), but that also means we're going to be making a lot of calls when rendering. Best to check if this call is fast enough first (it is, on my machine).

app/controllers/publications/overview/shortlist.js
sergiofenoll: Many-to-one relations aren't serialized by default by Ember Data, so we'd have to adapt the serializer for that to work (which is not a problem). But I don't see how that resolves the issue?

If I save a publication-flow with one linked piece and then another user saves that same piece (which has stale data, i.e. does not have the relationship set to the publication-flow), we still have the same problem.

app/controllers/publications/overview/shortlist.js
sergiofenoll: Only used in JS so tracking not necessary as far as I understand.

app/controllers/publications/overview/shortlist.js
ValenberghsSven: Both have potential risk.
Sometimes there are pieces that have lost the connection to their file (after a replace).
This could be one of the causes with concurrency.
Perhaps we should consider using `ModelWithModifier` as a concurrency check for editing pieces and act accordingly
Also, `modified` and `accessLevelLastModified` always being saved might need some change.
Out of scope here.

ValenberghsSven (approved):
made a minor change (route was not showing "op-te-starten" in address)
and error (page not opening) if service was not running.

PR #1708, opened 2023-04-14
KAS-3997: Rework signatures list

sergiofenoll:
Brings the signature list inline with the prototype on usertest.

This expands the `KDataTable` component slightly by adding the option to show an aside (like a sidebar).


app/styles/prototype/_signatures.scss
sergiofenoll, line 1: @brenner-company This prototype styling is taken in part from usertest. I tried to take only what was needed to render what I added in this PR (instead of the whole file), not sure if that is easier or harder for you to deal with when integrating it.

As an aside, maybe the sidebar can become a component in Kaleidos instead of having to use the CSS classes to define it? Not sure if that'd make it easier in the long run. Same for the scroll wrapper, perhaps that can be a component too, we use it a bunch of times and it may be easier to reuse if the structure is already laid out in a component.

brenner-company: Not sure if it is easier or harder either, guess I'll find out 😉

Anyway, I'll have to give the `.auk-sidebar` styling a closer look because it is an existing namespace within `au-kaleidos-css` and I want to make sure it doesn't have any side effects at the moment.

I'll try and pick this up before the end of this sprint.

sergiofenoll:
Closing because this PR is superseded by https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1714

PR #1720, opened 2023-04-24
optimalise highlight in title of searchresult

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3965


app/routes/search/agendaitems.js
```diff
@@ -62,6 +62,10 @@ export default class AgendaitemSearchRoute extends Route {
     if (entry.title && Array.isArray(entry.title)) {
       entry.title = entry.title.join('');
     }
+    if (agendaitem.highlight.shortTitle[0]?.split(" ").length < agendaitem.attributes.shortTitle?.split(" ").length) {
```
tdn: ~~This doesn't do the proper check, as the `agendaitem.highlight.shortTitle[0]` will still include the ems. To be discussed~~

tdn: oh wait, nevermind. It's the length of the split array that's checked here, which will work

tdn: test is likely failing here when `agendaitem.highlight` is `null`, and/or when `agendaitem.highlight.shortTitle` is falsy (undefined or null). 

Pretty sure `agendaitem.highlight.shortTitle` is always an Array when it's not null, but can't hurt to check either.

@StuyckRuben remember to not always go by the happy flow. Always try to break your own code ;-)

tdn (changes requested):

ValenberghsSven (commented):
do we want to check all highlights?
Now we only check the first one.
So not all hits are highlighted.


image

The data: 
image

ValenberghsSven:
Other than the "issue" mentioned above it seems fine.
Searching on other fields (like mandatee) without highlights in titles works (no errors)

app/routes/search/agendaitems.js
```diff
@@ -62,6 +62,10 @@ export default class AgendaitemSearchRoute extends Route {
     if (entry.title && Array.isArray(entry.title)) {
       entry.title = entry.title.join(' ');
     }
+    if ((agendaitem.highlight && agendaitem.highlight.shortTitle) && agendaitem.highlight.shortTitle[0]?.split(" ").length < agendaitem.attributes.shortTitle?.split(" ").length) {
+      const highlightShortTitleEmsRemoved = agendaitem.highlight.shortTitle[0].replace(/<em>|<\/em>/g, "");
+      entry.shortTitle = agendaitem.attributes.shortTitle.replace(highlightShortTitleEmsRemoved, agendaitem.highlight.shortTitle[0]);
```
tdn: change this to `replaceAll` and Sven's comment should be addressed

tdn (changes requested):

app/routes/search/agendaitems.js
tdn: correction: this should be in a for loop, since we're replacing the whole highlight snippet

tdn (approved):

PR #1722, opened 2023-04-25
KAS-3534: Assign mandatees to sign flow

sergiofenoll:
JIRA: https://kanselarij.atlassian.net/browse/KAS-3534

This PR allows you to pick and choose which ministers should be signers on a sign flow. By default the PM and whoever is the submitter on the subcase are chosen but users (permission system to be added later) can pick whomever has to sign.

It's added to both the side panel in the Handtekeningen list as well as the side panel in the Handtekeningen tab of the document preview.

Note that no Handetekenactiviteiten are actually created here, this will happen later when the integration with the backend service occurs.


ValenberghsSven:
already merged by resolving https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1738

PR #1724, opened 2023-04-26
KAS-4022: Add approvers to sign flow

sergiofenoll:
Adds a modal for adding approvers. Approvers are just stored in a flat list, later when we actually create the sign flow we will use the list to create the necessary activities.


ValenberghsSven:
already merged by resolving https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1738

PR #1729, opened 2023-04-28
Replace created date from piece when replacing the file

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4019


ValenberghsSven (commented):
not mentioned in ticket, but there is a replace in document viewer as well.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/documents/document-details-panel.js#L72

tdn:
@StuyckRuben see Sven's comment. Can you do this for the replace in the document-details-panel as well please?

app/components/documents/document-details-panel.js
```diff
@@ -82,6 +82,8 @@ export default class DocumentsDocumentDetailsPanel extends Component {
       }
       this.args.onChangeFile();
     }
+    const now = new Date();
+    this.args.piece.created = now;
```
tdn: @StuyckRuben I think these 2 lines belong inside the `if (this.replacementSourceFile) `, before the `yield this.args.piece.save();` on line 73.

Likely this is also the cause of the failing test

tdn (changes requested):

ValenberghsSven (approved):

PR #1730, opened 2023-04-28
when procedurestap is made confidential all the beslissingsfiche and …

StuyckRuben:
…their previous versions are now made confidential too
https://kanselarij.atlassian.net/browse/KAS-3993


ValenberghsSven (commented):
Personally I would add an await before `piece.save()` since we are no longer returning the save promise (sync saving in ember returns a promise) (prior to this, the await would be covered by `await Promise.all`).
But it could also be totally fine as-is without. Unsure without testing.

We only enter the IF condition if our current piece is not yet confidential.
So in the situation with 2 versions of decision report on a **normal** subcase with a first version "intern regering", BIS version manually set to "vertrouwelijk" > changing subcase to "beperkte toegang" should not work. Query should return only the BIS piece, which is confidential, so the first version is never updated because we don't enter the IF?
Could you test that for me @StuyckRuben?

ValenberghsSven (approved):

ValenberghsSven:
wait for tests

PR #1733, opened 2023-05-03
change prefLabel to altLabel on document types

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3976

### DIFF
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -19,7 +19,7 @@
             <div class="auk-o-flex auk-o-flex--vertical-center auk-o-flex--spaced">
               <span data-test-document-card-type class="auk-overline">
                 {{#unless this.loadPieceRelatedData.isRunning}}
-                  {{this.documentContainer.type.label}}
+                  {{this.documentContainer.type.altLabel}}
                 {{/unless}}
               </span>
             </div>
--- a/app/components/documents/document-details-panel.hbs
+++ b/app/components/documents/document-details-panel.hbs
@@ -145,7 +145,7 @@
           data-test-details-tab-document-type
           @key={{t "document-type"}}
         >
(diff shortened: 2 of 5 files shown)


ValenberghsSven (commented):
Are these all the labels? (there are more in our code than mentioned in the ticket)
Ex: "reeds gekoppelde documenten" doc card, the type pills.
I guess we want to keep some as-is.
Some alt labels are quite long. ex: "BA" = "Begrotingsakkoord van de minister van begroting"
Any css issues because of it? like overlapping text or text not wrapping correctly..
note: some of these comments are out of scope, just wanted to voice them somewhere.

kobemertens:
Looks right according to the ticket. It's a little unclear whether this change also needs to be applied here:
image. @tdn could you confirm that `altLabel` is not needed here?

ValenberghsSven:
> Looks right according to the ticket. It's a little unclear whether this change also needs to be applied here: image. @tdn could you confirm that `altLabel` is not needed here?

The radio buttons with `altLabels` doesn't seem reasonable.
It would increase the size of this component and uploading many files would require more scrolling.
The dropdown list is already using the `altlabel` it seems (MB > Ministerieel Besluit).
I also thought the altlabels were mainly for people who may not know all the abbreviations (viewers). Users who upload documents should know.

tdn:
Can confirm that altLabels are not needed for the radio buttons

tdn (approved):
Added two more instances where we'd want the altLabel to remain consistent.

Can be merged now I think

tdn:
About this:

> Ex: "reeds gekoppelde documenten" doc card, the type pills.

 "reeds gekoppelde documenten" made sense to change as well. Isn't in the ticket, but that's merely an oversight on my end.

 type pills can remain the same.

Also changed it in the document card for publications

PR #1734, opened 2023-05-03
KAS-4041: Verwijderen van dubbele @loading + @disabled setup binnen AuButton instanties

brenner-company:
**Tickets binnen Jira:** 
- https://kanselarij.atlassian.net/browse/KAS-4041


app/components/agenda/agendaitem/create-agendaitem.js
```diff
@@ -51,10 +51,7 @@ export default class CreateAgendaitem extends Component {
   }
 
   get isSaveDisabled() {
-    return (
-      this.selectedSubcases.length === 0 ||
-      this.addSubcasesToAgenda.isRunning
-    );
+    return this.selectedSubcases.length === 0;
```
ValenberghsSven: I get the change (and similar others)
In this case, the template has different entries.
`loading` = `this.addSubcasesToAgenda.isRunning`
`disabled` = `this.isSaveDisabled`
Now we conclude that 1 condition of `isSaveDisabled` is already checked by the loading so we can remove it.
But that created a dependency on the loading parameter.
A future editor may not realize this and remove the `@loading` state.
Then our getter is broken and double clicking during the active task becomes possible.
I am in favor of good save `disabled` conditions (getter or template) and not depend on the `loading` condition.
@tdn thoughts?


In the template it could be good to avoid duplication  if `loading` and `disabled` are the same.
But I am inclined to again keep both for clarity as well. (which in total would mean most this PR was not needed?)

brenner-company: I understand your remark. We'll just wait for @tdn for additional feedback 👍

(I have no issue with discarding any work if it makes more sense, but I have to admit I'll be a bit sad for the lost work 😉)

tdn: Hmmm. though one. I'm inclined to agree with @ValenberghsSven on this one, since it's better to be safe than sorry, and would keep both for the components where the conditions are not the same.

Does discarding it here mean this whole PR was for nothing @brenner-company ?

Perhaps we can resolve this by including tests that make sure double-clicks can't happen? Or is that a stupid idea since there's way too many?

brenner-company: > Hmmm. though one. I'm inclined to agree with @ValenberghsSven on this one, since it's better to be safe than sorry, and would keep both for the components where the conditions are not the same.

✅

> Does discarding it here mean this whole PR was for nothing @brenner-company ?

@tdn I think I can keep around 30-40% of the changes and revert the rest of them. It's a bit of effort wasted, but hey, at least I/we learnt for it 😉

ValenberghsSven (approved):
ok if no tests fail on buttons

brenner-company:
@ValenberghsSven You didn't have to do the revert yourself. I had this planned for the day after tomorrow 😅 Thanks for the help!

ValenberghsSven:
> @ValenberghsSven You didn't have to do the revert yourself. I had this planned for the day after tomorrow sweat_smile Thanks for the help!

Trying to close tickets, was not much work. Only flaky tests it seems so merging this

PR #1736, opened 2023-05-07
add (temporary) default extension for documents

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4026


app/components/documents/document-card.hbs
```diff
@@ -32,9 +32,7 @@
               {{#if this.loadFiles.isRunning}}
                 <AuLoader @padding="small"/>
               {{else}}
-                {{#let (or this.piece.file.derived.extension this.piece.file.extension) as |extension|}}
-                  {{this.piece.name}}.{{extension}}
-                {{/let}}
+                {{this.piece.name}}.{{this.extension}}
```
tdn: Does the following not work?

```
{{#let (or this.piece.file.derived.extension (or this.piece.file.extension this.extension)) as |extension|}}
  {{this.piece.name}}.{{extension}}
{{/let}}
```
If this works, then `extension` doesn't need to be `@tracked` in the js, and all the `loadExtension` logic can be avoided.

tdn (changes requested):

app/components/documents/document-card.hbs
ValenberghsSven: Pretty sure you can put everything in 1 `OR` statement = `(or X Y Z )`

app/config/constants.js
```diff
@@ -136,4 +136,7 @@ export default {
     USER_PAGEVIEW: 'Pageview (per rol)',
     FIRST_ACTION: 'Eerste actie',
   },
+  DEFAULTS: {
+    FALLBACK_EXTENSION : 'pdf',
+  }
```
ValenberghsSven: In the general scheme of this file this is not a good naming strategy.
`DEFAULTS` is very vague as to what you need defaults for.
Something like this would make more sense.
```
  FILE_EXTENSIONS: {
    FALLBACK : 'pdf',
    PDF: 'pdf',
    DOCX: 'docx',
  }
```

We are currently using hardcoded extensions (or file types) in some places which could be replaced with a constant
- helpers/document-icon
- utils/file-type-pill

ValenberghsSven (changes requested):
Not a fan of showing a possibly wrong file extension.
If we go forward with this, we need to change it everywhere (for viewing only)
It might be easier in that case to make getters in the model `piece`
```
  get fileExtension() {
    return this.file?.extension || CONSTANTS.FILE_EXTENSIONS.FALLBACK;
  }
```

```
  get derivedExtension() {
    return this.file?.derived?.extension || CONSTANTS.FILE_EXTENSIONS.FALLBACK;
  }
```
We could then do something like `(or piece.derivedExtension piece.fileExtension)` only where needed.
- document-card
- document-version-history-item
- linked-document-link
- version-card (preview)

the second getter may need some extra checks because if there is a file but no derived file we return the fallback extension instead of checking the actual file.


The getter is sync, while the async relation may not be loaded yet.
Depending on the view that they are used in, we should verify if those relations are loaded somewhere.
If not, we should await them before accessing the getter

tdn (approved):

tdn:
Changed the way to approach this by not showing the dot when the extension is not available, after discussing this with Sven

ValenberghsSven:
Changed 2 more occurrences, wait to see if any tests fail on it.

ValenberghsSven (approved):

PR #1737, opened 2023-05-08
feat: add minutes to meeting model

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-3959


app/models/meeting.js
```diff
@@ -33,7 +33,11 @@ export default class Meeting extends Model {
     async: true,
   })
   internalDocumentPublicationActivity;
-
+  @belongsTo('piece', {
+    inverse: 'meeting',
```
tdn: I see a possible conflict here with https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/e660ce9a47092a46b264d2d824b0b876c51f09c2/app/models/meeting.js#L44

Would rename this (see related app-kaleidos PR). Suggestion: `isMinutesForMeeting`

tdn (changes requested):

app/models/meeting.js
kobemertens: Good catch, I will make the necessary changes

tdn (approved):
Looks like there are some unrelated changes from another branch in here (agenda-detail and agendaitem-controls.js). Be sure to merge those first and maybe check this one again so nothing gets reverted. Maybe rebase this first?

Changes specific to this PR look good

app/models/minutes.js
```diff
@@ -0,0 +1,8 @@
+import Piece from './piece';
+import { attr, belongsTo } from '@ember-data/model';
+
+export default class Report extends Piece {
```
ValenberghsSven: `class Minutes`

app/models/meeting.js
```diff
@@ -33,6 +33,11 @@ export default class Meeting extends Model {
     async: true,
   })
   internalDocumentPublicationActivity;
+  @belongsTo('minutes', {
+    inverse: 'isMinutesForMeeting',
```
ValenberghsSven: A property/relation starting with 'is', 'has', 'can', etc sounds like a boolean.
perhaps just `minutesForMeeting` ?

ValenberghsSven (changes requested):

app/models/meeting.js
ValenberghsSven: @tdn thoughts? Goes against your suggestion

kobemertens: Good point, I will change it

app/models/minutes.js
kobemertens: This was fixed in another branch, will fix here as well

kobemertens: fixed in [f8da6c6](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1737/commits/f8da6c60a6961df6793899f8ab321b7dcc6b6d4b)

app/models/meeting.js
kobemertens: fixed in [38c8e0c](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1737/commits/38c8e0cf6bac0960870e7ea6ff6fe9985bd7efa4)

ValenberghsSven (approved):

PR #1742, opened 2023-05-09
add parameter to choose which doctypes are returned when making archive

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4027

- [ ] merge https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/9


app/utils/zip-agenda-files.js
```diff
@@ -21,11 +21,12 @@ async function constructArchiveName(agenda) {
   return `VR_zitting_${formattedDate}_${agendaName}_alle_punten.zip`;
 }
 
-async function fetchArchivingJob(agenda, mandateeIds, decisions= false) {
+async function fetchArchivingJob(agenda, mandateeIds, decisions= false, extensions = ["pdf"]) {
   let url = `/agendas/${agenda.id}/agendaitems/pieces/files/archive?decisions=${decisions}`;
   if (mandateeIds.length) {
     url += '&' + (new URLSearchParams({ mandateeIds }).toString());
   }
+  url += '&' + (new URLSearchParams({ extensions }).toString());
```
ValenberghsSven: I wouldn't put a default here but in the service instead so the extensions become optional and we fallback on pdf only (the requested default)
If users want other files (like docx) they need to specify it in the frontend

app/utils/zip-agenda-files.js
```diff
@@ -39,7 +40,8 @@ async function fetchArchivingJob(agenda, mandateeIds, decisions= false) {
 }
 
 async function fetchArchivingJobForAgenda(agenda, mandateeIds, decisions, store) {
-  const job = await fetchArchivingJob(agenda, mandateeIds, decisions);
+  const extensions = ["pdf"]; // TODO get this value from control provided by design
```
ValenberghsSven: Does this TODO make this PR unmergable? Not sure of the design status.
We could default to pdf in this PR and improve frontend in a different PR to keep this moving.

ValenberghsSven (changes requested):

app/utils/zip-agenda-files.js
tdn: Design is available. Tasking @StuyckRuben with it today :-)

ValenberghsSven (approved):
used fieldset (like the prototype) and added it to decisions modal

PR #1744, opened 2023-05-12
add pill on publications if there is a signing flow

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-3920

### DIFF
--- a/app/components/publications/documents/document-card-step.hbs
+++ b/app/components/publications/documents/document-card-step.hbs
@@ -45,6 +45,18 @@
               {{/if}}
             </Group.Item>
           </Toolbar.Group>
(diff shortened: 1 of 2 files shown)


ValenberghsSven (changes requested):
Best to wrap the new toolbar group in a feature flag.
`{{#if this.signaturesEnabled}}`
Do we want the move this check higher ` {{#if this.signMarkingActivity}}`? (also around the toolbar group). We just add an empty toolbar group if there is no data

`signaturesEnabled` = flag + permission to "manage-signatures" (for now anyway, ovrb should only be able to read based on their graph restrictions, "read-signatures" may be more appropriate (out of scope here)

ValenberghsSven (approved):

PR #1745, opened 2023-05-12
Feature/kas 4091 default co signing ministers

tdn:
Confirmed with design, all mandatees that are on the subcase need to be in the defaults for the sign flow.
The order is important as well:
1. Prime minister
2. Submitting minister
3. Other ministers on the subcase


tdn:
Superseded by https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1747/files

PR #1746, opened 2023-05-12
Feature/kas 4046 sign flow profile restrictions

ValenberghsSven:
Combines 2 tickets:
https://kanselarij.atlassian.net/browse/KAS-4046
https://kanselarij.atlassian.net/browse/KAS-4049
 
The first ticket mentions making it not possible for certain profiles to start signature flow.
In the signature route, it would make more sense to set the filtered ministers in favor of showing wrong ones but blocking the buttons
That filtering is the second ticket so made sense to keep them together.

## changes:

document-card:
- wrapped the download link of `signedPiece` with feature flag and general permissions.
- wrapped the download link with new permission for dossierbeheerder to only show the link if the `subcase.requestedBy` matches the organization mandatees (async action getter into service call)

(document-preview)  => document-details-panel
- wrapped the download link of `signedPiece` with feature flag and general permissions.
- wrapped the download link with new permission for dossierbeheerder to only show the link if the `subcase.requestedBy` matches the organization mandatees (async action getter into service call)

(document-preview)  => signatures-tab (already hidden by feature flag and permissions)
- wrapped the creation of sign-flow with new dossierbeheerder to only show them if the `subcase.requestedBy` matches the organization mandatees (async action getter into service call)
- show a warning message if viewing a document that does not match with mandatees.

permissions.js
- added new permission to view and manage only specific signature flows

signatures route/controller/template:
- hidden ministerfilter
- added the new permissions to filter out the results from the shortlist service to only get the `pieces`  where the `subcase.requestedBy` matches the organization mandatees (or nothing if they have no minister assigned)
- in that case, we don't use the local storage key.

nl-be.json
- added translation for the warning

### DIFF
--- a/app/routes/signatures/index.js
+++ b/app/routes/signatures/index.js
@@ -4,15 +4,23 @@ import { inject as service } from '@ember/service';
 
 export default class SignaturesIndexRoute extends Route {
   @service store;
+  @service currentSession;
 
   localStorageKey = 'signatures.shortlist.minister-filter';
 
   ministerIds = [];
 
-  beforeModel() {
-    this.ministerIds = JSON.parse(
-      localStorage.getItem(this.localStorageKey)
-    ) ?? [];
+  async beforeModel() {
+    if (this.currentSession.may('manage-only-specific-signatures')) {
+      const currentUserOrganization = await this.currentSession.organization;
+      const currentUserOrganizationMandatees = await currentUserOrganization.mandatees;
+      const currentUserOrganizationMandateesIds = currentUserOrganizationMandatees?.map((mandatee) => mandatee.id);
+      this.ministerIds = currentUserOrganizationMandateesIds;
+    } else {
+      this.ministerIds = JSON.parse(
+        localStorage.getItem(this.localStorageKey)
+      ) ?? [];
+    }
   }
 
   async model() {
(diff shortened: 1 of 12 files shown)


app/routes/signatures/index.js
```diff
@@ -4,15 +4,23 @@ import { inject as service } from '@ember/service';
...
-  beforeModel() {
-    this.ministerIds = JSON.parse(
-      localStorage.getItem(this.localStorageKey)
-    ) ?? [];
+  async beforeModel() {
+    if (this.currentSession.may('manage-only-specific-signatures')) {
+      const currentUserOrganization = await this.currentSession.organization;
+      const currentUserOrganizationMandatees = await currentUserOrganization.mandatees;
+      const currentUserOrganizationMandateesIds = currentUserOrganizationMandatees?.map((mandatee) => mandatee.id);
+      this.ministerIds = currentUserOrganizationMandateesIds;
```
tdn, line 18: What happens here if no ministers are set?

Default = no filter and the user sees everything?
Or no filter and the user sees nothing? (which it should be in this case I guess)

translations/nl-be.json
```diff
@@ -1208,6 +1208,7 @@
   "start-signing": "Tekenen opstarten",
   "cannot-sign": "Kan niet ondertekenen",
   "cannot-sign-message": "Dit stuk is niet verbonden aan een agendapunt en kan niet aangeboden worden om getekend te worden.",
+  "cannot-create-sign-flow-message": "Dit stuk is niet verbonden aan uw organisatie en kan niet aangeboden worden om getekend te worden.",
```
tdn: Heb dit tekstje licht aangepast naar "Dit stuk is niet ingediend door uw mandataris en kan niet aangeboden worden om getekend te worden."

tdn (approved):

PR #1750, opened 2023-05-16
Bug/kas 4088 lange emails

kobemertens:
Fixes https://kanselarij.atlassian.net/browse/KAS-4088

### DIFF
--- a/app/components/signatures/create-sign-flow.js
+++ b/app/components/signatures/create-sign-flow.js
@@ -8,6 +8,8 @@ import { TrackedArray } from 'tracked-built-ins';
 import { startOfDay } from 'date-fns';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 
+const MAX_EMAIL_DISPLAY_LENGTH = 65;
+
 /**
  */
 export default class SignaturesCreateSignFlowComponent extends Component {
@@ -141,6 +143,12 @@ export default class SignaturesCreateSignFlowComponent extends Component {
       mandatees.map((id) => this.store.findRecord('mandatee', id))
     );
   }
+
+  ellipsify = (word) => {
+    return word.length > MAX_EMAIL_DISPLAY_LENGTH
+      ? word.slice(0, MAX_EMAIL_DISPLAY_LENGTH - 1) + '…'
+      : word;
+  };
 
   /**
    * Filter the selected ministers for signing so that it only contains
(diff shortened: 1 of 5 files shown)


kobemertens:
@brenner-company, ik heb in [03ed5aa](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1750/commits/03ed5aa5c41c6de9f3676f49272faddec9f88e79) code toegevoegd om email adressen die niet in de textbox passen af te kappen met 3 puntjes. Dit is momenteel hardcoded op 70 characters, maar het zou beter zijn moest dit via css gebeuren (iets als `text-overflow: ellipsis`). Niet super belangrijk maar als dat voor u niet veel werk is zou dat beter zijn.

sergiofenoll (approved):
As you say a CSS fix would be better, but this is okay.

brenner-company:
Check! Issue was even zoeken, maar had te maken met de manier waarop default grids in elkaar zitten (specifieker: 'grid blowout' - https://css-tricks.com/preventing-a-grid-blowout/). Controleer je nog even kort of alles daar in orde voor is, @SchrodingersCat00?

app/components/signatures/create-sign-flow.js
```diff
@@ -8,6 +8,8 @@ import { TrackedArray } from 'tracked-built-ins';
 import { startOfDay } from 'date-fns';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 
+const MAX_EMAIL_DISPLAY_LENGTH = 55;
+
```
ValenberghsSven: unused constant. (if we would use it, I would argue putting it in the `constants.js` or `config.js`

ValenberghsSven (approved):
works, I will remove the unused constant before merging.
plz note. This is a linter error that should not have been committed.
Jenkins fails on it
image

This should have been prevented by using the pre-commit hook

ValenberghsSven (commented):
I approved this, then I realized we have a <AbbreviatedText> component.
Used here (with a maxlength of 35) we get this.

image

Hovering over the icon shows the entire email adress.
The small chars vs the wide chars do yield a visual difference.
35 length is a bit too much for the wide, the popup from the icon is out of frame a bit


@tdn Thoughts? The current fix is ok for solving the mentioned bug.

There is a hard char limit from the input mask validation still. Do we override that? sensible limit for emails needed.
Also, the mask is very strict (no capitals allowed for validation)

PR #1752, opened 2023-05-17
Feature/kas 4102 tweak signatures tab

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4102

Updates the layout of the signatures overview

### DIFF
--- a/app/controllers/signatures/ongoing.js
+++ b/app/controllers/signatures/ongoing.js
@@ -1,6 +1,8 @@
 import Controller from '@ember/controller';
 import { tracked } from '@glimmer/tracking';
+import fetch from 'fetch';
 import { PAGINATION_SIZES } from 'frontend-kaleidos/config/config';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
 
 export default class SignaturesOngoingController extends Controller {
   queryParams = [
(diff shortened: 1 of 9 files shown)


app/templates/signatures/ongoing.hbs
```diff
@@ -20,22 +20,30 @@
...
       @onChange={{fn (mut this.sort)}}
     />
     <th>{{t "status"}}</th>
+    <th></th>
   </:header>
   <:body as |signFlow|>
     {{#let signFlow.signSubcase.signMarkingActivity as |signMarkingActivity|}}
       {{#let signMarkingActivity.piece as |piece|}}
-        <td>
+        <td class="au-u-flex au-u-flex--spaced-tiny">
```
ValenberghsSven: the flex here messes with the rows for me.
image
Longer titles mean a large column size, but lots of whitespace
image

app/templates/signatures/ongoing.hbs
```diff
@@ -59,6 +76,29 @@
...
+              {{t "open"}}
+            </AuLinkExternal>
+          {{else}}
+            <AuButton
+              @skin="naked"
+              @icon="chevron-right"
+              @hideText={{true}}
+              @disabled={{true}}
+            >
+              {{t "open"}}
```
ValenberghsSven: Is this a placeholder?
Perhaps a different button icon & tooltip to indicate the link to signinghub is not working (maybe you have no rights or it got removed from signinghub)

app/controllers/signatures/ongoing.js
```diff
@@ -20,4 +22,41 @@ export default class SignaturesOngoingController extends Controller {
...
+    const signSigningActivities = await signSubcase.signSigningActivities;
+    const mandatees = await Promise.all(
+      signSigningActivities.map((activity) => activity.mandatee)
+    );
+    const persons = await Promise.all(
+      mandatees
+        .sort((m1, m2) => m1.priority - m2.priority)
+        .map((mandatee) => mandatee.person)
+    );
+    return persons.map((person) => person.fullName);
```
ValenberghsSven, line 44: I think this is right? For agenda it makes sense to show them based on priority.
For signatures, is the `requestBy` more important or not? Leave as-is for now.

ValenberghsSven (changes requested):
Small points:

1. ongoing.hbs still has `{{page-title (t "ongoing")}}`, needs to be "follow-up"?
2. The name in router.js is still "wordt-verwerkt". Needs changing?
3. the index route has no name. Does that need to be "handtekeningen/op-te-starten" or not? It made sense not to give it a name when there was only 1 subroute.
4. Button that does nothing (but looks the same like the normal one)

app/templates/signatures/ongoing.hbs
ValenberghsSven: with pill
image

app/templates/signatures/ongoing.hbs
ValenberghsSven: I changed the icon to a "broken-link".

app/templates/signatures/ongoing.hbs
ValenberghsSven: removed the flex from the column, looks better

ValenberghsSven:
> Small points:
> 
>     1. ongoing.hbs still has `{{page-title (t "ongoing")}}`, needs to be "follow-up"?
> 
>     2. The name in router.js is still "wordt-verwerkt". Needs changing?
> 
>     3. the index route has no name. Does that need to be "handtekeningen/op-te-starten" or not? It made sense not to give it a name when there was only 1 subroute.
> 
>     4. Button that does nothing (but looks the same like the normal one)

1. changed it (checked prototype)
2. changed it
3. added dutch name for index to `router.js`
4. changed the icon so it's different from the actual link

ValenberghsSven (approved):
Despite my own minor changes the main work is good

tdn (approved):
Approving this, since we'll have to elaborate on this view in the future anyway

PR #1751, opened 2023-05-16
Feature/kas 2793 remove signed document

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-2793

- Added a new permission to remove signflow related data
- Added a service call to clean up everything related to a sign flow.
- Added dropdown in document-card actions (with permission)
- Added trash button in document details panel in doc viewer (with permission)

## Wait with merging until KAS-4099 is resolved (PR will be based on this branch). Some fixes made there to delete the activities
- [ ] https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1754

PR #1754, opened 2023-05-17
KAS-4099 implement removing sign-flow data from signature tab

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4099
https://kanselarij.atlassian.net/browse/KAS-2793

Original PR: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1751
This branch contains fixes to the signature-service method to completely remove all data models.
- Added a new permission to remove signflow related data
- Added a service call to clean up everything related to a sign flow.
- Added dropdown and action in document details panel in doc viewer (with permission)
- Added verify modal for this destructive action in document details panel 
- Added action with tooltip in signature tab
- Added verify modal for this destructive action in signature tab

Added serializer because we set some `belongsTo` relations inversely (activity knows sign-subcase)
Some data might be added by the digital-signing service.
Saving signSubcase after loading all relations may result in overwriting lists with stale data (no reload before save).

Also added the `hasMany` to serialize.
Even though Ember should not serialize it (saving sign-subcase does not save the list (This used to happen in earlier ember versions)) > deleting sign-subcase after deleting all hasMany activities yields an error in console.
image

This error is not present when reloading the `signsubcase` model before deleting.

### DIFF
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -167,6 +167,17 @@
                   </AuLink>
                   <AuHr />
                 {{/if}}
+                {{#if this.mayRemoveSignFlow}}
+                  <AuButton
+                    @skin="link"
+                    @alert={{true}}
+                    {{on "click" this.deleteSignFlow}}
+                    role="menuitem"
+                  >
+                    {{t "delete-signed-piece"}}
+                  </AuButton>
+                  <AuHr />
+                {{/if}}
                 <AuButton
                   data-test-document-card-delete
                   @skin="link"
@@ -256,6 +267,17 @@
   @message={{t "delete-document-container-message"}}
   @onConfirm={{this.verifyDeleteDocumentContainer}}
   @onCancel={{this.cancelDeleteDocumentContainer}}
+  @confirmMessage={{t "delete"}}
+  @confirmIcon="trash"
+  @alert={{true}}
+/>
+
+<ConfirmationModal
+  @modalOpen={{this.isOpenVerifyDeleteSignFlow}}
+  @title={{t "delete-signed-piece"}}
+  @message={{t "verify-delete-signed-piece"}}
+  @onConfirm={{this.verifyDeleteSignFlow}}
+  @onCancel={{this.cancelDeleteSignFlow}}
   @confirmMessage={{t "delete"}}
   @confirmIcon="trash"
   @alert={{true}}
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -167,17 +167,6 @@
                   </AuLink>
                   <AuHr />
                 {{/if}}
-                {{#if this.mayRemoveSignFlow}}
-                  <AuButton
-                    @skin="link"
-                    @alert={{true}}
-                    {{on "click" this.deleteSignFlow}}
-                    role="menuitem"
-                  >
-                    {{t "delete-signed-piece"}}
-                  </AuButton>
-                  <AuHr />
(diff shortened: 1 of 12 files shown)


ValenberghsSven:
The modals are staying up during the process, allowing double clicks.
Need to prevent those...

app/components/web-components/vl-modal-verify.hbs
```diff
@@ -6,7 +6,7 @@
   {{! template-lint-disable no-down-event-binding}}
   {{on "keydown" this.keyDown}}
 >
-  <div class="auk-modal auk-modal--small" role="dialog">
+  <div class="auk-modal auk-modal--medium" role="dialog">
```
ValenberghsSven, line 9: We only use this modal in 3 places (coming from one). Small is not big enough.

app/components/documents/document-details-panel.hbs
```diff
@@ -179,23 +179,39 @@
...
+                @iconAlignment="left"
+                download
+              >
+                {{@piece.signedPiece.name}}.{{@piece.signedPiece.file.extension}}
+              </AuLinkExternal>
+              <AccessLevelPill 
+                @accessLevel={{await @piece.signedPiece.accessLevel}} 
+                @isEditable={{false}}
+              />
+              <AuDropdown
```
tdn: I think a `{{#if (user-may "remove-signatures")}}` would be appropriate here. Now this is only checked by `this.currentSession.may('manage-signatures');` in `isSignaturesEnabled`.

If we have a separate permission for removal, might as well use it consistently.

app/components/documents/document-card.hbs
```diff
@@ -259,4 +259,4 @@
   @confirmMessage={{t "delete"}}
   @confirmIcon="trash"
   @alert={{true}}
-/>
+/>
```
tdn, line 262: Just a sidenote, as this wasn't specified in the original issue. When we delete a document container now without first deleting the signed piece, the signed document still lives in the db and on the system (it's now orphaned).

We have a follow-up issue for that (KAS-2794), but I'm wondering if we should disable the removal of pieces that have signed pieces for now. That would force the user to first delete the signed piece, and then the original.

tdn (changes requested):
Two small points:
1. Use `may remove-signatures` for the action next to the pill on the 'Handtekeningen' tab
2. Disable the container removal action for original pieces with a signed piece to avoid keeping orphaned signed pieces in the db.

app/components/documents/document-card.hbs
ValenberghsSven: This is not the only place where you can remove document containers. (batch edit will also delete&cleanup / document-card actions..) 
I would argue we fix this separately

ValenberghsSven:
> Two small points:
> 
>     1. Use `may remove-signatures` for the action next to the pill on the 'Handtekeningen' tab
> 
>     2. Disable the container removal action for original pieces with a signed piece to avoid keeping orphaned signed pieces in the db.

1. I think you meant document-details-panel
2. out of scope? this is not the only place where we can delete pieces/containers. Best to dedicate it's own ticket.

ValenberghsSven:
I'm also of the opinion that the use-case where a user would delete an entire document after it has been signed would be user error or emergency action. Should not happen to often.
Do we want to disable this in total? Admin should be able to do it.
Maybe it is better to include the cleanup in the deletion process regardless and add additional warnings/information in the verify popup?

tdn (approved):

PR #1755, opened 2023-05-17
KAS-4093: Toasts for sign flows

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4093

Adds toasts on sign flow creation success/error

### DIFF
--- a/app/controllers/signatures/index.js
+++ b/app/controllers/signatures/index.js
@@ -7,9 +7,11 @@ import { trackedFunction } from 'ember-resources/util/function';
 import { TrackedArray } from 'tracked-built-ins';
 
 export default class SignaturesIndexController extends Controller {
+  @service intl;
   @service router;
   @service store;
   @service signatureService;
+  @service toaster;
 
   @tracked piece = null;
   @tracked decisionActivity = null;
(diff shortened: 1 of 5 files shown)


app/styles/auk/_auk-additions.scss
```diff
@@ -400,3 +400,7 @@ table tr.lt-expanded-row {
     text-transform: none;
   };
 }
+
+.auk-alert-stack {
+  z-index: 10000;
+}
```
ValenberghsSven: the alert stack should already by quite high up the z-index scale.
https://github.com/kanselarij-vlaanderen/au-kaleidos-css/blob/main/_s-z-index.scss#L12

sergiofenoll: It's because AuModals are set to z-index 9999, so the alerts don't show in the document preview modal. None of the alerts did, so any errors were invisible, but now it's quite obvious because (with z-index 1700) the toasts about sign flow creation where hidden until the user clicked off the modal.

app/controllers/signatures/index.js
```diff
@@ -200,27 +202,47 @@ export default class SignaturesIndexController extends Controller {
...
+        );
+      } else {
+        this.toaster.success(
+          this.intl.t('document-was-sent-to-signinghub'),
+          this.intl.t('successfully-started-sign-flow')
+        );
+      }
+    } catch {
+      this.closeSidebar();
+      await this.router.refresh(this.router.routeName);
```
ValenberghsSven, line 241: closing and refreshing executed on fail and success > `finally` block ?

app/services/signature-service.js
```diff
@@ -54,7 +54,10 @@ export default class SignatureService extends Service {
     await signSubcase.save();
 
     // Prepare sign flow: create preparation activity and send to SH
-    await uploadPiecesToSigninghub(signFlow, [piece]);
+    const response = await uploadPiecesToSigninghub(signFlow, [piece]);
+    if (!response.ok) {
+      throw new Error('Failed to upload piece to Signing Hub');
+    }
   }
```
ValenberghsSven: I'm guessing we want to implement the delete feature here later when this fails.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1754.
Right now, sending to github fails but nothing seems to happen (you can start multiple failed processes), after refresh you get the view with the status pill.

Could we do this different? Keep the entire process but add a status ('failed to be sent to signinghub') with a button to retry instead of destroying everything and starting over?

app/styles/auk/_auk-additions.scss
ValenberghsSven: I see, this must be in the ember-appuniversum repo.

app/controllers/signatures/index.js
sergiofenoll: Perhaps, but the problem is that the toasts need to be created after we refresh, otherwise we show them briefly and they get lost during the refresh.

If you set it in a `finally` block, you also need to show the toast in the finally block, and you need to track the state for that (error or success, many or one).

sergiofenoll: Perhaps, but the problem is that the toasts need to be created after we refresh, otherwise we show them briefly and they get lost during the refresh.

If you set it in a `finally` block, you also need to create the toast in the finally block, and you need to track the state for that (error or success, many or one).

ValenberghsSven: keep as-is

app/services/signature-service.js
ValenberghsSven: [KAS-4107](https://kanselarij.atlassian.net/browse/KAS-4107)

ValenberghsSven: Will be implementing the delete feature that was just merged

ValenberghsSven: added delete when failed

ValenberghsSven (approved):
made small changes, code was ok

PR #1756, opened 2023-05-22
Feature/kas 3918 weergave notulen

kobemertens:
Needs https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/392 for the backend


ValenberghsSven (approved):
Changed/added a few things:
- feature flag for minutes
- permission for minute editing
- serializer rule to make meeting.minutes read-only


In general:
- There are some hardcoded dutch strings right now.
Ideally we get those from a translation key (even though we most likely only ever use dutch).
But it's mainly so we can reuse existing keys + 1 central location to keep our dutch texts that will be displayed.
- there are parts of the code commented. Since this is a work in progress I guess it's fine but we will have to cleanup eventually
- I was unsure if non-editors were allowed to see minutes so I excluded them for now.

This one is ok to be merged. Feature can be hidden with feature flag and user feedback is needed to proceed.

PR #1757, opened 2023-05-22
add all documents connected to a subcase will now be made confidentia…

StuyckRuben:
…l when subcase is made confidential
https://kanselarij.atlassian.net/browse/KAS-4071

### DIFF
--- a/app/components/subcase/subcase-titles-panel.js
+++ b/app/components/subcase/subcase-titles-panel.js
@@ -8,12 +8,10 @@ import { action } from '@ember/object';
  */
 export default class SubcaseTitlesPanel extends Component {
   @tracked isEditing = false;
-  @tracked startEditingConfidential = false;
 
   @action
   startEditing() {
     this.isEditing = true;
-    this.startEditingConfidential = !!this.args.subcase.confidential;
   }
 
   @action
--- a/app/components/subcase/subcase-titles-panel.js
+++ b/app/components/subcase/subcase-titles-panel.js
@@ -8,10 +8,12 @@ import { action } from '@ember/object';
  */
 export default class SubcaseTitlesPanel extends Component {
   @tracked isEditing = false;
+  @tracked startEditingConfidential = false;
 
   @action
   startEditing() {
     this.isEditing = true;
+    this.startEditingConfidential = !!this.args.subcase.confidential;
   }
 
   @action
--- a/app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
+++ b/app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
@@ -47,6 +47,7 @@ export default class SubcaseTitlesPanelEdit extends Component {
     );
     if (this.args.subcase.confidential) {
       yield this.pieceAccessLevelService.updateDecisionsAccessLevelOfSubcase(this.args.subcase);
+      yield this.pieceAccessLevelService.updateSubmissionAccessLevelOfSubcase(this.args.subcase);
     }
     this.args.onSave();
   }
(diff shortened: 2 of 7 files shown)


app/services/piece-access-level-service.js
```diff
@@ -131,6 +131,29 @@ export default class PieceAccessLevelService extends Service {
...
+
+    await Promise.all(pieces.map(async (piece) => {
+      const accessLevel = await piece.accessLevel;
+      if (accessLevel.uri !== vertrouwelijk.uri) {
+        piece.accessLevel = vertrouwelijk;
+        await piece.save();
+      }
+      await this.updatePreviousAccessLevels(piece);
+    }));
+  }
```
ValenberghsSven: This code does not cover a case where the piece is already `INTERN_SECRETARIE`. Having a `strengthenAccessLevelToConfidential` would be better in that case.
Look at the `strengthenAccessLevelToInternRegering` method

app/services/piece-access-level-service.js
```diff
@@ -131,6 +131,29 @@ export default class PieceAccessLevelService extends Service {
...
+    const submissionActivities = await this.store.query('submission-activity', {
+      'filter[subcase][:id:]': subcase.id,
+      include:'pieces',
+    });
+    const pieces = [];
+    for (const submissionActivity of submissionActivities.toArray()) {
+      let submissionPieces = await submissionActivity.pieces;
+      submissionPieces = submissionPieces.toArray();
+      pieces.push(...submissionPieces);
+    }
```
ValenberghsSven: This list of pieces from submissions will contain previous versions of each other. (and duplicates when subcase has been postponed)
Later on we update previous access levels so some pieces will be processed twice (second time no save).

We might be better off making a set with only the last versions of each and run that through a "strengthen" method (see other comment)

ValenberghsSven (changes requested):

app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
```diff
@@ -47,6 +47,7 @@ export default class SubcaseTitlesPanelEdit extends Component {
     );
     if (this.args.subcase.confidential) {
       yield this.pieceAccessLevelService.updateDecisionsAccessLevelOfSubcase(this.args.subcase);
+      yield this.pieceAccessLevelService.updateSubmissionAccessLevelOfSubcase(this.args.subcase);
     }
```
ValenberghsSven: This happens on every change, not just when setting confidential.
Should we listen to the specific change from non-confidential to confidential?

Changing a title (fix a typo) would update all pieces (maybe some are public and released to valvas). Not sure how plausible that scenario is. 

- personal note: this type of "behind the scenes" logic should be mentioned somewhere to the user (maybe with an option to cancel) > "Met deze actie zullen alle documenten vertrouwelijk worden. Verder gaan?"
Or an info tooltip next to the confidential toggle.

ValenberghsSven: Wat we are basically doing is saving the user the trouble of going to the batch edit window to set all access levels manually

app/components/subcase/subcase-titles-panel.js
```diff
@@ -8,10 +8,12 @@ import { action } from '@ember/object';
  */
 export default class SubcaseTitlesPanel extends Component {
   @tracked isEditing = false;
+  @tracked startEditingConfidential = false;
 
   @action
   startEditing() {
     this.isEditing = true;
+    this.startEditingConfidential = !!this.args.subcase.confidential;
```
ValenberghsSven: I get the idea but we just need to check the toggling in the edit component. Will change this a bit

ValenberghsSven (approved):

PR #1759, opened 2023-05-23
fix correctly changes url when changing document versions in document…

StuyckRuben:
…-preview-modal

add queryparam to show active tab
https://kanselarij.atlassian.net/browse/KAS-4058


app/routes/document.js
```diff
@@ -11,6 +11,9 @@ export default class DocumentRoute extends Route {
     isSigning: {
       refreshModel: true,
       as: 'aanbieden_voor_handtekenen',
+    },
+    activeTab: {
+      refreshModel: false,
     }
   }
```
ValenberghsSven: 1. we show `id??activeTab=versions` (or signatures). We can alias a dutch name for the left hand part.
2. now that we have this, we could replace the `isSigning` usage with `activeTab` with some tweaks to were it was used. Could be considered "out of scope"

ValenberghsSven (commented):
Seems to work from initial (limited) testing

app/controllers/document.js
```diff
@@ -26,6 +26,20 @@ export default class DocumentController extends Controller {
...
   }
 
+  @action
+  tabChanged(tabName){
+    this.activeTab = tabName;
+  }
+
+  @action
+  setSelectedVersion(piece) {
+    this.router.transitionTo(
```
ValenberghsSven: `replaceWith` ipv `transitionTo` to stop browser history additions on tab switch

ValenberghsSven (changes requested):
Tried `replaceWith`, history still seems to get added making closing the viewer a problem.
changing tabs also adds history.
The viewing of document is slowed down significantly with the constant visual refreshes happening (the m-header shows up briefly when switching tabs and selecting a version).
I don't think a `transitionTo` or `replaceWith` are the solution here.
Could we replace the model in route and reset the controller?
Trying out some things myself, check in with me before picking this up

ValenberghsSven (commented):
@tdn
My changes:
- in order to get a dutch name on the right hand side of the QP I had to either put dutch in the templates/js files or translate where needed (choose the latter)
- address url "?activeTab" > "?tab" is just setting an alias in the QP
- to avoid tab switching from causing a refresh i added `replace: true` to QP
- to avoid version switching causing the m-header to pop up briefly I added a override to the loading and prevented the bubbling of it. 
We now show a blank auModal with loaders when model is loading.
Looks better, perhaps not perfect (we could try and show a replacement sidebar while loading but opted not to)
- those 2 + the `replaceWith` means browser history does not update on tab switching or version switching.
- removed `isSigning` QP in favor of tabs
- tweaked rules for when we transition to "tab=Handtekeningen" when not allowed, we default to "details"


notes:
- During writing this, I think the `activeTab` duplication on both controller and component could probably be cleaned up a bit and use `args.activeTab` instead. Could be wrong
- not sure about the translation for the tabs. If there is a way to alias it it might be better.
We could also not have the full name but just single letter "tab=t" en "tab=v"  or some form of mapping from english to dutch

tdn:
> notes:
> 
> * not sure about the translation for the tabs. If there is a way to alias it it might be better.
>   We could also not have the full name but just single letter "tab=t" en "tab=v"  or some form of mapping from english to dutch

Agreed that Dutch names might be cleaner in the URL here.

@StuyckRuben can you change `activeTab` to just `tab` please? Also be sure to check out and review @ValenberghsSven's changes to learn from them. Especially the way the translation keys were used to handle the tab names while keeping the code in English is pretty cool to see. Wouldn't have thought of that myself :-)

After the change from `activeTab` to just `tab` I think this one can be merged as soon as tests pass.

ValenberghsSven (approved):

PR #1761, opened 2023-05-26
disable edit and delete button on documents with signflow

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4105


app/components/documents/batch-details/batch-documents-details-modal.js
```diff
@@ -54,12 +54,16 @@ export default class BatchDocumentsDetailsModal extends Component {
 
     this.rows = yield Promise.all(
       latestDocs.map(async (piece) => {
+        const signMarkingActivity = await piece.signMarkingActivity;
+        const signCompletionActivity = await piece.signCompletionActivity;
+        const signedPiece = await piece.signedPiece;
```
ValenberghsSven: - prefer a feature flag of some sort. flag off > edits enabled
- I get loading everything, but only 1 needs to be true so we could stop loading the data if 1 is true (most likely marking activity) if marking-activity does not exist but the other do 2, that would be a data issue I think.
- perhaps a new util method would be better since we can reuse that in more places. (or service call in the style of `await signatureService.hasSignFlow(piece)` (feature flag could be in the service)

app/components/documents/batch-details/batch-documents-details-modal.js
```diff
@@ -54,12 +54,16 @@ export default class BatchDocumentsDetailsModal extends Component {
...
         const row = new Row();
         row.piece = piece;
         row.name = piece.name;
         row.accessLevel = piece.accessLevel;
         row.documentContainer = await piece.documentContainer;
         row.documentType = row.documentContainer.type;
+        row.canBeEditedOrDeleted = signMarkingActivity || signCompletionActivity || signedPiece ? false : true;
```
ValenberghsSven: - `row.hasSignFlow` Since it is highly dependent on signatures? that would invert the boolean so true means no deletion

app/components/documents/batch-details/document-details-row.hbs
```diff
@@ -7,11 +7,12 @@
       <AuCheckbox
         @checked={{@isSelected}}
         @onChange={{fn @onSelect @row}}
+        @disabled={{not @row.canBeEditedOrDeleted}}
       />
     </td>
   {{/if}}
   <td>
-    {{#if @isEditingEnabled}}
+    {{#if @row.canBeEditedOrDeleted}}
```
ValenberghsSven: purely looking at the incoming parameters I would argue that this replaces a permission "manage-documents" with a check on data (which could mean non-editors can edit/delete documents now when there is no signFlow attached.

Luckily, the 3 places where we use this batch doc edit component are all covered by the permission.
The edit button does not exist for them (look for `this.openBatchDetails`)
But these rows were built with a non-editor viewer in mind I think. (not used as such yet)
So I would check both `@isEditingEnabled` and the extra boolean

app/components/documents/batch-details/document-details-row.hbs
```diff
@@ -69,7 +70,7 @@
         >
           {{t "undo"}}
         </AuButton>
-      {{else}}
+      {{else if @row.canBeEditedOrDeleted}}
         <AuButton
           data-test-document-details-row-delete
```
ValenberghsSven: Not showing the button is fine I guess.
Would we rather show an info icon stating why? not sure how much users case about this.
Could we fall back to the book icon instead? (expand `{#if (not @isEditingEnabled)}}` with the boolean)

app/components/documents/batch-details/document-details-row.hbs
```diff
@@ -7,11 +7,12 @@
       <AuCheckbox
         @checked={{@isSelected}}
         @onChange={{fn @onSelect @row}}
+        @disabled={{not @row.canBeEditedOrDeleted}}
```
ValenberghsSven: This checkbox is disabled, but the top batch editing is still available and will select this disabled checkbox.
image

ValenberghsSven: I think we need to change the "select all" function to only select rows that do not have the new boolean

app/components/documents/document-card.hbs
```diff
@@ -127,6 +127,7 @@
               (and
                 @isEditable
                 (user-may "manage-documents")
+                this.mayEditOrDelete
               )
```
ValenberghsSven: - This hides the entire dropdown button so no need to hide the individual options later on.
- `hasSignFlow` or similar here.

- My opinion (to discuss) is to maybe have an option with info as to why there are no actions available (either an option with info, or replace the dropdown icon with an info icon. Perhaps to discuss with the users.

app/components/documents/document-card.hbs
```diff
@@ -138,6 +139,7 @@
                 @disabled={{this.loadPieceRelatedData.isRunning}}
                 @alignment="right"
               >
+              {{#if this.mayEditOrDelete}}
                 <AuButton
```
ValenberghsSven: hiding the option is not needed as the dropdown itself is hidden
There are 4 actions in this dropdown (3 edit, 1 link to signFlow preview)
when signFlow exists, 3 options are hidden by the boolean and the 4th is mutually exclusive so this would result in a dropdown with no options (which throws errors at the moment)

app/components/documents/document-card.js
```diff
@@ -64,6 +66,15 @@ export default class DocumentsDocumentCardComponent extends Component {
...
 
+  @task
+  *loadPieceMayBeEditedOrDeleted() {
+    const signCompletionActivity = yield this.piece.signCompletionActivity;
+    const signedPiece = yield this.piece.signedPiece;
+    if (!this.signMarkingActivity && !signCompletionActivity && !signedPiece){
+      this.mayEditOrDelete = true;
+    }
+  }
+
```
ValenberghsSven: This could fall under `loadSignatureRelatedData`
Could also be util/service call (loading `signMarkingActivity` is ok, since the second await/yield will be resolved instantly)

app/components/documents/document-details-panel.js
```diff
@@ -45,6 +48,16 @@ export default class DocumentsDocumentDetailsPanel extends Component {
...
 
+  @task
+  *loadPieceMayBeEditedOrDeleted() {
+    const signMarkingActivity = yield this.args.piece.signMarkingActivity;
+    const signCompletionActivity = yield this.args.piece.signCompletionActivity;
+    const signedPiece = yield this.args.piece.signedPiece;
+    if (!signMarkingActivity && !signCompletionActivity && !signedPiece){
+      this.mayEditOrDelete = true;
+    }
+  }
```
ValenberghsSven: signature related data so `loadSignatureRelatedData` like other places (util/service call)

ValenberghsSven (changes requested):

ValenberghsSven (approved):

PR #1764, opened 2023-06-01
add isGenerated argument to doc card to change the label

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4072

### DIFF
--- a/app/components/documents/document-card-list.hbs
+++ b/app/components/documents/document-card-list.hbs
@@ -10,6 +10,7 @@
         @didDeleteContainer={{@didDeleteContainer}}
         @onOpenUploadModal={{@onOpenUploadModal}}
         @onAddPiece={{@onAddPiece}}
+        @isGenerated={{false}}
       />
     </div>
   {{else}}
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -41,7 +41,7 @@
               {{! no piece.created in legacy data }}
               {{#if this.piece.created}}
                 <p data-test-document-card-primary-source-created={{if this.piece.file.derived ""}}>
-                  {{t "uploaded-at"}}
+                  {{t this.label}}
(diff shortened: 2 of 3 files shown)


app/components/documents/document-card.hbs
```diff
@@ -41,7 +41,7 @@
               {{! no piece.created in legacy data }}
               {{#if this.piece.created}}
                 <p data-test-document-card-primary-source-created={{if this.piece.file.derived ""}}>
-                  {{t "uploaded-at"}}
+                  {{t this.label}}
```
kobemertens: Works fine like this, however I would move the translation into the `label` getter, this way the name `label` better reflects its contents.

app/components/documents/document-card-list.hbs
```diff
@@ -10,6 +10,7 @@
         @didDeleteContainer={{@didDeleteContainer}}
         @onOpenUploadModal={{@onOpenUploadModal}}
         @onAddPiece={{@onAddPiece}}
+        @isGenerated={{false}}
```
kobemertens, line 13: Would remove since this is falsy by default

ValenberghsSven: I had him add this because I was thinking in function of the list decisions in de subcase route.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/templates/cases/case/subcases/subcase/decision.hbs
We use the document-card-list there to show decisions (multiple if subcase was postponed)
We will probably need to have `@isGenerated={{@isGenerated}}` in this card-list template so that we can call `<Documents::DocumentCardList>` with `isGenerated` (after calculating which one in the route).

Could remove this for now.

kobemertens: ok, works for me to leave it as well

ValenberghsSven (approved):

PR #1768, opened 2023-06-05
KAS-4110 cypress test for support issues previous sprints

MathiasZabot:


ValenberghsSven (approved):
should be good, wait for cypress.

ValenberghsSven (approved):

PR #1769, opened 2023-06-06
KAS-4118: Add secretary data

sergiofenoll:
Adds secretary related relationships and properties.

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/394


app/models/agenda.js
```diff
@@ -18,6 +18,7 @@ export default class Agenda extends Model.extend(LoadableModel) {
   @belongsTo('agendastatus', { inverse: null, async: true }) status;
   @belongsTo('agenda', { inverse: 'previousVersion', async: true }) nextVersion; // Set in agenda-approve-service, read-only here
   @belongsTo('agenda', { inverse: 'nextVersion', async: true }) previousVersion; // Set in agenda-approve-service, read-only here
+  @belongsTo('user', { inverse: 'secretaryForAgendas', async: true }) secretary;
```
ValenberghsSven: Since we don't have the implementation yet it's hard to say but we save agenda very frequently during daily work.
So if this gets changes during that, the risk of "reverting" the change due to concurrency is very real.

@tdn Does it make more sense to connect the secretary to the meeting instead?
I think this might be a miscommunication (agenda and meeting are both called "agenda" by the users)
But I don't think agenda A will have a different secretary than agenda B or C.
So we just have duplication now + if you do change the secretary last minute then only the latest agenda will have the correct secretary.

ValenberghsSven (commented):
some comments in the app-kaleidos PR aswell

app/models/agenda.js
tdn: Yes, this will be the secretary for the meeting. My mistake of not making this clear in the story

tdn (approved):

app/models/mandatee.js
```diff
@@ -7,6 +7,7 @@ export default class Mandatee extends Model {
   @attr('number') priority;
   @attr('datetime') start;
   @attr('datetime') end;
+  @attr('boolean') isCurrentSecretary;
```
ValenberghsSven: this boolean is no longer used / implemented on app-kaleidos PR

ValenberghsSven (approved):
Minor change, can be done here or in the follow-up PR's to reduce conflicts

PR #1775, opened 2023-06-13
Feature/kas 4117 rework op te volgen lijst digitaal tekenen

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4117

Related PR: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/402

### Haven't been able to properly test this since activities and statuses aren't synched.


app/components/search/minister-filter.hbs
```diff
@@ -2,39 +2,41 @@
   <Auk::Loader />
 {{else}}
   <Auk::CheckboxTree
-    @label={{t "current-ministers"}}
+    @label={{t this.label}}
```
kobemertens: This label should say "Huidige Ministers" or "Alle Ministers". This label describes what the user will be selecting when clicking the checkbox, not what the function of the selected ministers is. If the function is ambiguous and needs to be described, the label should be put somewhere else (maybe above the checkboxes)

kobemertens: like this one
image

kobemertens: In this case I would not create a separate getter for this. I would just write `@label="{{t (if @label @label "current-ministers")}}` or even `@label="{{t (or @label "current-ministers)}}"`. I would also pass the entire label instead of the translation key because this fits the argument name more and is more consistent with other code in the codebase.

app/components/search/minister-filter.js
```diff
@@ -38,6 +38,14 @@ export default class SearchMinisterFilterComponent extends Component {
...
   }
 
+  get label() {
+    if(this.args.label) {
+      return this.args.label;
+    } else {
+      return "current-ministers";
+    }
+  }
+
```
kobemertens: Would remove this, or rename to a more descriptive name since there are many labels in this component

app/controllers/signatures/ongoing.js
```diff
@@ -3,8 +3,15 @@ import { tracked } from '@glimmer/tracking';
...
 import { PAGINATION_SIZES } from 'frontend-kaleidos/config/config';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
 
 export default class SignaturesOngoingController extends Controller {
+  @service router;
+  @service mandatees;
+  @service intl;
+  @service currentSession
```
kobemertens: This one is never used

app/controllers/signatures/ongoing.js
```diff
@@ -16,12 +23,38 @@ export default class SignaturesOngoingController extends Controller {
...
+    },
+  ];
+
+  statusFilterItems = [
+    {
+      label: this.intl.t('signing-status-to-be-prepared'),
+      value: CONSTANTS.SIGNFLOW_STATUSES.PREPARED,
+    }, 
+    {
+      label: this.intl.t('signing-status-to-be-signed'),
```
kobemertens: Just to be sure, the MARKED status should be annotated with `signing-status-to-be-signed`? Because I see there is also a TO_BE_SIGNED status (although that is not yet in use).

app/controllers/signatures/ongoing.js
```diff
@@ -55,8 +75,38 @@ export default class SignaturesOngoingController extends Controller {
       );
       if (response.ok) {
         const result = await response.json();
-        return result.url;
+        window.location.replace(result.url);
```
kobemertens: I assume we want users to be able to use the back button to return to this page? In that case you should use `window.location.href = result.url`, otherwise this page will be overwritten in the browsers history. (see https://developer.mozilla.org/en-US/docs/Web/API/Location/replace)

app/routes/signatures/ongoing.js
```diff
@@ -18,22 +19,49 @@ export default class SignaturesOngoingRoute extends Route {
       refreshModel: true,
       as: 'sorteer',
     },
+    mandatees: {
+      refreshModel: true,
+      as: 'indieners',
+    },
+    statuses: {
+      refreshModel: true,
+      as: 'statuses',
```
kobemertens: Should be "statussen"

app/templates/signatures/ongoing.hbs
```diff
@@ -1,104 +1,120 @@
...
+  <m.sidebar>
+    <div class="au-c-sidebar">
+      <div class="au-c-sidebar__content">
+        <div
+          id="search-subroute-filters-area"
+          class="au-o-box au-o-box--small au-o-flow au-o-flow--tiny"
+        >
+          <div class="au-c-card au-o-box au-o-box--small">
+            <Auk::CheckboxTree
+              @label={{t "status"}}
```
kobemertens: This should be "Selecteer Alle" as well and a separate label should be used (as with the minister filter)

kobemertens (changes requested):

kobemertens:
Would be nice to display a loader over the whole route while the model is loading for the first time. Now when the "Op te volgen" the tab is clicked, it is highlighted and it takes some time to transition, meanwhile the user gets no feedback.

app/templates/signatures/ongoing.hbs
```diff
@@ -1,104 +1,120 @@
...
+                >
+                  {{t "open"}}
+                </AuButton>
+              </td>
+                </:body>
+              </KDataTable>
+            {{else}}
+              <Auk::EmptyState
+                @bordered={{true}}
+                @message={{t "no-results-found-modify-term"}}
```
kobemertens: This one needs to change, there is no search term here

kobemertens: Maybe also differentiate between no items because of filters and no items because there are no items in the database. So 2 different messages

app/templates/signatures/ongoing.hbs
```diff
@@ -1,104 +1,120 @@
...
+                        {{type.label}}
+                      {{else}}
+                        {{t "no-document-type"}}
+                      {{/if}}
+                    {{/let}}
+                  </AuPill>
+                </td>
+              {{/let}}
+              <td>
+                {{#let (await (this.getMandateeNames signFlow)) as |names|}}
```
kobemertens: Not that important, but there is no way for a user to differentiate between "loading" and "no minister" here. Might be useful to display a loader while the mandatees are loading, like we do for the status pill

app/controllers/signatures/ongoing.js
ValenberghsSven: It is used in the template of this controller.
Missing semicolon though.

ValenberghsSven:
> Would be nice to display a loader over the whole route while the model is loading for the first time. Now when the "Op te volgen" the tab is clicked, it is highlighted and it takes some time to transition, meanwhile the user gets no feedback.

@StuyckRuben for this issue, take a look at this piece of code:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-4117-rework-design-digitaal-tekenen/app/routes/publications/overview/_base.js#L69
or this one:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/routes/agenda/agendaitems.js#L146

We want to bubble to the loading route sometimes, we can check the transition and not bubble if only queryParams have changed

app/templates/signatures/ongoing.hbs
StuyckRuben: Would this not be inconsistent with the rest of the code because in Search it doesn't use a seperate label for "Selecter Alle" either?

ValenberghsSven:
I'm going to take over a bit.
I see that refreshing with filters on does not work properly. The address bar has the filter id's but the filter is blank.

ValenberghsSven:
The refreshing with mandatee filter works now, aswell as resetting the page number when needed.
The statusses filter does not work yet because we have our own list instead of the model of the status, which is from another PR

app/controllers/signatures/ongoing.js
```diff
@@ -16,12 +23,38 @@ export default class SignaturesOngoingController extends Controller {
...
+      label: this.intl.t('signing-status-prepared'),
+      value: CONSTANTS.SIGNFLOW_STATUSES.PREPARED,
+    }, 
+    {
+      label: this.intl.t('signing-status-marked'),
+      value: CONSTANTS.SIGNFLOW_STATUSES.MARKED,
+    },
+    {
+      label: this.intl.t('signing-status-signed'),
+      value: CONSTANTS.SIGNFLOW_STATUSES.SIGNED,
```
tdn: The idea of adding the `SIGNFLOW_STATUSES` constants in `app/config/constants.js` was so you could loop over them, while we could still control which statuses to show in the filter and which to hide (at the time of the ticket we only supported 3 statuses). Even that is obsolete in the meantime as we now support all statuses, so you can just get them from the database.

These are concept resources with a label in the database, so there's no need for translation keys.

I changed this so `statusFilterItems` is collected from the resources store using the conceptScheme for the statuses. Also split off the status filter in a component for easier reuse and cleaner code

app/routes/signatures/ongoing.js
```diff
@@ -18,22 +20,64 @@ export default class SignaturesOngoingRoute extends Route {
...
+            'person': {
+              ':id:': params.mandatees.join(','),
+            }
+          }
+        }
+      }
+    }
+    if (params.statuses?.length > 0) {
+      filter['status'] = {
+        ':uri:': params.statuses.join(','),
```
tdn: This needed to be done using `:id:`, so the filters are combined using OR instead of AND

tdn (approved):
Made some changes in follow-up commits. After those this seems to work well when proxying to dev now.

tdn:
In hindsight, there's another issue that we'll need to fix. The mandatees shown in the list are not the same as when you filter.

app/templates/signatures/ongoing.hbs
tdn: Changed this to use the built-in KDataTable message (more generic "Geen resultaten gevonden"). No way to know if this is because of filters or not

app/controllers/signatures/ongoing.js
tdn: Also: renamed this to `navigateToSignFlow`

ValenberghsSven (approved):

PR #1776, opened 2023-06-14
Feature/kas 4124 abbreviated reviewer address

sergiofenoll:
Uses the `Auk::AbbreviatedText` component as recommended by Sven to cut off long email addresses of reviewers and notifications. Also changes the validation to no longer take into account length, or rather only check for the true max length of emails as defined in the pretinent RFCs: https://www.rfc-editor.org/errata/eid1690

Validation no longer uses inputmask because adapting the mask is really annoying and does not work as expected (it looks regex-y, but it doesn't really work as expected and you seemingly can't escape characters as you would in a normal regex engine).

Cfr. inputmask's default email mask: https://robinherbots.github.io/Inputmask/#/documentation#dynamic-masks
```
//email mask
  $(selector).inputmask({
    mask: "*{1,20}[.*{1,20}][.*{1,20}][.*{1,20}]@*{1,20}[.*{2,6}][.*{1,2}]",
    greedy: false,
    onBeforePaste: function (pastedValue, opts) {
      pastedValue = pastedValue.toLowerCase();
      return pastedValue.replace("mailto:", "");
    },
    definitions: {
      '*': {
        validator: "[0-9A-Za-z!#$%&'*+/=?^_`{|}~\-]",
        casing: "lower"
      }
    }
  });
```


app/components/signatures/email-modal.js
```diff
@@ -1,30 +1,17 @@
...
-
   @action
   onInputEvent(event) {
     this.emailBuffer = event.target.value;
   }
 
   get isValid() {
-    return this.inputmask.isValid(this.emailBuffer);
+    return this.emailBuffer?.length <= 254
+      && /.+\@.+\..+/.test(this.emailBuffer);
```
ValenberghsSven: Unnecessary escape character: \@.eslint[no-useless-escape](https://eslint.org/docs/rules/no-useless-escape) op deze lijn

ValenberghsSven (changes requested):

app/components/signatures/email-modal.js
sergiofenoll: Ah shit, ik heb de pre-commit git hook niet geënabled op mijn laptop, sorry :disappointed:

ValenberghsSven (approved):

PR #1779, opened 2023-06-14
Feature/kas 4044 minutes pdf export

kobemertens:


app/controllers/agenda/minutes.js
```diff
@@ -185,12 +214,68 @@ export default class AgendaMinutesController extends Controller {
...
+
+    const newPiecePart = this.store.createRecord('piece-part', {
+      value: this.currentPiecePart.value.value,
+      created: new Date(),
+      previousPiecePart: this.currentPiecePart.value,
+      minutes: newVersion,
+    });
+
+    await newVersion.save();
+    await newPiecePart.save();
```
tdn, line 266: Unsure why we need to `await newVersion.save();` first here, and not first `await newPiecePart.save();`?

app/routes/agenda/minutes.js
```diff
@@ -54,11 +54,11 @@ export default class AgendaMinutesRoute extends Route {
...
     controller.meeting = meeting;
-    // const agenda = this.modelFor('agenda').agenda;
-    // controller.agenda = agenda;
-    // controller.defaultAccessLevel = this.defaultAccessLevel;
+    const agenda = this.modelFor('agenda').agenda;
+    controller.agenda = agenda;
     controller.editor = null;
     controller.isEditing = false;
     controller.isFullscreen = false;
+    controller.editor = null;
```
tdn, line 61: Double with line 59

tdn (approved):
made two tiny changes. 

Code in the controller is a bit messy, but would rather merge this now and clean it up later, since we'll have to come back to this anyway once all the mandatee-stuff around secretary has been cleared up.

app/controllers/agenda/minutes.js
```diff
@@ -7,11 +7,13 @@ import constants from 'frontend-kaleidos/config/constants';
 import { SECRETARIS_NAME } from 'frontend-kaleidos/config/config';
 import { task as trackedTask } from 'ember-resources/util/ember-concurrency';
 import { dateFormat } from 'frontend-kaleidos/utils/date-format';
+import { deleteFile } from 'frontend-kaleidos/utils/document-delete-helpers';
+import VRDocumentName from 'frontend-kaleidos/utils/vr-document-name';
 
 function renderAttendees(attendees) {
```
tdn, line 13: Lots of these functions outside the controller would be better as utils and/or getters inside the controller. But we'll handle this later so this can be tested.

kobemertens: Why?

PR #1783, opened 2023-06-19
add multiple snippets in highlight

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4123


app/routes/search/agendaitems.js
```diff
@@ -69,6 +68,13 @@ export default class AgendaitemSearchRoute extends Route {
         entry.shortTitle = entry.shortTitle.replaceAll(highlightShortTitleEmsRemoved, agendaitem.highlight.shortTitle[i]);
       }
     } 
+    if ((agendaitem.highlight && agendaitem.highlight.title) && agendaitem.highlight.title[0]?.split(" ").length < agendaitem.attributes.title?.split(" ").length ) {
```
kobemertens: would make sense to make this behavior a function since you are doing exactly the same for the `title` as well as `shortTitle`

kobemertens: Does this condition need to be more than just `agendaitem.highlight?.title`? Seems like a redundant optimization to me and reduces readability.

app/routes/search/agendaitems.js
```diff
@@ -69,6 +68,13 @@ export default class AgendaitemSearchRoute extends Route {
...
+      entry.title = agendaitem.attributes.title;
+      for (let i = 0; i < agendaitem.highlight.title.length; i++){
```
kobemertens: I would consider using a `for ... of` loop. `for (const titleHighlight of agendaitem.highlight.title)`

app/routes/search/agendaitems.js
```diff
@@ -69,6 +68,13 @@ export default class AgendaitemSearchRoute extends Route {
...
+        const highlightTitleEmsRemoved = agendaitem.highlight.title[i].replace(/<em>|<\/em>/g, "");
+        entry.title = entry.title.replaceAll(highlightTitleEmsRemoved, agendaitem.highlight.title[i]);
```
kobemertens: I would suggest to use an autoformatter since your lines tend to get very long

kobemertens (changes requested):

PR #1787, opened 2023-06-23
feat: use the status attached to signflow

kobemertens:
implements https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-4131&assignee=63c552d46178fcc941d857c9&assignee=UNASSIGNED_USER_ID


app/components/signature-pill.js
```diff
@@ -11,84 +11,80 @@ export default class SignaturePillComponent extends Component {
...
+      case to_be_signed:
+        return this.intl.t('to-sign');
+      case SIGNED:
+        return this.intl.t('signed');
+      case REFUSED:
+        return this.intl.t('refused');
+      case CANCELED:
+        return this.intl.t('cancelled');
+      default:
+        return '';
```
tdn: We have a codelist with labels for this specific reason.

You'll need to get the `concept` resource for `this.status.uri` from the store for each status pill.
Those have a label.

tdn: The translations can then be removed as well. (do check if they're used anywhere else)

tdn (changes requested):

app/components/signature-pill.js
kobemertens: implemented in [3115039](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1787/commits/31150395fc58d04507ff7e8fadf6e97772e81e1d)

app/components/signature-pill.js
```diff
@@ -11,84 +11,49 @@ export default class SignaturePillComponent extends Component {
...
-  get skin() {
-    if (this.isRefused || this.isCancelled) {
-      return "error";
-    } else if (this.signingHubUrl) {
-      return "link"
-    } else {
-      return "ongoing"
-    }
-  }
+    return await (await signSubcase.signFlow).status;
```
tdn: I get why we need to `await` both here, but I get shivers up and down my spine when I see this double `await` 😅

I might put the signflow in a variable here

tdn: In hindsight. Couldn't this just be an async getter?

I suppose that would clutter the rest of the code with `await this.status` too much. I'm ok with it for now, but would prefer someone who works in Ember more often than me to take a look at it.

app/components/signature-pill.js
```diff
@@ -11,84 +11,49 @@ export default class SignaturePillComponent extends Component {
...
+    const { SIGNED, REFUSED } = constants.SIGNFLOW_STATUSES;
+    if (!this.status.value || this.status.value.uri === REFUSED) {
+      return null;
     }
-    return "";
-  }
 
-  loadData = task(async () => {
+    const currentUser = this.currentSession.user;
+    const status = this.status.value;
```
tdn: this assignment is quite unnecessary. Removing it

app/components/signature-pill.js
```diff
@@ -11,84 +11,49 @@ export default class SignaturePillComponent extends Component {
...
-          const response = await fetch(
-            `/signing-flows/${signFlow.id}/pieces/${piece.id}/signinghub-url?collapse_panels=false`
-          );
-          if (response.ok) {
-            const result = await response.json();
-            this.signingHubUrl = result.url;
-          } 
-        }
+    const signSubcase = await signMarkingActivity.signSubcase;
+    const piece = await pieceArg;
```
tdn: Why not `await this.args.piece` directly?

tdn (approved):
Did a bit of code cleanup (see comments), but otherwise works and looks fine.

@ValenberghsSven you have the final say to whether or not this can be merged. I don't know how often we use things like async `trackedFunction`

app/components/signature-pill.js
SchrodingersCat00: What do you mean by async getter? Because as far as I know there is no such thing

app/components/signature-pill.js
kobemertens: > Also note that after an await, the this context should not be accessed as it could lead to destruction/timing issues.

from https://ember-resources.pages.dev/funcs/util_function.trackedFunction
That is why I created this variable

app/components/signature-pill.js
kobemertens: This was indeed not necessary, was a leftover of some experimentation with `trackedFunction`

app/components/signature-pill.js
kobemertens: I guess this is kind of the same as an async "getter" (which will be a regular function because getters cannot be async) with the benefit of not having to use any promise helpers. Ember (and Redpencil) seems to be shifting towards ember-resources based patterns so that is why I used them here. There is still some learning to do but it seems to work fine.

app/components/signature-pill.js
kobemertens: Thinking about this, I realised my approach probably won't solve the problem since this is a reference to the same object `this` would be pointing to, which has the same problems as using `this` directly has.

app/components/signature-pill.js
```diff
@@ -11,84 +11,48 @@ export default class SignaturePillComponent extends Component {
...
-  @tracked isRefused = false;
-  @tracked isCancelled = false;
-
-  constructor() {
-    super(...arguments);
-
-    this.loadData.perform();
-  }
+  status = trackedFunction(this, async () => {
+    const signMarkingActivity = await this.args.signMarkingActivity;
```
ValenberghsSven, line 17: stuff coming in from `args` should not need to be awaited.

app/components/signature-pill.js
```diff
@@ -11,84 +11,48 @@ export default class SignaturePillComponent extends Component {
...
+    const signSubcase = await signMarkingActivity.signSubcase;
+    const signFlow = await signSubcase.signFlow;
```
ValenberghsSven, line 20: can `signSubcase` and `signFlow` be null? unsure.
A screenshot of error on ACC seems to have an occurence (could be old data)

app/components/signature-pill.js
```diff
@@ -11,84 +11,48 @@ export default class SignaturePillComponent extends Component {
...
+    if (!this.status.value || this.status.value.uri === REFUSED) {
+      return null;
     }
-    return "";
-  }
 
-  loadData = task(async () => {
+    const currentUser = this.currentSession.user;
+    const piece = await this.args.piece;
     const signMarkingActivity = await this.args.signMarkingActivity;
```
ValenberghsSven: same, `args` should not have to be awaited.

app/components/signature-pill.js
```diff
@@ -11,84 +11,48 @@ export default class SignaturePillComponent extends Component {
...
-  get skin() {
-    if (this.isRefused || this.isCancelled) {
-      return "error";
-    } else if (this.signingHubUrl) {
-      return "link"
-    } else {
-      return "ongoing"
-    }
-  }
+    return await signFlow.status;
```
ValenberghsSven: Since the updates will happen in the backend we might have to do a `await signFlow.belongsTo('status').reload()` to get the changed status without a hard refresh (F5) to update the store. Need to test later.
Can leave it for now.

ValenberghsSven (commented):
Minor changes, nothing serious.
I am thinking:
Why are we not using a service call (inside trackedFunction is fine) to call the url.
We got some code duplication because we are fetching the same url in 3 places ( so far).

Don't see the difference between a general await in the constructor vs trackedFunction.
Not sure if I am a fan yet of those functions

kobemertens:
> Don't see the difference between a general await in the constructor vs trackedFunction.
Not sure if I am a fan yet of those functions

First advantage over await in constructor is that `trackedFunction` approach does not break reactivity. When an argument changes, the `trackedFunction` value changes as well. This is not the case for the former approach as the constructor is not run again after argument changes. This can be patched by adding a render modifier but this approach is very much discouraged.

https://www.pzuraq.com/blog/introducing-use <- Interesting read
More specifically this: https://www.pzuraq.com/blog/introducing-use#resources

app/components/signature-pill.js
tdn: Then this can also be used to solve the "accessing `this` after `await`" issue.

You can do:
```
const piece = this.args.piece;
const signMarkingActivity = this.args.signMarkingActivity;
if (this.status.uri !== SIGNED) {
   ... rest of the code ...
} 
return null;
```

ValenberghsSven:
> > Don't see the difference between a general await in the constructor vs trackedFunction.
> > Not sure if I am a fan yet of those functions
> 
> First advantage over await in constructor is that `trackedFunction` approach does not break reactivity. When an argument changes, the `trackedFunction` value changes as well. This is not the case for the former approach as the constructor is not run again after argument changes. This can be patched by adding a render modifier but this approach is very much discouraged.
> 
> https://www.pzuraq.com/blog/introducing-use <- Interesting read More specifically this: https://www.pzuraq.com/blog/introducing-use#resources

I can see the benefit for trackedFunction in some cases where we need reactivity (but you still have actions/tracked properties/tasks you can use in that case). Multiple solutions to a problem
In this case however, why/when would the args (piece and markingActivity) change once the component is loaded?

app/components/signature-pill.js
kobemertens: Why not? Not sure from where, but there is a promise being passed as arg here

ValenberghsSven: if args is a promise we should be awaiting it in controller/component where we use this component.
2 out 4 places where this pill is used have the data loaded before rendering the component. Should be 4 out of 4.
I would argue that if there is no `signMarkingActivity` then we should not even render the pill at all.

app/components/signature-pill.js
kobemertens: This is just copied from the old code, we should ask @tdn.

app/components/signature-pill.js
kobemertens: Same as above, should be fixed somewhere else first so there is no promise passed as arg

kobemertens:
> I can see the benefit for trackedFunction in some cases where we need reactivity (but you still have actions/tracked properties/tasks you can use in that case). Multiple solutions to a problem

Yes, and I think it is beneficial to explore different solutions to see what they offer.

>  In this case however, why/when would the args (piece and markingActivity) change once the component is loaded?

On the "op-te-volgen" page I don't see this happening but we should aim to make our components reusable anywhere. I can for instance imagine a scenario where we have some kind of signflow inspector panel where you can see details for a signflow. When you click on another signflow, the piece argument to the signature-pill in the panel might change and we want the detail panel to update accordingly.

app/components/signature-pill.js
tdn: If there's a signMarkingActivity there should always be a `signSubcase` and `signFlow`. Anything else would idd be data inconsistency.

ValenberghsSven: > If there's a signMarkingActivity there should always be a `signSubcase` and `signFlow`. Anything else would idd be data inconsistency.

fine with me to keep this as-is

ValenberghsSven:
> On the "op-te-volgen" page I don't see this happening but we should aim to make our components reusable anywhere. I can for instance imagine a scenario where we have some kind of signflow inspector panel where you can see details for a signflow. When you click on another signflow, the piece argument to the signature-pill in the panel might change and we want the detail panel to update accordingly.

In that case, I would personally hide certain components while loading the new signFlow and showing them when the new data is loaded in order to prevent showing wrong states in between (f.e. some data from signFlow 2 with a pill from signFlow 1 because data is still updating). 
Rerendering a component is an easy way to ensure that no stale/wrong data is present. Showing a loader instead also gives visual feedback to the user when the process takes some time.
It also makes the components less complex since we don't have to take into account that the args could change at any time.

kobemertens:
I agree that a lot of times we will end up replacing a component after an argument changes (because of loading). But I also believe that reactivity is such a core concept in Ember that it feels counterproductive to break it. Also, a benefit of not having to replace the component is that it will not refetch data that is not affected by an argument change.

app/components/signature-pill.js
```diff
@@ -11,84 +12,53 @@ export default class SignaturePillComponent extends Component {
...
-          } 
+    if (status.uri !== REFUSED) {
+      const piece = await this.args.piece;
+      const signFlow = await signSubcase.signFlow;
+      const signFlowCreator = await signFlow.creator;
+      const currentUser = this.currentSession.user;
+      if (
+        piece &&
+        signFlowCreator.id === currentUser.id &&
+        status.uri !== SIGNED
```
tdn: Found while optimizing another PR. It would be nice if we check for `MARKED` status here as well. Signflows with that status will never have a SigningHub URL, so the call will always fail here.

ValenberghsSven (approved):

PR #1789, opened 2023-06-23
feat: add delete action

kobemertens:
implements https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-4139&assignee=63c552d46178fcc941d857c9&assignee=UNASSIGNED_USER_ID


app/utils/document-delete-helpers.js
```diff
@@ -60,7 +64,7 @@ export async function deletePiece(pieceOrPromise) {
  */
 export async function deleteReport(reportOrPromise) {
   const report = await reportOrPromise;
-  const pieceParts = await report.pieceParts.toArray();
+  const pieceParts = (await report.pieceParts).toArray();
```
ValenberghsSven: are the brackets needed?
if yes, It might be better to just put the `toArray()` in the for loop and not have the await wrapped in brackets.
`for (const piecePart of pieceParts.toArray()) {`

app/utils/document-delete-helpers.js
```diff
@@ -22,6 +22,10 @@ export async function deleteDocumentContainer(documentContainerOrPromise) {
         await deletePiece(latestPiece);
       } else if (latestPiece.constructor.modelName === 'report') {
         await deleteReport(latestPiece);
+      } else {
+        console.debug(
+          'Piece subclass might not be removed propery, add override to document-delete-helpers.js'
+        );
```
ValenberghsSven, line 28: do we need console debug? we only need to add minutes modelname later and we are good.

ValenberghsSven (commented):
When you delete the report and immediately make a new one there is already data in the editors
Also: are we going to also fix document-view delete here or not? You asked in the ticket. I would say yes.

app/utils/document-delete-helpers.js
kobemertens: I guess we only want this in development mode, but I do think we should leave it. If we ever create another piece subtype, we won't forget to remove its relations this way.

app/utils/document-delete-helpers.js
kobemertens: It did not work without, will move the `toArray()`

kobemertens: fixed in [53df4f4](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1789/commits/53df4f4ffd9a503400cc29b3b113757e94552faf)

kobemertens:
>  Also: are we going to also fix document-view delete here or not? You asked in the ticket. I would say yes.

I did not check, but I assume the `document-view` uses the service to remove pieces? So in that case it should work already

kobemertens:
> When you delete the report and immediately make a new one there is already data in the editors

This is fixed in [0d62406](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1789/commits/0d62406630bee73727c6508e56998ed6ecef75b0)

app/utils/document-delete-helpers.js
ValenberghsSven: I would argue you can still forget if you don't use this helper.
If you do use this helper for future sub-classes, you should look at what the code does and if it's suitable for your situation.
In that case, A comment would have been enough.

We could also add comments to the resource file for piece but this is not exclusive to piece. Perhaps we will have more polymorphism later on with a similar issue.
Perhaps we could extend the readme file and add some notes on polymorphism in the frontend.

In hindsight, we could probably avoid using the modelName and just make deletes optional.
we try to load `piece.pieceParts`. If it's empty, no delete should happen.

kobemertens: What is the problem with having a log message printed anyway?

ValenberghsSven (approved):

app/utils/document-delete-helpers.js
ValenberghsSven: not really a problem if debug flag is not on. fine as-is

PR #1792, opened 2023-06-26
add functionality in bulk edit modal to mark piece for signing and cr…

StuyckRuben:
…eate signflow upon saving
https://kanselarij.atlassian.net/browse/KAS-4145


tdn (approved):
Separated one yield statement. Otherwise looks & works good

ValenberghsSven (approved):

PR #1793, opened 2023-06-27
Feature/kas 4144 rework op te starten lijst digitaal tekenen

sergiofenoll:
> **Warning**
> This PR requires https://github.com/kanselarij-vlaanderen/digital-signing-service/pull/26 and the related backend PR to work properly

Makes the signatures overview ("Op te starten" lijst) in-line with the new design and let it work with the new bundling system. Updates the signatures-tab in the document preview to also cope with marking activities potentially not existing.

:information_source: Because this relies on the status URI to do the filtering, there's some delay between preparing a sign flow (sending it to SH) and the status URI being set. This may result in the items that were selected in the "Op te starten" list still remaining in the list after creating the sign flow. We may want to tweak the `gracePeriod` on the delta rules, or rely on the existence of activities for that list.

### DIFF
--- a/app/services/signature-service.js
+++ b/app/services/signature-service.js
@@ -8,43 +8,42 @@ export default class SignatureService extends Service {
   @service intl;
   @service currentSession;
 
-  async createSignFlow(piece, decisionActivity, signers, approvers, notified) {
-    // Create sign flow, sign subcase and marking activity
-    const { signFlow, signSubcase } = await this.markDocumentForSignature(
-      piece,
-      decisionActivity
-    );
+  async createSignFlow(signFlows, signers, approvers, notified) {
+    for (let signFlow of signFlows) {
+      const signSubcase = await signFlow.signSubcase;
+      // Attach signers
+      await Promise.all(
+        signers.map((mandatee) => {
+          const record = this.store.createRecord('sign-signing-activity', {
+            signSubcase,
+            mandatee,
+          });
+          return record.save();
+        })
+      );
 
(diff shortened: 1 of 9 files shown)


app/services/signature-service.js
```diff
@@ -125,11 +124,11 @@ export default class SignatureService extends Service {
...
-  async removeSignFlow(piece) {
-    const signMarkingActivity = await piece.signMarkingActivity;
-    if (signMarkingActivity) {
-      const signSubcase = await signMarkingActivity.signSubcase;
-      const signFlow = await signSubcase?.signFlow;
+  async removeSignFlow(signFlow) {
+    if (signFlow) {
+      const signSubcase = await signFlow.signSubcase;
+      const signMarkingActivity = await signSubcase.signMarkingActivity;
+      const piece = await signMarkingActivity.piece;
```
tdn, line 131: @ValenberghsSven I seem to remember the order of removing things here was important. Can you double check this please?

tdn (approved):
Looks good to me, pending review of @MikiDi, succesful tests & double-check on signFlow removal by @ValenberghsSven

sergiofenoll:
Handtekeningen > Op te starten list test is failing here because the logic is completely different, but until #1792 is merged there's no in-app way to mark a document for signing. Do we wait with merging until the other one is merged and then fix the tests before merging, or do we temporarily disable that test?
@ValenberghsSven

ValenberghsSven (approved):
pair review

PR #1796, opened 2023-06-28
Feature/kas 4129 agenda detail constant refresh

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4129

Took awhile to find this one.
There is an active bug in Ember where the "transitionTo incorrectly re-runs the model hook when query params do not change".
Linked the issue in JIRA ticket.
In some cases (fe. when going from agenda.documents to agenda.agendaitems) we did not change query params, but the model kept refreshing to due to "refreshModel" begin true on the filter QP. Even though there was no change in QP.

This had some side-effects:
Switching to a subroute of 1 agendaitem (agenda.agendaitem.agendaitem.**) refreshed the entire parent route (agenda.agendaitems)
The links in the sidebar were always pointing to the "index" route, even if we were on the "documents" route at the time.
We were reloading some views multiple times (users did report the agenda seemed slower, could this be part of it?)

In order to fix this:
- refreshModel has to be false for the filter not to force the model reload.
- in order to refresh the model when we change the filter I added a manual transition and a model reload with the new filter.
- have default ember loading behavior take care of the transitions to loading pages on all levels

I tested this through and all the loaders seemed to be shown in the correct places.
I could no longer reproduce a situation where the parent loader was shown when the child routes were traversed
Running this through Jenkins.

### DIFF
--- a/app/components/agenda/agenda-header/agenda-tabs.hbs
+++ b/app/components/agenda/agenda-header/agenda-tabs.hbs
@@ -13,6 +13,7 @@
   {{#if this.firstAgendaitem}}
     <Auk::Tab
       @route="agenda.agendaitems.agendaitem"
+      @current-when="agenda.agendaitems.agendaitem"
       @models={{this.modelsForDetailRoute}}
     >
       {{t "detail"}}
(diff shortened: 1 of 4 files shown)


sergiofenoll (approved):
Tested locally, works as expected

PR #1797, opened 2023-06-29
label on document card

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4072

### DIFF
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -58,8 +58,11 @@ export default class DocumentsDocumentCardComponent extends Component {
   }
 
   get label() {
-    console.log(this.args.label);
-    return this.intl.t(this.args.label);
+    if (isPresent(this.args.label)){
+      return this.intl.t(this.args.label);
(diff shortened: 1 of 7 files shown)


app/components/documents/document-card.js
```diff
@@ -58,11 +58,8 @@ export default class DocumentsDocumentCardComponent extends Component {
   }
 
   get label() {
-    if (this.args.isGenerated) {
-      return this.intl.t('created-on');
-    } else {
-      return this.intl.t('uploaded-at');
-    }
+    console.log(this.args.label);
+    return this.intl.t(this.args.label);
```
tdn: 2 remarks here:
- remove the console.log please
- it would be best if this getter defaults to `uploaded-at`, since `this.args.label` could be empty (and will be in alle existing cases). Only when it's overridden in the arguments does it need to be something else.

tdn (changes requested):

ValenberghsSven (approved):

PR #1798, opened 2023-06-30
Feature/KAS-4167 Make the signature pills poll their status regularly

sergiofenoll:
Uses an equivalent construction to what we do in the publication pills to poll the status.

We might want to replace usage of models and relationships (`signMarkingActivity.signSubcase; signSubcase.signFlow; signFlow.status`) with just a query. The query won't be cached, but given that we're explicitly trying to reload data that's probably a good thing in this case (and it means we don't need to do a bunch of manual reloads to make the code work when ran a second time with the same `signMarkingActivity`).
TBD if we want to change this, now, some other time, or never.

### DIFF
--- a/app/components/signature-pill.js
+++ b/app/components/signature-pill.js
@@ -25,7 +25,7 @@ export default class SignaturePillComponent extends Component {
     cancel(this.scheduledRefresh);
   }
 
-  schedulePublicationActivitiesRefresh() {
+  scheduleSignFlowStatusRefresh() {
     if (![REFUSED, SIGNED, CANCELED].includes(this.data?.value?.status?.uri)) {
       this.scheduledRefresh = later(this, async () => {
         this.args.signMarkingActivity.reload();
@@ -69,7 +69,7 @@ export default class SignaturePillComponent extends Component {
       }
     }
 
-    this.schedulePublicationActivitiesRefresh();
+    this.scheduleSignFlowStatusRefresh();
 
     return {
       signingHubUrl,
--- a/app/components/signature-pill.js
+++ b/app/components/signature-pill.js
@@ -1,9 +1,14 @@
 import Component from '@glimmer/component';
+import { later, cancel } from '@ember/runloop';
+import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
 import fetch from 'fetch';
 import constants from 'frontend-kaleidos/config/constants';
 import { task as trackedTask } from 'ember-resources/util/ember-concurrency';
 import { task } from 'ember-concurrency';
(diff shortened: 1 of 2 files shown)


app/components/signature-pill.js
```diff
@@ -12,8 +17,30 @@ export default class SignaturePillComponent extends Component {
...
 
+  scheduledRefresh;
+  @tracked triggerTask;
+
+  willDestroy() {
+    super.willDestroy(...arguments);
+    cancel(this.scheduledRefresh);
+  }
+
+  schedulePublicationActivitiesRefresh() {
```
tdn: Perhaps better to rename this to `scheduleSignflowStatusRefresh`

app/components/signature-pill.js
```diff
@@ -42,13 +69,15 @@ export default class SignaturePillComponent extends Component {
...
+    this.schedulePublicationActivitiesRefresh();
+
     return {
       signingHubUrl,
       status,
     };
   });
 
-  data = trackedTask(this, this.loadData);
+  data = trackedTask(this, this.loadData, () => [this.triggerTask]); // Make the resource dependant on this.triggerTask
```
tdn, line 77: I'm not sure I understand what's happening here.

So on line 72, on the first run of `this.loadData`, we load the first `status` and schedule the polling task. 
This will reload the `signFlow.status` after `SIGN_FLOW_STATUS_REFRESH_INTERVAL_MS` and set `this.triggerTask = new Date();`, which will trigger `this.loadData` because this trackedTask uses `this.triggerTask` as an argument ( https://ember-resources.pages.dev/funcs/util_ember_concurrency.trackedTask )

On line 43 in `this.loadData`, we then cancel the next `this.scheduledRefresh` to load the `status`, then we're back at line 72 and schedule the next task again.

So in other words, wouldn't it make more sense to put the reload code from lines 32 to 34 in `this.loadData` on lines 44 to 46? That way the scheduling task just updates the trigger for the trackedTask.
Or does this cause problems in Ember because you can't `reload()` on the first `await` of a relation?

tdn (changes requested):
Just a small rename of the copied scheduling task, and just a look at my comment on how the scheduled task reloads data (perhaps it can be done in loadData?).

Ran the branch locally, and the update call seems to work well as far as I can test using mock login

app/components/signature-pill.js
sergiofenoll: That also definitely works, I didn't want to change the original code too much. At that point we might be better off writing something like

tdn (approved):

ValenberghsSven (approved):

PR #1800, opened 2023-06-30
fix: filter out sign flows w/o piece and do cache busting in a silly way

sergiofenoll:
cache-busting: when an agenda/meeting is finalised the changes don't invalidate the calls to the sign flows that we make in this route. to always get a fresh call we just add timestamp to the call. we do this in the `:has:meeting` because the value of a `:has:` filter is never used, mu-cl-resources just checks if it's set to *something*.

### DIFF
--- a/app/routes/signatures/index.js
+++ b/app/routes/signatures/index.js
@@ -76,10 +76,10 @@ export default class SignaturesIndexRoute extends Route {
         'sign-subcase.sign-marking-activity.piece.document-container.type'
       ].join(','),
       page: {
-        number: params.page,
-        size: params.size,
+        number: params.pageSignaturesIndex,
+        size: params.sizeSignaturesIndex,
       },
-      sort: params.sort,
+      sort: params.sortSignaturesIndex,
     });
   }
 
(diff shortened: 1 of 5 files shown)


ValenberghsSven (approved):

app/routes/signatures/index.js
```diff
@@ -40,15 +40,19 @@ export default class SignaturesIndexRoute extends Route {
...
       status: {
         ':uri:': CONSTANTS.SIGNFLOW_STATUSES.MARKED,
       },
       'decision-activity': {
-        ':lte:start-date': (new Date()).toISOString().slice(0, 10), // Cache-busting
         treatment: {
           agendaitems: {
             agenda: {
-              ':has:meeting': true,
+              ':has:meeting': `ye-date-added-for-cache-busting-${(new Date()).toISOString()}`,
```
tdn, line 55: "ye olde date" ? 😆

sergiofenoll: trying to write `yes` very quickly and forgot to type the `s` . shakespeare style code :joy:

tdn (approved):
tested locally and works well

PR #1807, opened 2023-07-03
KAS-4177 Fix/pagination and pill color

sergiofenoll:
Small fixes for signatures

### DIFF
--- a/app/components/k-data-table.js
+++ b/app/components/k-data-table.js
@@ -46,10 +46,4 @@ export default class KDataTableComponent extends Component {
   onPreviousPage() {
     this.args.onChangePage(this.args.page - 1);
   }
-
-  @action
-  onChangeSize() {
-    this.args.onChangePage(0);
-    this.args.onChangeSize(...arguments);
-  }
 }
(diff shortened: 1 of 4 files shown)


app/components/k-data-table.js
```diff
@@ -46,4 +46,10 @@ export default class KDataTableComponent extends Component {
   onPreviousPage() {
     this.args.onChangePage(this.args.page - 1);
   }
+
+  @action
+  onChangeSize() {
+    this.args.onChangePage(0);
+    this.args.onChangeSize(...arguments);
+  }
```
ValenberghsSven: is this to change page to 0 when size is changed?
This wont cover the load (if you are on page 2 and filter the data you want to end up on page 0 again)
In other routes we have this
```
if (this.lastParams.anyFieldChanged(Object.keys(params).filter((key) => key !== 'page'))) {
      params.page = 0;
    }
```

sergiofenoll: Thanks, didn't know about the `Snapshot` construction. I've added that to this route and removed the changes to `KDataTable`.

tdn (approved):
Works well locally

PR #1810, opened 2023-07-04
KAS-4174 change status on create and filter on preparation-activity

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4174

Change status when creating/sending to SH (helps with the index view + the pills in other views
Filter view on preparation-activity (none should exist)

### DIFF
--- a/app/services/signature-service.js
+++ b/app/services/signature-service.js
@@ -4,7 +4,7 @@ import ENV from 'frontend-kaleidos/config/environment';
 import fetch from 'fetch';
 import constants from 'frontend-kaleidos/config/constants';
 
-const MARKED = constants.SIGNFLOW_STATUSES.MARKED;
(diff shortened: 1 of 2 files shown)


app/services/signature-service.js
```diff
@@ -43,6 +43,8 @@ export default class SignatureService extends Service {
       // set creator
       const creator = await this.currentSession.user;
       signFlow.creator = creator;
+      const status = await this.store.findRecordByUri('concept', PREPARED);
+      signFlow.status = status;
```
tdn, line 47: If we do this, we need to set the status back to MARKED after `uploadPiecesToSigninghub` fails

tdn (changes requested):
Need to set the status back to MARKED if the upload to signinghub fails.

tdn (approved):
We delete the whole signflow when the upload fails, so status doesn't need to be changed back.

PR #1813, opened 2023-07-05
feat: use decisionmaking-flow closed date

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?modal=detail&selectedIssue=KAS-3753

updated @ValenberghsSven:
Also includes https://kanselarij.atlassian.net/browse/KAS-4256


app/routes/cases/index.js
```diff
@@ -33,9 +33,9 @@ export default class CasesIndexRoute extends Route {
       },
     };
     if (params.showArchivedOnly) {
-      options['filter[case][is-archived]'] = true;
+      options['filter[:has:closed]'] = true;
     } else {
-      options['filter[case][is-archived]'] = false;
+      options['filter[:has-no:closed]'] = true;
     }
```
ValenberghsSven, line 39: On normal queries we shouldn't use `:has:` and `:has-no:` on properties since that is not really supported.
We should either check if there is a date or nothing.

ValenberghsSven (changes requested):
My opinion:
Closing a case is not the same as a case that was "deleted".
Right now we don't offer regular closing of cases (the government saying "We will never touch this case again, it's done").
But there is a difference to a "deleted" case no? Depends on the reason of "deletion"
You could say that a case is closed with a certain reason (maybe from a concept codelist?) 


If we do go through with the original strategy then the following needs to happen:
- Search still uses the flag to find the archived ones so that will no longer work.
```
        "isArchived":  [
          "^https://data.vlaanderen.be/ns/dossier#Dossier.isNeerslagVan",
          "http://mu.semte.ch/vocabularies/ext/isGearchiveerd"
        ],
```
- subcase model still has `isArchived`
- query needs to exist to remove the property from all cases and subcases
- the boolean is still being set on creation of case and subcase.
- cleanup in the domain resources.

app/routes/cases/index.js
ValenberghsSven: seems to work. Might not work 100% of the time?
I remember having a conversation that this was not really supported but was a long time ago.
Leaving it as-is for now.

sergiofenoll (approved):

ValenberghsSven:
1 test faalt nog omdat daar nog foute PATCH call instaan. Pas dat vandaag aan

ValenberghsSven (approved):
Tests look fine now besides some general flaky issue that is also on DEV

PR #1814, opened 2023-07-06
add try catch to show toaster if save fails

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4161

- [ ] merge https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/412


ValenberghsSven (changes requested):
The Jira ticket is not well groomed and very vague.
If something goes wrong with saving the report we should do more cleaning up of data.
We should rather investigate why it went wrong.
Just catching the error will put our data in an inconsistent state.
Also we use `this.saveReport` twice, so it would make more sense to catch the error inside `saveReport` method.

I would be inclined to close the PR and throw the ticket on the backlog for refinement.
Do we have the steps to reproduce to investigate the error?

StuyckRuben:
> The Jira ticket is not well groomed and very vague. If something goes wrong with saving the report we should do more cleaning up of data. We should rather investigate why it went wrong. Just catching the error will put our data in an inconsistent state. Also we use `this.saveReport` twice, so it would make more sense to catch the error inside `saveReport` method.
> 
> I would be inclined to close the PR and throw the ticket on the backlog for refinement. Do we have the steps to reproduce to investigate the error?



> The Jira ticket is not well groomed and very vague. If something goes wrong with saving the report we should do more cleaning up of data. We should rather investigate why it went wrong. Just catching the error will put our data in an inconsistent state. Also we use `this.saveReport` twice, so it would make more sense to catch the error inside `saveReport` method.
> 
> I would be inclined to close the PR and throw the ticket on the backlog for refinement. Do we have the steps to reproduce to investigate the error?

According to Mathias the error occured while both him and Johan were trying to simultaneously save the changes they made while editing the report

ValenberghsSven:
so a concurrency issue, it can happen at any stage of the saving process which means some data could have gone through and mess up the data from the other user.
Just catching errors is not a fix-all solution.
The first suggestion is to train users not to edit/generate the decision at the same time.

For the few times that they do we might have to consider a `modelWithModifier` like fix or something.
Still considering closing the PR and grooming the ticket after investigating possible long term fixes.

@tdn  thoughts?

tdn:
@ValenberghsSven closing this and grooming the ticket is Ok for me.

> Just catching the error will put our data in an inconsistent state.

That makes this PR unmergeable. Or does the data go in an inconsistent state regardless of whether we catch it? if so, then maybe we can merge this and expand on it in a new ticket?

ValenberghsSven:
> @ValenberghsSven closing this and grooming the ticket is Ok for me.
> 
> > Just catching the error will put our data in an inconsistent state.
> 
> That makes this PR unmergeable. Or does the data go in an inconsistent state regardless of whether we catch it? if so, then maybe we can merge this and expand on it in a new ticket?

The data can become inconsistent before we catch the eventual error.
So catching it would only help provide feedback to the user that something went wrong and prevent throwing an uncaught error upwards.
We could merge an error handling flow now (no data cleanup) and try to catch concurrency earlier in a later ticket.
But we should move the try catch inside `saveReport()` since we use that method twice and only 1 is wrapped in try catch

tdn:
I think we have room to fix the concurrency issues in the scope of this ticket.

ValenberghsSven:
Talked about this with @tdn 
The `modelWithModifier` is very strict for concurrency.
It also requires many checks pre edit / pre finalizing save.
If any of those checks fail you need a fallback.
For something as general as the piece model this would mean a lot of work because it is used in so many contexts.
I need to look into this.
Perhaps we could use a part of the logic, but less of the strictness for now.

ValenberghsSven:
Will be taking over this ticket and not sure when yet.
This branch is becoming stale so closing this for now.

PR #1815, opened 2023-07-06
remove decisionActivity relationship from piece

StuyckRuben:
add decisionActivity relationship to report
https://kanselarij.atlassian.net/browse/KAS-4159

### DIFF
--- a/app/models/piece.js
+++ b/app/models/piece.js
@@ -40,8 +40,6 @@ export default class Piece extends Model {
 
   @belongsTo('submission-activity', { inverse: 'pieces', async: true })
   submissionActivity;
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
We never do `piece.decisionActivity` so shouldn't really have any impact.
We do use it in queries
```
    const pieces = await this.store.query('piece', {
      'filter[decision-activity][subcase][:id:]': subcase.id,
```
This seems to be working with `piece` or `report` since it's polymorphic
Guessing it should be `report` ?

tdn (changes requested):
See Sven's comment. Make sure all queries using this relationship are adapted to `report` and still work as expected

tdn (approved):

PR #1816, opened 2023-07-06
Feature/kas 4183 pill unclickable in list

sergiofenoll:
Make the signature pill unclickable in the ongoing signatures list and make the row click open SH in a new tab.


ValenberghsSven (approved):
can't check actual links to SH myself, assuming they work

PR #1820, opened 2023-07-07
add argument to statusfilter to exclude certain statuses

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4185

### DIFF
--- a/app/components/signatures/status-filter.js
+++ b/app/components/signatures/status-filter.js
@@ -3,10 +3,15 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency';
-import { isPresent } from '@ember/utils';
+import { isPresent, isBlank } from '@ember/utils';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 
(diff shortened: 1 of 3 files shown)


app/components/signatures/status-filter.js
```diff
@@ -35,7 +40,9 @@ export default class SignaturesStatusFilterComponent extends Component {
           label: status.label,
           value: status.id
         };
-        this.statusFilterItems.push(statusItem);
+        if (isBlank(this.args.excludedStatuses) || this.args.excludedStatuses.indexOf(status.uri) == -1 ) {
+          this.statusFilterItems.push(statusItem);
+        }
         if (this.args.selectedStatusIds?.indexOf(statusItem.value) > -1) {
```
tdn: I'd rather see this whole `if` and `push` statement included under the `if (isBlank(this.args.excludedStatuses) || this.args.excludedStatuses.indexOf(status.uri) == -1 ) {` check, since this could theoretically add an excluded status to `selectedStatuses`. It won't do much in the actual app, if it happens at all, but it's just cleaner that way.

tdn (changes requested):
Small code style change requested. 

Everything else works and looks fine to me. Failing test seems to be unrelated to this change

tdn (approved):

PR #1821, opened 2023-07-11
Feature/kas 4104 solve deprecations

sergiofenoll:
Replaces the deprecated code that I encountered while going through the app manually. A lot of the files are touched because some of the deprecations were very wide-spread in our codebase.

The behaviour to throw on deprecations is left enabled so that we can catch any missed deprecations in the test runs. ~We should probably undo this change before merging though.~ We may want to leave the behaviour enabled so new deprecated code doesn't sneak in.

We might also want to add linter rules for the deprecated behaviour, so that we don't accidentally commit new deprecated code. Will have to look into that.

An overview of the deprecated code that has been replaced:
- Implicit injector by Ember simple auth → implicit injection is deprecated, setup needs to be manual now
- Extended array-like methods are deprecated 
  - `sortBy`: use built-in `sort`
  - `mapBy`: use built-in `map`
  - `filterBy`: use built-in `filter`
  - `findBy`: use built-in `filter`
  - `toArray`: use built-in `slice`
  - `pushObject`: use built-in `push`
  - `pushObjects`: use built-in `push`
  - `firstObject`: use `[0]` or `at(0)`
  - `lastObject`: use `at(-1)`
  - `objectAt`: use `at`
  - `addObject`: use `addObject` from `utils/array-helpers`
  - `addObjects`: use `addObjects` from `utils/array-helpers`
  - `removeObject`: use `removeObject` from `utils/array-helpers`
  - `removeObjects`: use `removeObjects` from `utils/array-helpers`
  - use a `TrackedArray` where you depend on `push`/`addObject`/`removeObject`
- don't use a `hasMany` relationship witouth awaiting it
- make polymorphic relationships explicit


ValenberghsSven:
there are 2 usages of `A()`. Does that need to be TrackedArray? one is a tracked (document-card), 1 is a local list (utils/documents)

sergiofenoll:
> there are 2 usages of `A()`. Does that need to be TrackedArray? one is a tracked (document-card), 1 is a local list (utils/documents)

I don't think it's strictly necessary because we don't use the array's reactiveness in either. In `utils/documents.js` it's just a list that we return (and in the single caller of that function we iterate over the resulting array and do nothing more with it).

In `document-card.js` we just iterate over the arrays and then copy them (`.slice()` is peppered throughout the component).

I'm fine with replacing them but I don't think it's strictly necessary.

ValenberghsSven:
Converted this to draft.
This branch has become too old to update and merge.

A better solution:
A PR for each change starting from the easiest.
After merge we need to recheck if new code introduces a new deprecation.
When we cover all aspects of a deprecation, we can enable the option in deprecation workflow.
Keeping this PR open to see how certain issues were solved and reuse that code.

I am going to work on this between reviews because I want this ticket done. (Ember 5 is out and we can't switch without this)

ValenberghsSven:
making a task list to keep track (not sure why this task list is so spaced, can't seem to fix it)


- [x] Implicit injector by Ember simple auth → implicit injection is deprecated, setup needs to be manual now

Extended array-like methods are deprecated:
- [x] sortBy: use built-in sort
- [x] mapBy: use built-in map
- [x]  filterBy: use built-in filter
- [x]  findBy: use built-in filter
- [x]  toArray: use built-in slice
- [x]  pushObject: use built-in push
- [x]  pushObjects: use built-in push
- [x]  firstObject: use [0] or at(0)
- [x]  lastObject: use at(-1)
- [x]  objectAt: use at
- [x]  addObject: use addObject from utils/array-helpers
- [x]  addObjects: use addObjects from utils/array-helpers
- [x]  removeObject: use removeObject from utils/array-helpers
- [x]  removeObjects: use removeObjects from utils/array-helpers
- [x]  use a TrackedArray where you depend on push/addObject/removeObject

- [x]  don't use a hasMany relationship witouth awaiting it
- [x]  make polymorphic relationships explicit
- [x]  Appuniversum deprecations
- [x]  add deprecation-workflow to prevent more known deprecations being added in new code

ValenberghsSven:
All tasks done, 99% of deprecations are gone, jenkins was green.
Closing this one in favor of all the smaller ones

PR #1822, opened 2023-07-11
modify present-for-signing now creates signflow instead of opening document-preview

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4191

### DIFF
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -221,15 +221,13 @@
                 {{/if}}
                 {{#if this.mayCreateSignMarkingActivity}}
                   {{#let (t "signatures") as |signTab| }}
-                    <AuLink
+                    <AuButton
                       @skin="link"
-                      @route="document"
-                      @model={{this.piece.id}}
-                      @query={{hash tab=signTab}}
+                      {{on "click" this.markDocumentForSigning}}
                       role="menuitem"
                     >
                       {{t "present-for-signing"}}
-                    </AuLink>
+                    </AuButton>
                     <AuHr />
                   {{/let}}
                 {{/if}}
--- a/app/components/documents/document-card.hbs
+++ b/app/components/documents/document-card.hbs
@@ -220,16 +220,14 @@
                   </AuButton>
                 {{/if}}
                 {{#if this.mayCreateSignMarkingActivity}}
-                  {{#let (t "signatures") as |signTab| }}
-                    <AuButton
-                      @skin="link"
-                      {{on "click" this.markDocumentForSigning}}
-                      role="menuitem"
-                    >
-                      {{t "present-for-signing"}}
-                    </AuButton>
-                    <AuHr />
-                  {{/let}}
+                  <AuButton
+                    @skin="link"
+                    {{on "click" this.markDocumentForSigning}}
+                    role="menuitem"
+                  >
+                    {{t "present-for-signing"}}
+                  </AuButton>
+                  <AuHr />
                 {{/if}}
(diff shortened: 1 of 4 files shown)


tdn (changes requested):
I tested this, and the `document-card` doesn't refresh automatically when clicking the "Aanbieden voor handtekenen" button. The `signFlow` is created, but the card, pill and actions remain the same, so the action can be clicked again, document can be removed, etc. 

This is a blocking change for this to be merged

ValenberghsSven:
@tdn I'll take over because my ticket requires this to continue.

ValenberghsSven (approved):
Made 1 small change (awaiting async code and reloading data after), rest was good

ValenberghsSven:
Hmm, on second though, the "present for signing" does not work on agenda documents (shouldn't) and subcases (shouldn't if there are multiple decision-activities when postponed/reproposed)
Since we used to link to doc preview that would always work, but now we should hide present for signing if there is no decision-activity (for now), going to add that check to `mayCreateSignMarkingActivity`

ValenberghsSven:
@tdn final approval?

app/templates/agenda/agendaitems/agendaitem/documents.hbs
```diff
@@ -44,6 +44,7 @@
       @onAddPiece={{perform this.addPiece}}
       @didDeleteContainer={{this.refresh}}
       @isEditable={{true}}
+      @decisionActivity={{this.decisionActivity}}
```
sergiofenoll, line 47: Should we also add this to `app/templates/cases/case/subcases/subcase/documents.hbs`?

sergiofenoll: I'm guessing we don't need to, I assume most of the document editing will happen from the agendaitem view & it's trivial to navigate there from the subcase view.

ValenberghsSven: I wouldn't, there can be more than 1 `decisionActivity` in this view if a subcase was postponed/reproposed.
Unless we can always assume this means the latest one...
Smells risky

We could offer the option if there is only exactly 1.
And not offer the option if there are 0 (not proposed yet) or more than 1 (proposed again)

But that could also mean that the subcase was postponed but not yet reproposed (also 1 `decisionActivity`)
so we could be marking the documents for signatures on the wrong `decisionActivity` if they attempt to mark before reproposing

ValenberghsSven: Right now, we don't offer marking for signature in the batch edit on the subcase documents view.
So somewhere we already decided not to allow that in this view?

sergiofenoll: Ok, perfect, I used that same assumption in my own PR. We should find a way to encode/share these kinds of  assumptions, but that's a larger issue in our codebase than just here. Thanks for the info

app/components/documents/document-card.hbs
```diff
@@ -221,15 +221,13 @@
                 {{/if}}
                 {{#if this.mayCreateSignMarkingActivity}}
                   {{#let (t "signatures") as |signTab| }}
```
sergiofenoll: I would remove this `{{#let}}}` since we no longer need it.

sergiofenoll (approved):

app/components/documents/document-card.hbs
ValenberghsSven: true

PR #1824, opened 2023-07-12
trim piece name on save

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4166

### DIFF
--- a/app/components/documents/batch-details/batch-documents-details-modal.js
+++ b/app/components/documents/batch-details/batch-documents-details-modal.js
@@ -116,7 +116,7 @@ export default class BatchDocumentsDetailsModal extends Component {
         }
         // container is cleaned up in deletePiece if no more pieces exist
       } else {
-        piece.name = row.name;
+        piece.name = row.name?.trim();
         // does not check for relationship changes
         let accessLevelHasChanged = false;
         let hasChanged = piece.dirtyType === 'updated';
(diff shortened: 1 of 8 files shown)


sergiofenoll (changes requested):
The same change you applied to `app/controllers/agenda/documents.js` also needs to occur in:

- `app/controllers/agenda/agendaitems/agendaitem/documents.js` → `savePiece`
- `app/controllers/cases/case/subcases/subcase/documents.js` → `savePiece`
- `app/components/documents/document-card.js` → `addPiece`

We might want to just add the trim to the `piece` serializer since that will cover all our cases (in the frontend at least), but we can do that down the line.

ValenberghsSven (changes requested):
- just realized we allow empty text as name for documents in some cases, seems wrong and we should fix that later. perhaps not use the optional parameter and just except that there should be a text? I suggested the optional myself but this name should not be optional in hindsight
- There are a place where we save `piece` but name is not trimmed.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/documents/batch-details/batch-documents-details-modal.js#L119

Also here where we get the name from the previous version (which can be untrimmed)
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/documents/document-card.js#L198
However, this seems to be covered by the `addPiece` in this component where you added it. Can you verify? (manually change the name with the ember-inspector and save, then upload bis)

ValenberghsSven (approved):

PR #1826, opened 2023-07-12
KAS-4206 Added keepMarkingActivity argument to removeSignflow

tdn:
HOTFIX

Keep the marking activity after trying to send a signFlow to SigningHub. Right now, in case of an error on their end, our signFlow gets totally removed, which is confusing for the users. This fix keeps it in the list by removing all the `activities` and `creator`, and keeping the `signFlow`, `signSubcase`, `signMarkingActivity` & resetting the `status`.


ValenberghsSven (approved):
pair reviewed

PR #1829, opened 2023-07-12
Feature/kas 4199 editable pieces with sign flows

sergiofenoll:
Show the dropdown for pieces that have a MARKED sign flow. Delete and re-creates sign flows as needed.

### DIFF
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -241,6 +241,7 @@ export default class DocumentsDocumentCardComponent extends Component {
     }
 
     try {
+      this.newPiece.name = this.newPiece.name.trim();
       yield this.args.onAddPiece(this.newPiece, this.signFlow);
       this.pieceAccessLevelService.updatePreviousAccessLevel(this.newPiece);
       this.loadVersionHistory.perform();
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -214,7 +214,7 @@ export default class DocumentsDocumentCardComponent extends Component {
   @task
   *addPiece() {
     try {
-      yield this.args.onAddPiece(this.newPiece);
+      yield this.args.onAddPiece(this.newPiece, this.signFlow);
       this.pieceAccessLevelService.updatePreviousAccessLevel(this.newPiece);
       this.loadVersionHistory.perform();
       this.newPiece = null;
@@ -273,6 +273,9 @@ export default class DocumentsDocumentCardComponent extends Component {
   @task
   *deleteDocumentContainerWithUndo() {
     yield timeout(DOCUMENT_DELETE_UNDO_TIME_MS);
+    if (this.signFlow) {
+      yield this.signatureService.removeSignFlow(this.signFlow);
+    }
     yield deleteDocumentContainer(this.documentContainer);
     this.args.didDeleteContainer?.(this.documentContainer);
   }
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -25,6 +25,7 @@ export default class DocumentsDocumentCardComponent extends Component {
    * @argument onAddPiece: action triggered when a new version has been added
    * @argument bordered: determines if the card has a border
    * @argument label: used to determine what label should be used with the date
+   * @argument decisionActivity: if a decision-activity is linked (specifically via agenda-item-treatment)
    */
   @service store;
   @service router;
@@ -74,7 +75,8 @@ export default class DocumentsDocumentCardComponent extends Component {
   get mayCreateSignMarkingActivity() {
     return !this.signMarkingActivity
       && this.signaturesEnabled
-      && this.currentSession.may('manage-signatures');
+      && this.currentSession.may('manage-signatures')
+      && !!this.args.decisionActivity;
   }
 
   get mayShowEditDropdown() {
@@ -342,6 +344,12 @@ export default class DocumentsDocumentCardComponent extends Component {
 
   @action
   async reloadAccessLevel() {
+    await this.loadPieceRelatedData.perform();
+  }
+
+  @action
+  async markDocumentForSigning() {
+    await this.signatureService.markDocumentForSignature(this.piece, this.args.decisionActivity);
     await this.loadPieceRelatedData.perform();
   }
 
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -83,8 +83,15 @@ export default class DocumentsDocumentCardComponent extends Component {
       && this.currentSession.may('manage-documents')
       && this.loadSignatureRelatedData.isIdle
       && this.loadSignatureRelatedData.performCount > 0
-      && (!this.hasSignFlow
-          || (this.hasMarkedSignFlow && !!this.args.decisionActivity))
+      && (!this.hasSignFlow || this.hasMarkedSignFlow)
+    );
+  }
+
+  get mayShowUploadNewVersion() {
+    return (
+      !this.args.hideUpload
+        && (!this.hasSignFlow
+            || (this.hasMarkedSignFlow && !!this.args.decisionActivity))
     );
   }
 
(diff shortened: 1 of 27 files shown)


app/components/documents/document-card.js
```diff
@@ -74,6 +76,17 @@ export default class DocumentsDocumentCardComponent extends Component {
...
   }
 
+  get mayShowEditDropdown() {
+    return (
+      this.args.isEditable
+      && this.currentSession.may('manage-documents')
+      && this.loadSignatureRelatedData.isIdle
+      && this.loadSignatureRelatedData.performCount > 0
+      && (!this.hasSignFlow
+          || (this.hasMarkedSignFlow && !!this.args.decisionActivity))
```
sergiofenoll: This condition is to stop us from showing the dropdown in the `subcase/documents` route. Since we don't have a decisionActivity there, we can't do `Nieuwe versie opladen`, because we can't re-create the sign flow.

We might instead want to only show the options that are possible to execute (`Wijzigen` and `Document verwijderen`).

translations/nl-be.json
```diff
@@ -1248,5 +1248,6 @@
   "delete-all-sign-flow-tooltip": "Verwijder het volledige handtekenproces uit Kaleidos, zodat het document opnieuw ondertekend kan worden",
   "to-start-up": "Op te starten",
   "to-follow-up": "Op te volgen",
-  "stop-sign-flow": "Tekenen stopzetten"
+  "stop-sign-flow": "Tekenen stopzetten",
+  "sign-flow-was-sent-while-you-were-editing": "De sign flow is reeds verstuurd naar de Handtekenmap terwijl u probeerde te editeren."
```
ValenberghsSven: This translation is most likely too technical to show to the user.
"Dit document is reeds verstuurd ... "?
We also don't specifiy that because of this we block the upload of a new version of a document

ValenberghsSven (approved):
Code looks fine, have not tested it manually but jenkins is green so that helps.

Minor comment on the translation but is not blocking for approval

translations/nl-be.json
sergiofenoll: The toast title should be `Je wijzigingen kunnen niet opgeslagen worden.`, not sure if that's enough for users, but maybe explicitly telling them that uploads/edits/deletes weren't saved is better. Will change and make it simpler.

app/components/documents/document-card.js
tdn: Yes, that's probably a good idea, since the users were not happy about "losing" their features.

sergiofenoll: :+1: Changes made, though thinking about it I guess we could still show `Nieuwe versie opladen` in the subcase route by getting the decisionActivity from the old sign flow (before deleting it) and then re-using it to create the new one.

I can make that change too if desired, otherwise we can keep it in the back of our minds if the question ever comes from the users.

ValenberghsSven: I would wait for an explicit ask from the users.
I've heard numerous times that "the agenda view" is where secretary will be marking the documents.
If the marking has been toggled then that would have happened in the agenda views.
It makes more sense that adding new versions of documents with markings would also be done in the agenda views.
No dropdown/actions on subcase views docs when a markingActivity is present for now seems ok.

edit:
Edit document and delete document container allowed should be ok when it is only marked.

app/components/documents/document-card.js
```diff
@@ -202,8 +226,21 @@ export default class DocumentsDocumentCardComponent extends Component {
 
   @task
   *addPiece() {
+    if (this.signFlow) {
+      const status = yield this.signFlow.belongsTo('status').reload();
+      if (status.uri !== CONSTANTS.SIGNFLOW_STATUSES.MARKED) {
+        yield this.deleteUploadedPiece.perform();
+        this.router.refresh();
```
ValenberghsSven: A refresh from component? Should be the responsibility of controllers/routes.
Also, why are we refreshing here?
We could just do `deleteUploadedPiece` and show the warning.

app/components/documents/document-card.js
```diff
@@ -241,7 +278,19 @@ export default class DocumentsDocumentCardComponent extends Component {
   }
 
   @action
-  verifyDeleteDocumentContainer() {
+  async verifyDeleteDocumentContainer() {
+    if (this.signFlow) {
+      const status = await this.signFlow.belongsTo('status').reload();
+      if (status.uri !== CONSTANTS.SIGNFLOW_STATUSES.MARKED) {
+        this.router.refresh();
```
ValenberghsSven: same here

app/components/documents/document-card/edit-modal.js
```diff
@@ -92,6 +94,19 @@ export default class DocumentsDocumentCardEditModalComponent extends Component {
   }
 
   saveEdit = task(async () => {
+    if (this.args.signFlow) {
+      const status = await this.args.signFlow.belongsTo('status').reload();
+      if (status.uri !== CONSTANTS.SIGNFLOW_STATUSES.MARKED) {
+        this.router.refresh();
```
ValenberghsSven: just call the `onCancel` and call it a day? no router refresh needed in component

ValenberghsSven (commented):
Missed it on my first run through. Refresh from component should not be needed.
Should be the concern of the controller with callbacks if that is needed.

app/components/documents/document-card.js
sergiofenoll: Refresh aren't strictly needed, but it's to force the user to see the page with the new data. To prevent them mindlessly dismissing the toast and just keep trying the action and such. As long as the page isn't refreshed we're still in a stale/inconsistent state.

app/services/signature-service.js
```diff
@@ -206,6 +206,18 @@ export default class SignatureService extends Service {
     return false;
   }
 
+  async hasMarkedSignFlow(piece) {
+    const signaturesEnabled = !!ENV.APP.ENABLE_SIGNATURES;
+    if (signaturesEnabled) {
```
tdn, line 211: I realize this was already the case in `hasSignFlow`, but just a consideration I'm making now: if we turn the feature flag off after we've created some `signFlows`. That will mean all edit functionalities become available again, and we could end up with `signFlows` that have different/no pieces in de database.

Not something to fix in this PR, but something to remember if we ever enable and then disable the feature flag: we need a "cleanup" function to check for inconsistent `signFlows`.

tdn (changes requested):
This fixes the issue for the document card, but not for the "Bulk edit" modal or for the DocumentPreviewSidebar.

As I understand it, @ValenberghsSven is taking care of the "Bulk edit" functionalities already due to conflicts with his PR.
He told me he'll pick both of these up.

app/components/documents/document-card.js
ValenberghsSven: tested this locally and the only issue is that the actions dropdown stays visible after we reloaded the status.
So running `loadSignatureRelatedData` again is enough (only 1 card needs to be refreshed, not the entire route)

app/services/signature-service.js
ValenberghsSven: if we are willingly turning off the feature flag for any reason we should just do a full cleanup of the signing graph + data?
In time the feature flag should be removed and the feature fully integrated in the app.

ValenberghsSven: If we have this feature running on prod once we should never disable it in my opinion.
They can just choose not to use it at that stage (but automatic things will keep happening on marked pieces)

app/components/documents/document-card.js
```diff
@@ -311,4 +365,11 @@ export default class DocumentsDocumentCardComponent extends Component {
     }
     return false;
   }
+
+  @action
+  async cancelEditPiece() {
+    await this.loadPieceRelatedData.perform();
```
ValenberghsSven, line 372: this `loadPieceRelatedData` already contains the `loadSignatureRelatedData`.
Just noticed that now.
It might be better to just load that one. Will adjust the other ones I just added

app/components/documents/document-card/edit-modal.js
ValenberghsSven: called local `this.cancelEdit()` which also deletes any uploaded files and call `args.onCancel`

ValenberghsSven:
did not fully test the edit/deletion of document route with a marked piece / concurrent status change.

tdn:
Closing this since this branch was merged into https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1831

Will be merged into development from there

PR #1830, opened 2023-07-13
edit filter on 'op te starten' to check if decision are released and approved

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4198


tdn (approved):
Pair-reviewed

PR #1832, opened 2023-07-13
Feature/kas 4076 old mandatees for organisations

kobemertens:


sergiofenoll (approved):
Nice :+1: Tested locally, works as advertised :)

About the selector being slow: part of it is that we're always creating a new fresh request (because we pass in `this.refreshDate`, which is `new Date()`). If we would send something like `import { startOfDay } from 'date-fns'; startOfDay(new Date())` instead, we at least would have the same request being sent for a day, which could help with caching.
**But** that would still not resolve the issue because every time we set the link between the mandatee and a user-organization, we're clearing the cache key for the mandatees (because we've just mutated a mandatee record). So even if we had cacheable calls, they would immediately get busted as soon as we save.

tdn (approved):

PR #1835, opened 2023-07-14
added updatePreviousAccessLevels when creating new version of report 

StuyckRuben:
added strengthenAccessLevelToInternRegering for the previousVersions when decisionResult is updated to UITGESTELD or INGETROKKEN

https://kanselarij.atlassian.net/browse/KAS-4208


app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -149,6 +150,13 @@ export default class AgendaitemDecisionComponent extends Component {
           piece
         );
       }
+      let reportToUpdate = this.report;
+      while (reportToUpdate) {
+        await this.pieceAccessLevelService.strengthenAccessLevelToInternRegering(
+          reportToUpdate
+        );
+        reportToUpdate = await reportToUpdate.previousPiece
+      }
```
sergiofenoll: Is the `while` loop necessary? `strengthenAccessLevelToInternRegering` should already update the previous versions.

sergiofenoll (changes requested):

ValenberghsSven (changes requested):
Looking at this I might have made mistake when creating this ticket.
I said that the report needs to be updated when changing the decision-result-code.
However I am not sure this is correct.
Trying to puzzle together why I thought that.
in this ticket it is clearly stated that the access rights on reports stay unchanged
https://kanselarij.atlassian.net/browse/KAS-2008

Also mentioned here that report rights should stay unchanged
https://kanselarij.atlassian.net/browse/KAS-3826

We do need to update when adding BIS
https://kanselarij.atlassian.net/browse/KAS-3938

I might had it confused with setting confidential > access levels of report should become confidential.
`updateDecisionsAccessLevelOfSubcase`
But that feature should already be ok and also updates previous versions. Going to test that

app/components/agenda/agendaitem/agendaitem-decision.js
ValenberghsSven: see my comment. I think this entire block should go.
Report access levels stay unchanged (but there should be a difference between the original report and a BIS report)

ValenberghsSven (approved):
Bit biased, but only 1 line needs to be added.
Tested it and it works.

PR #1839, opened 2023-07-17
removed label argument on DocumentCard

StuyckRuben:
add functionality to DocumentCard to check what label should be if no label argument present

https://kanselarij.atlassian.net/browse/KAS-4170


app/components/documents/document-card.js
```diff
@@ -125,12 +127,26 @@ export default class DocumentsDocumentCardComponent extends Component {
         'filter[:id:]': id,
         include: 'document-container,document-container.type,access-level',
       });
-
+    const loadPiecePart = (id) => 
+      this.store.queryOne('piece-part', {
+        'filter[report][:id:]': id,
+      });
```
ValenberghsSven: this only covers pieces from generated decision report.
pieces from generated minutes will not have this property.

app/components/documents/document-card.js
```diff
@@ -125,12 +127,26 @@ export default class DocumentsDocumentCardComponent extends Component {
...
     if (this.args.piece) {
       this.piece = this.args.piece; // Assign what we already have, so that can be rendered already
       this.piece = yield loadPiece(this.piece.id);
       this.documentContainer = yield this.piece.documentContainer;
       yield this.loadVersionHistory.perform();
+      if (isPresent(this.args.label)) {
+        this.label = this.args.label;
+      } else {
+        const piecePart = yield loadPiecePart(this.piece.id);
+        console.log(piecePart);
```
ValenberghsSven: `console.log` still in here

app/components/documents/document-card.js
```diff
@@ -125,12 +127,26 @@ export default class DocumentsDocumentCardComponent extends Component {
...
+      });
     if (this.args.piece) {
       this.piece = this.args.piece; // Assign what we already have, so that can be rendered already
       this.piece = yield loadPiece(this.piece.id);
       this.documentContainer = yield this.piece.documentContainer;
       yield this.loadVersionHistory.perform();
+      if (isPresent(this.args.label)) {
+        this.label = this.args.label;
+      } else {
+        const piecePart = yield loadPiecePart(this.piece.id);
```
ValenberghsSven: This will run on every document-card.
Even on those were we know in advance that there will never be a piece-part (like agendaitem documents)

Assuming we have to query twice (once for report, once for minutes) this will increase the general query load.
Assuming only pieces that are generated are of subtype "report" or "minutes"  may have pieceparts.
Perhaps we could check that? 
`this.args.piece.constructor.modelName === 'report' || this.args.piece.constructor.modelName === 'minutes'` > 
check for pieceParts
In case of `report`, legacy decision reports have the type "report" but have no pieceParts so extra query is needed
In case of `minutes`, there is no legacy so can we assume that if we find that type it will always be generated? if so we could skip the extra query here.

app/components/documents/document-card.js
```diff
@@ -125,12 +127,26 @@ export default class DocumentsDocumentCardComponent extends Component {
...
+        console.log(piecePart);
+        if (piecePart) {
+          this.label = this.intl.t('created-on');
+        } else {
+          this.label = this.intl.t('uploaded-at');
+        }
```
ValenberghsSven: Check the `altLabel` comment.
We already have the default 'uploaded-at' in the `label` getter.

We should only set an alternative here when needed and fall back on the default (or args.label) in other cases.

app/components/documents/document-card.js
```diff
@@ -52,6 +52,8 @@ export default class DocumentsDocumentCardComponent extends Component {
   @tracked hasSignFlow = false;
   @tracked hasMarkedSignFlow = false;
 
+  @tracked label;
+
```
ValenberghsSven: there is already a getter called `label`.
```
  get label() {
    if (isPresent(this.args.label)){
      return this.intl.t(this.args.label);
    }
    return this.intl.t('uploaded-at');
  }
```
So unsure what ember is going to get, this new tracked value or the getter.
I would opt to call this tracked `altLabel` and only fill it when needed

ValenberghsSven (changes requested):
Made many comments.
To summarize:
```
@tracked altLabel;

  get label() {
    if (isPresent(this.args.label)){
      return this.intl.t(this.args.label);
    }
    if (isPresent(this.altLabel)){
      return this.altLabel;
    }
    return this.intl.t('uploaded-at');
  }
  
  *loadPieceRelatedData(){
  ....
      const loadReportPiecePart = (id) => 
      this.store.queryOne('piece-part', {
        'filter[report][:id:]': id,
      });
    const loadMinutesPiecePart = (id) => 
      this.store.queryOne('piece-part', {
        'filter[minutes][:id:]': id,
      });
      if (this.args.piece) {
      ......
  
      // check for alternative label
      const modelName = this.args.piece.constructor.modelName;
      if (!isPresent(this.args.label)) {
        let piecePart;
        if (modelName === 'report') {
          piecePart = yield loadReportPiecePart(this.piece.id);
        } else if (modelName === 'minutes') {
          piecePart = yield loadMinutesPiecePart(this.piece.id);
        }
        this.altLabel = piecePart ?  this.intl.t('created-on') : null;
      }
```
Not sure if it would be better to just return `'created-on'` and do the translation in the getter (where we do the translation for the other labels)

The only downside is that we show "uploaded-at" for a moment while loading and than changes to "created-on", which could be solved later if that is an issue.

sergiofenoll (approved):
I think the usage of `label` is a bit confusing here ("Aangemaakt op" or "Geüpload op" aren't really the label of the card) and I wonder if we aren't just overcomplicating this (simply using an alternative like "Toegevoegd op" covers both cases of creation and uploading and doesn't force us to fetch data and add extra logic).

But given the requirements of the ticket this looks fine.

PR #1842, opened 2023-07-18
fix: don't reload whole agendaitems route when searching

sergiofenoll:
Full reload (and showing the loading state) means that we de-render the search field and that users can't automatically keep typing.

This commit should still preserve the fix intended in KAS-4129, though without making use of router.transitionTo and just using the refreshModel behaviour of the queryParams again. Setting filter as a tracked field instead of using `set()` seems to fix the behaviour.

https://kanselarij.atlassian.net/browse/KAS-4129

Somewhat undoes https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1796


tdn (approved):
Works well locally

PR #1844, opened 2023-07-24
notified on signSubcase and approver on signApprovalActivity now saved as lowercase

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4220

### DIFF
--- a/app/components/signatures/create-sign-flow.js
+++ b/app/components/signatures/create-sign-flow.js
@@ -111,7 +111,7 @@ export default class SignaturesCreateSignFlowComponent extends Component {
 
   @action
   saveApprover(approver) {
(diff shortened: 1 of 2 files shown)


tdn (approved):
Added the conversion to lowercase a bit earlier, so the user sees the same email address as the one that will be in SigningHub.

Did leave the extra conversion as well in the service, since that adds an extra level of safety.

ValenberghsSven (approved):
did not manually test, but code looks pretty straightforward.
I could argue we could do a lowercase in the `<Signatures::EmailModal>` but a final `toLowerCase` when creating the actual model seems most safe in most cases anyway.

PR #1846, opened 2023-07-24
cypress/KAS-4146-publication-reports-time-difference-in-CI

MathiasZabot:


ValenberghsSven:
Changed it a bit to make naming more clear.
instead of checking 1 set date with formatting we now regex 3 dates.
So if the date value is yesterday, today (09-08-2023) or tomorrow it should still pass.
Should also work when the month/year changes around time of testing
image

ValenberghsSven (approved):
Can merge when tests are green.
In order to fully test we would need to time the test suite so it runs during midnight.

PR #1847, opened 2023-07-28
Feature/kas 1908 secretaris instellen

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-1908

- [ ] merge https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/394


kobemertens:
Should be ready for review

app/components/mandatees/secretary-panel/edit.js
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * @argument onCancel
+   * @argument onSave
+   * @argument {Date} referenceDate: Date to get active Mandatees for
+   * @argument secretary
+   */
+
+  @tracked secretary;
+
+  VISIBLE_ROLES = [
```
sergiofenoll: We generally don't make members (variables inside a class) uppercase.

app/components/mandatees/secretary-panel/edit.js
```diff
@@ -0,0 +1,39 @@
...
+
+  save = task(async () => {
+    await this.args.onSave?.(this.secretary);
+    this.isEditing = false;
+  });
+
+  @action
+  cancel() {
+    this.args.onCancel?.();
+  }
```
sergiofenoll: This seems unused.

app/components/mandatees/secretary-panel/edit.js
```diff
@@ -0,0 +1,39 @@
...
+
+  @tracked secretary;
+
+  VISIBLE_ROLES = [
+    CONSTANTS.MANDATE_ROLES.SECRETARIS,
+    CONSTANTS.MANDATE_ROLES.WAARNEMEND_SECRETARIS,
+  ];
+
+  save = task(async () => {
+    await this.args.onSave?.(this.secretary);
```
sergiofenoll, line 22: If you save a secretary, then edit again and don't select anything (the dropdown will show the saved secretary) and then save, you overwrite the secretary and empty it.

Is this desired? From a "what you see" I perspective I wouldn't say so, as the UI makes it looks like you're just re-saving the old secretary.

Either disable saving if no secretary is selected, or pass `this.selectedSecretary` here (to re-save the already saved secretary).

app/components/meeting/edit-meeting-modal.js
```diff
@@ -30,10 +33,16 @@ export default class MeetingEditMeetingComponent extends Component {
   @tracked documentPublicationActivity;
   @tracked themisPublicationActivity;
   @tracked plannedDocumentPublicationDate;
+  @tracked secretary;
   @tracked extraInfo;
   @tracked _meetingNumber;
   @tracked _numberRepresentation;
 
+  VISIBLE_ROLES = [
```
sergiofenoll: Same as above, shouldn't be uppercased.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -128,51 +163,78 @@ export default class MeetingEditMeetingComponent extends Component {
...
     this.args.meeting.extraInfo = this.extraInfo;
+    this.args.meeting.secretary = this.secretary;
     this.args.meeting.plannedStart = this.startDate || now;
     this.args.meeting.kind = this.selectedKind;
     this.args.meeting.number = this.meetingNumber;
     this.args.meeting.numberRepresentation = this.numberRepresentation;
     this.args.meeting.mainMeeting = this.selectedMainMeeting;
 
+    if (this.args.meeting.secretary != this.secretary) {
+      const decisionActivities = yield this.store.query('decision-activity', {
```
sergiofenoll: We might want to make this a call to `queryAll` here w.r.t. pagination. A regular query will return max 20 items IIRC, and we might encounter a lot more results which we would miss without paginating. `queryAll` does pagination behind the scenes to get all items.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -128,51 +163,78 @@ export default class MeetingEditMeetingComponent extends Component {
...
+        'filter[treatment][agendaitems][agenda][created-for][:id:]':
+          this.args.meeting.id,
+      });
+      for (let decisionActivity of decisionActivities.toArray()) {
```
sergiofenoll: Use `.slice()` instead of `.toArray()`, the latter is deprecated.

app/controllers/settings/users/user.js
```diff
@@ -14,6 +14,15 @@ export default class SettingsUsersUserController extends Controller {
   @tracked showUnblockMembership = false;
   @tracked selectedPerson = null;
 
+  VISIBLE_ROLES = [
```
sergiofenoll: As before, don't uppercase members.

app/templates/agenda/agendaitems/agendaitem/index.hbs
```diff
@@ -69,4 +69,10 @@
     @allowEditing={{user-may "manage-agendaitems"}}
     @onSave={{this.saveGovernmentAreas}}
   />
+  <Mandatees::SecretaryPanel
```
sergiofenoll: This is only inside the `if this.subcase` branch, for the first agendaitem ("Goedkeuring van het verslag") we won't show this. Is this desired? I think in theory they might still want to update the secretary.

sergiofenoll (changes requested):

app/components/mandatees/secretary-panel/edit.js
kobemertens: fixed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/components/mandatees/secretary-panel/edit.js
kobemertens: removed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/components/meeting/edit-meeting-modal.js
kobemertens: fixed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/components/meeting/edit-meeting-modal.js
kobemertens: fixed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/components/meeting/edit-meeting-modal.js
kobemertens: fixed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/controllers/settings/users/user.js
kobemertens: fixed in [23025ec](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/23025ecee59adad6ba602b19fac0b45a84baa3ec)

app/templates/agenda/agendaitems/agendaitem/index.hbs
kobemertens: Not sure, @tdn what do you think?

app/components/mandatees/secretary-panel/edit.js
kobemertens: Disabled save button in [df093bc](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1847/commits/df093bcadb6f16be1aedf478aed3894db6b6f07d)

sergiofenoll (approved):
Waiting on backend changes, but this can be merged!

app/serializers/user.js
```diff
@@ -1,6 +1,11 @@
 import ApplicationSerializer from './application';
 
-const SKIP_SERIALIZED = ['loginActivity', 'account', 'person'];
+const SKIP_SERIALIZED = [
+  'loginActivity',
+  'account',
+  'secretaryForAgendas',
+  'secretaryForDecisions',
+];
```
tdn: Are we sure `person` doesn't need to be skipped here? (Not sure what this does actually)

app/components/subcases/subcase-header.js
```diff
@@ -105,6 +105,11 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
   *proposeForAgenda(meeting) {
     this.isAssigningToOtherAgenda = false;
     this.isLoading = true;
+    let decisionActivity = yield this.store.queryOne('decision-activity', {
+      'filter[subcase][:id:]': this.args.subcase.id,
+    });
+    decisionActivity.secretary = meeting.secretary;
```
tdn: This throws `Uncaught TypeError: Cannot set properties of null (setting 'secretary')` when the secretary of the `meeting` is not set, which can happen for any meeting that was created before these changes.

tdn: It complains because decisionActivity is `null`, which makes sense, since this `subcase` hasn't been proposed for an `agenda` yet. 

Not sure when `decisionActivity` might ever not be null actually... 

I'd be inclined to remove this added code block completely.

app/services/agenda-service.js
```diff
@@ -128,11 +129,15 @@ export default class AgendaService extends Service {
...
 
+    // default secretary
+    const currentApplicationSecretary = await this.mandatees.getCurrentApplicationSecretary();
+
     // decision-activity
     const decisionActivity = await this.store.createRecord('decision-activity', {
       subcase,
       startDate: meeting.plannedStart,
       decisionResultCode,
+      secretary: currentApplicationSecretary
```
tdn: We actually want to check whether the `meeting` has a `secretary` set, that might differ from the `currentApplicationSecretary`. Then we'd use the `currentApplicationSecretary` if the `meeting.secretary` is null (which it can't be for meetings created after this change, but we want to cover this edge case for older agendas).

tdn (changes requested):
Needs a few small changes to cover edge cases (see comments)

tdn:
I would also be inclined to merge https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1863 first, and then re-merge development into this branch.

app/serializers/user.js
tdn: As correctly pointed out by @sergiofenoll , user has no `secretaryForAgendas` nor `secretaryForDecision` relation

tdn (approved):
Implemented the changes, and seems to work fine now.

checked edge cases:
- add subcase to agenda without secretary -> application secretary is chosen => OK
- add subcase to agenda with different secretary than default -> meeting secretary is chosen => OK
- change meeting secretary -> all agendaitems/decisionActivities are updated to this new secretary => OK

So for me, this can be merged!

PR #1849, opened 2023-08-03
Feature/kas 4193 secretaris toevoegen aan notulen editor

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4193

- [x] first merge https://github.com/kanselarij-vlaanderen/minutes-report-generation-service/pull/1


app/routes/agenda/minutes.js
```diff
@@ -47,7 +51,7 @@ export default class AgendaMinutesRoute extends Route {
       }
     }
     const meetingMinutes = await meeting.minutes;
-    return { minutes: meetingMinutes, mandatees, notas, announcements };
+    return { minutes: meetingMinutes, mandatees, notas, announcements, secretary };
```
sergiofenoll: Fetching the secretary in the route isn't a good idea, because it can be updated without reloading the route.

- Click on "Inhoud updaten" of the meeting minutes of an agenda, observe that the current secretary is added
- Click on "Acties" > "Vergadering wijzigen"
- Update the secretary of the meeting
- Click on "Inhoud updaten" again, this time the secretary will not be updated
- Refresh the page
- Click on "Inhoud updaten", now the correct secretary is set

Instead of passing the secretary from the route, fetch the secretary when you're updating the contents of the meeting minutes, that way we're always using the "fresh" secretary.

app/controllers/agenda/minutes.js
```diff
@@ -114,11 +113,12 @@ function getAttendees(model) {
     .map(mandateeName)
     .filter((x) => !seen.has(x));
 
+  const secretary = mandateeName(model.secretary);
```
sergiofenoll: This should probably deal with a secretary that's null/undefined (for existing agenda's where no secretary is set).

```js
let secretaryName = '';
const secretary = /*fetch secretary here*/;
if (secretary) {
  secretaryName = mandateeName(secretary);
}

// ....

return {
  primeMinister,
  viceMinisters,
  ministers,
  secretary: secretaryName,
};

sergiofenoll (changes requested):

sergiofenoll (approved):

tdn (approved):
Made a small change for an edge case where there is no secretary, but works fine otherwise.

One remark: there is no automatization yet when the secretary of the meeting changes. I.e., the minutes are not automatically updated. This is fine for this story, but might need to revisit this in the future

PR #1850, opened 2023-08-04
SignaturePill now conform except for signed pill

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4215

### DIFF
--- a/app/components/signature-pill.js
+++ b/app/components/signature-pill.js
@@ -25,10 +25,7 @@ export default class SignaturePillComponent extends Component {
   get isClickable() {
     const signingHubUrl = this.data?.value?.signingHubUrl;
     const route = this.data?.value?.route;
-    if (signingHubUrl || route) {
-      return true;
-    } 
-    return false;
+    return !!signingHubUrl || !!route;
   }
 
   willDestroy() {
--- a/app/components/signature-pill.js
+++ b/app/components/signature-pill.js
@@ -12,7 +12,6 @@ const { SIGNED, REFUSED, CANCELED, MARKED } = constants.SIGNFLOW_STATUSES;
 /**
  * @param signMarkingActivity {SignMarkingActivityModel|Promise<SignMarkingActivityModel>}
  * @param piece {Piece|Promise<Piece>}
- * @param isClickable {boolean} Defaults to true, can be used to disable the click behaviour
  */
 export default class SignaturePillComponent extends Component {
   @service intl;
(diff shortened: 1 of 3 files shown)


app/components/signature-pill.js
```diff
@@ -24,8 +23,12 @@ export default class SignaturePillComponent extends Component {
...
 
   get isClickable() {
-    // use passed in argument, otherwise default to true
-    return this.args.isClickable ?? true;
+    const statusUri = this.data.value?.status?.uri;
+    const signingHubUrl = this.data.value?.signingHubUrl;
+    if (statusUri === REFUSED || statusUri === CANCELED || statusUri === SIGNED || !signingHubUrl) {
+      return false;
+    } 
+    return true;
```
tdn: I always get a bell ringing in my head when a `true` or `false` is returned after an `if` statement:

In my opinion, it's cleaner to return the condition as a boolean directly then. In this case: 
```
return signingHubUrl && statusUri !== REFUSED && statusUri !== CANCELED && statusUri !== SIGNED; 
```

app/components/signature-pill.js
```diff
@@ -24,8 +23,12 @@ export default class SignaturePillComponent extends Component {
   @tracked triggerTask;
 
   get isClickable() {
-    // use passed in argument, otherwise default to true
-    return this.args.isClickable ?? true;
+    const statusUri = this.data.value?.status?.uri;
+    const signingHubUrl = this.data.value?.signingHubUrl;
+    if (statusUri === REFUSED || statusUri === CANCELED || statusUri === SIGNED || !signingHubUrl) {
```
ValenberghsSven: Initial thought:
Should we make this similar to code used below?
`if ([REFUSED, SIGNED, CANCELED].includes(this.data?.value?.status?.uri) || !signingHubUrl) {`

After closer inspection:
Do we even need to check the status uri here?
In the loading of the data, we already check for these 3 statuses (+ MARKED) and we don't load the `signingHubUrl` if any of those 4 statussen are active.
So checking status again doesn't really make sense.
Also, in the case of `MARKED` the pill should be clickable because we have a route that leads to `signatures.index` in that case (even though it might not show up in that list yet because of constraints)

So a clickable pill should be when there is a `this.data?.value?.route` or a `this.data?.value?.signingHubUrl`

app/components/signature-pill.js
```diff
@@ -89,14 +92,15 @@ export default class SignaturePillComponent extends Component {
...
     if (statusUri === REFUSED || statusUri === CANCELED) {
       return 'error';
     } else if (signingHubUrl) {
-      return 'link';
+      return 'default';
+    } else if (statusUri === SIGNED) {
+      return 'success';
     }
-    return 'ongoing';
+    return 'link';
```
ValenberghsSven: when we have a link, we should have the skin link? (`"link"` is blue, `"default"` is grey)
So we need to show the blue link when we have a signingHubUrl or the route to signatures?

```
 const route = this.data.value?.route;
 
 .....
 
else if (signingHubUrl || route) {
   return 'link';
```

ValenberghsSven: Also, the css class `au-c-pill--link` shows a hover with a pointer. `au-c-pill-default` does not.

ValenberghsSven (changes requested):

tdn (approved):

ValenberghsSven (approved):
Made some small changes.
Template had an `or` with only 1 value.
changed the `if x or y return true else false` to `return !!x || !!y`

PR #1851, opened 2023-08-07
change decision editor edit button to one button per editor

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4249

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision.hbs
+++ b/app/components/agenda/agendaitem/agendaitem-decision.hbs
@@ -15,7 +15,7 @@
             this.loadReport.isIdle
             (user-may "manage-decisions")
             (not this.isEditing)
-            (or (not this.pieceParts this.report))
+            (not this.pieceParts this.report)
             this.enableDigitalAgenda
           )
         }}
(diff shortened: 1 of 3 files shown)


app/components/agenda/agendaitem/agendaitem-decision.hbs
```diff
@@ -15,7 +15,7 @@
             this.loadReport.isIdle
             (user-may "manage-decisions")
             (not this.isEditing)
-            (or this.pieceParts (not this.report))
+            (or (not this.pieceParts this.report))
```
sergiofenoll: `(or (not this.pieceParts this.report))` is a bit confusing.

`(not arg1 arg2 ...)` will in essence do `[arg1, arg2].every((param) => !!(param))` (see [`not` helper definition](https://github.com/jmurphyau/ember-truth-helpers/blob/master/packages/ember-truth-helpers/src/helpers/not.ts)). If it's the intention to do that, the wrapping `(or )` is not necessary.

If you want to do `!this.pieceParts || !this.report` you should write that as `(or (not this.pieceParts) (not this.report))` instead.

sergiofenoll (changes requested):
Looks okay apart from the detail about the `(or (not))` helper.

sergiofenoll (approved):

PR #1869, opened 2023-08-25
Feature/kas 111 beslissingsfiche klaarmaken voor digitaal ondertekenen

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-111

- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/416
- [ ] https://github.com/kanselarij-vlaanderen/digital-signing-service/pull/32
- [ ] https://github.com/kanselarij-vlaanderen/yggdrasil/pull/58


app/components/documents/document-card.js
```diff
@@ -172,9 +171,19 @@ export default class DocumentsDocumentCardComponent extends Component {
...
+        'filter[decision-activity][:id:]' : decisionActivity.get('id'),
+      });
+      this.signFlow = signFlow;
+      const signSubcase = yield this.signFlow.signSubcase;
+      this.signMarkingActivity = yield signSubcase.signMarkingActivity;
+    } else {
+      this.signMarkingActivity = yield this.args.piece.belongsTo('signMarkingActivity').reload();
+      const signSubcase = yield this.signMarkingActivity?.signSubcase;
+      this.signFlow = yield signSubcase?.signFlow;
+    }
```
sergiofenoll: I don't understand this if-else block. If the piece has a decision activity, why do we need to fetch the sign flow through it?

This will also give unexpected results when an agendaitem has multiple pieces (or a BF) that are marked to be signed (because this query will return a random sign flow, not necessarily the one related to the current piece.

Can't we just use the logic in the else branch for all pieces?

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -8,6 +8,7 @@ import CONSTANTS from 'frontend-kaleidos/config/constants';
 import addLeadingZeros from 'frontend-kaleidos/utils/add-leading-zeros';
 import VRDocumentName from 'frontend-kaleidos/utils/vr-document-name';
 import { deleteFile } from 'frontend-kaleidos/utils/document-delete-helpers';
+import { generateDecisionReport } from 'frontend-kaleidos/utils/generate-pdf';
```
sergiofenoll: Noticed the new https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/decision-report-generation.js, will replace this with that ASAP, and probably add one for minutes generation/adapt the existing service to also generate the minutes.

app/utils/generate-pdf.js
```diff
@@ -0,0 +1,56 @@
...
+    // response status.
+    if (error instanceof SyntaxError) {
+      throw new Error(
+        `Backend response contained an error (status: ${response.status})`
+      );
+    } else {
+      throw error;
+    }
+  }
+}
```
sergiofenoll, line 25: @ValenberghsSven I added these util functions before you added the `DecisionReportGeneration` service. I will start using the service in this PR as well, but I'm wondering if we shouldn't use the error handling here in the service? Right now the toaster (like the component code it got extracted from) only shows a toast with a generic error if something goes wrong, showing the error from the service might make debugging actual issues easier?

ValenberghsSven: Made the service to avoid duplication, any error handling is welcome

sergiofenoll: Added here https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1869/commits/9809bae3f2652520d1d1f07f7ba27973f39ee6ad

app/components/agenda/agendaitem/agendaitem-decision.js
kobemertens: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1887 I implemented the service for the minutes generation here

app/components/documents/document-preview/signatures-tab.hbs
```diff
@@ -21,7 +21,7 @@
         </AuButton>
       {{/if}}
     </div>
-  {{else if this.agendaitem}}
+  {{else}}
     {{#if this.loadCanManageSignFlow.isRunning}}
```
ValenberghsSven: this if check was removed, but further down in this component nothing changed?
Ex. a document of a meeting (agenda/documenten) now shows 
"De beslissing gekoppeld aan dit document is nog niet vrijgegeven" which doesn't make sense.

This If check needs to be less narrow but this might be too broad.

Decisions can't be marked in this view because of how we load the route.
` this.decisionActivity` is empty because we have not accounted for this being a treatment report instead of an agendaitem piece

app/components/signature-pill.js
```diff
@@ -57,7 +57,13 @@ export default class SignaturePillComponent extends Component {
...
     if (status) {
       if (status.uri === MARKED) {
-        route = "signatures.index";
+        const piece = await this.args.piece;
+        const isReportOrMinutes = await piece.isReportOrMinutes;
+        if (isReportOrMinutes) {
+          route = "signatures.decisions";
+        } else {
+          route = "signatures.index";
+        }
```
ValenberghsSven: isReportOrMinutes is a boolean property, should not need await.

Also, shorthanded:
`route = isReportOrMinutes ? "signatures.decisions" : "signatures.index"`

app/components/signatures/create-sign-flow.hbs
```diff
@@ -1,203 +1,13 @@
...
-  />
-{{/if}}
-
-{{#if this.showApproversModal}}
-  <Signatures::EmailModal
-    @title={{t "add-approval"}}
-    @onConfirm={{this.saveApprover}}
-    @onCancel={{toggle "showApproversModal" this}}
+{{#if @isReportOrMinute}}
+  <Signatures::CreateSignFlow::ReportOrMinute
```
ValenberghsSven: `::ReportOrMinutes` ?

app/config/permissions.js
```diff
@@ -72,6 +72,7 @@ const groups = [
       'manage-users',
       'manage-alerts',
       'manage-minutes',
+      'manage-secretary-signatures',
```
ValenberghsSven, line 76: Add a description in the list above

app/services/decision-report-generation.js
```diff
@@ -13,12 +13,36 @@ export default class DecisionReportGeneration extends Service {
...
-    if (!resp.ok) {
-      this.toaster.error(this.intl.t('error-while-exporting-pdf'));
-      return;
+    try {
+      const response = await fetch(`/generate-decision-report/${report.id}`);
+      try {
+        const data = await response.json();
+        if (!response.ok) {
+          throw new Error(
+            `Backend response contained an error (status: ${response.status}): ${JSON.stringify(data)}`
```
ValenberghsSven, line 22: guessing this could be the new `generate-pdf` util, but since we are moving action towards the backend I will check/change that in that PR

app/services/signature-service.js
```diff
@@ -94,43 +94,41 @@ export default class SignatureService extends Service {
...
-        decisionActivity,
-        status: status,
-      });
-      await signFlow.save();
-      const signSubcase = this.store.createRecord('sign-subcase', {
+    // TODO: Shouldn't the short & long title be coming from the agendaitem. Also when would show or edit this data?
+    const signFlow = this.store.createRecord('sign-flow', {
+      openingDate: now,
+      shortTitle: _case?.shortTitle,
+      longTitle: _case?.title,
```
ValenberghsSven, line 106: do we need a fallback for these 2 titles if there is no case? Also see TODO above, are these values ever used? Still wondering about that (also since current cases have no option for long titles)

app/templates/agenda/minutes.hbs
```diff
@@ -100,40 +100,36 @@
...
+                @closable={{false}}
+              />
+            {{else}}
+              <AuAlert
+                @title={{t "no-minutes-for-meeting"}}
+                @icon="circle-info"
+                @skin="info"
+                @size="small"
+                @closable={{false}}
+              />
```
ValenberghsSven, line 118: so all legacy + meetings up until now wont have digital minutes, but they will have a relevant manual pdf in `/documents` (this meeting or the next, need to check). Out of scope for now.

app/templates/agenda/minutes.hbs
```diff
@@ -100,40 +100,36 @@
...
+            {{/if}}
+          </Auk::Panel::Body>
+          {{#if (and this.currentPiecePart.value (user-may "manage-minutes"))}}
```
ValenberghsSven: `user-may` first? Second condition should not be checked if the first one is false

app/templates/signatures/ongoing-decisions.hbs
```diff
@@ -0,0 +1,94 @@
...
+              @excludedStatuses={{this.excludedStatuses}}
+            />
+          </div>
+        </div>
+      </div>
+    </div>
+  </m.sidebar>
+  <m.content @scroll={{true}}>
+    <KDataTable
+      data-test-route-search-publication-data-table
```
ValenberghsSven: this and others. Don't copy test selectors, they should be unique.
Only if you edit tests and make a unique test selector (following some own custom rules) + update the selector files in the cypress folders is this ok.

app/templates/agenda/minutes.hbs
```diff
@@ -100,40 +100,36 @@
...
+            <Auk::Panel::Body
+              class="au-o-flow {{if this.isEditing 'au-u-background-gray-100'}}"
+            >
               <AuContent>
                 <SanitizeHtml
                   @raw={{true}}
                   @value={{this.currentPiecePart.value.value}}
```
ValenberghsSven, line 128: out of scope. There are some minutes changes in this PR which is very confusing.
Also this `currentPiecePart.value.value` is not new, but it's very vague what we are doing.
Could we rename the value to something else like "htmlContent" because that is what we are doing.
tasks having a value and then that result having a value is hard to read.

app/components/documents/document-card.js
```diff
@@ -81,7 +81,6 @@ export default class DocumentsDocumentCardComponent extends Component {
     return !this.signMarkingActivity
       && this.signaturesEnabled
       && this.currentSession.may('manage-signatures')
-      && !!this.args.decisionActivity;
   }
```
ValenberghsSven: this may have some side effects. Now you can mark any document for signing it seems (ex. meeting documents)

app/components/signatures/create-sign-flow/piece.js
```diff
@@ -103,15 +95,19 @@ export default class SignaturesCreateSignFlowComponent extends Component {
...
 
+  @action
+  startEditSigners() {
+    this.showMinisterModal = true;
+  }
+
   @action
   saveApprover(approver) {
-    this.approvers.addObject(approver.toLowerCase());
+    this.approvers.addObject(approver);
```
ValenberghsSven: this was to show the users the correct sanitized emails upon entering them.
When they type in capitals, we immediately change it to lowerCase so we can show that in the view.
example without
image

app/components/documents/document-card.js
ValenberghsSven: every doc is now markable, checks needs to be refined instead of removed

app/templates/signatures/ongoing-decisions.hbs
```diff
@@ -0,0 +1,94 @@
+{{page-title (t "decisions-and-minutes-ongoing") }}
+
+<AuMainContainer class="l-search-container au-u-background-gray-100" as |m|>
+  <m.sidebar>
+    <div class="au-c-sidebar">
+      <div class="au-c-sidebar__content">
+        <div
+          id="search-subroute-filters-area"
```
ValenberghsSven: this is from the search route, does not belong here (also in ongoing route, must have been copied before)

app/controllers/signatures/ongoing-decisions.js
```diff
@@ -0,0 +1,103 @@
...
+  @action
+  onChangeStatus(statuses) {
+    this.statuses = statuses;
+  }
+
+  getDecisionActivityOrMeeting = async (signFlowOrPromise) => {
+    const signFlow = await signFlowOrPromise;
+    const decisionActivity = await signFlow.decisionActivity;
+    if (decisionActivity) {
+      return decisionActivity;
```
ValenberghsSven, line 85: does this work? As far as I can tell we don't always get/set the correct `decisionActivity` when it is a report (like the signatures-tab in document preview)

app/controllers/signatures/ongoing-decisions.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  getMeetingDate = async (signFlowOrPromise) => {
+    const record = await this.getDecisionActivityOrMeeting(signFlowOrPromise);
+
+    const modelName = record?.constructor?.modelName;
+    if (modelName === 'decision-activity') {
+      return record.startDate;
+    } else if (modelName === 'meeting') {
+      return record.plannedStart;
```
ValenberghsSven, line 100: this is a semi copy. it is true that we initially set the startDate of the decision-activity with the plannedStart of the meeting but that one can be changed later and this startDate is never updated. (low chance, can happen)
Why not get meeting.plannedStart in both cases for accuracy?

app/controllers/signatures/decisions.js
```diff
@@ -0,0 +1,250 @@
...
+import { trackedFunction } from 'ember-resources/util/function';
+import { TrackedArray } from 'tracked-built-ins';
+import { PAGINATION_SIZES } from 'frontend-kaleidos/config/config';
+
+const MANDATORY_SORT_OPTION = 'decision-activity';
+const DEFAULT_SORT_OPTIONS = [
+  '-decision-activity.start-date',
+  MANDATORY_SORT_OPTION,
+  'sign-subcase.sign-marking-activity.piece.name',
+];
```
ValenberghsSven, line 17: copied code, how do you sort on `decision-activity` and the `start-date` when it is not available for minutes.

app/templates/signatures/decisions.hbs
```diff
@@ -0,0 +1,210 @@
...
+    @clickable={{false}}
+    @loading={{this.createSignFlow.isRunning}}
+    @asideVisible={{and this.showSidebar (not this.createSignFlow.isRunning)}}
+    @content={{@model}}
+    @page={{this.pageSignaturesIndex}}
+    @size={{this.sizeSignaturesIndex}}
+    @numberOfItems={{@model.length}}
+    @totalNumberOfItems={{@model.meta.count}}
+    @onChangeSize={{fn (mut this.sizeSignaturesIndex)}}
+    @onChangePage={{fn (mut this.pageSignaturesIndex)}}
```
ValenberghsSven: sizeSignaturesDecisions & pageSignaturesDecisions  instead

app/components/signatures/select-secretary-modal.js
```diff
@@ -0,0 +1,46 @@
...
+import { action } from '@ember/object';
+import { task } from 'ember-concurrency';
+import { inject as service } from '@ember/service';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
+
+export default class SignaturesSelectSecretaryModalComponent extends Component {
+  @service mandatees;
+
+  @tracked selectedSecretary;
+  @tracked secretaryOptions = [];
```
ValenberghsSven: this model is not used anywhere. are you supposed to be able to select secretaries? Not implemented in the signatures.decisions route yet

app/components/signatures/create-sign-flow/report-or-minute.js
```diff
@@ -0,0 +1,62 @@
...
+    if (!this.args.decisionActivities) {
+      return;
+    }
+    const decisionActivities = this.args.decisionActivities.toArray();
+
+    const [head, ...tail] = decisionActivities;
+    const secretary = await head.secretary;
+
+    for (let decisionActivity of tail) {
+      const _secretary = await decisionActivity.secretary;
```
ValenberghsSven: `decisionActivityOrMeeting` here to avoid future confusion

ValenberghsSven (commented):
review / will be fixing some of the comments made.

app/components/signatures/select-secretary-modal.js
ValenberghsSven: implemented it in the mentioned route.

app/templates/agenda/minutes.hbs
ValenberghsSven: follow up ticket made KAS-4354

app/components/signatures/select-secretary-modal.js
ValenberghsSven: removed fully, edit of secretary not wanted

PR #1872, opened 2023-08-28
Modals not closable on loading

kobemertens:
- chore: bump appuniversum
- feat: make confirmation modals not closable on loading
- chore: make modals not closable on loading


sergiofenoll (changes requested):
The ticket specifically mentions the batch document edit, but that one isn't tackled in this PR. It doesn't use `AuModal`, instead still relies on the old `Auk::Modal` (where you can pass `@closeDisabled` to `<Auk::Modal::Header/>`

I assume there's still other older style modals, I recommend you list the usages of `Auk::Modal` and go through them as well.

ValenberghsSven (approved):
Some footers were not done yet so cancel could be pressed.
Pair-reviewed and tested half with @tdn 

I think we got most of it
Checking what jenkins does with this (some tests may try to press disabled buttons now)

PR #1878, opened 2023-09-01
cypress/KAS-4274-signing-profile-tests

MathiasZabot:


ValenberghsSven:
only failing test is also failing on DEV so merging

PR #1881, opened 2023-09-06
chore: add digital-signing error message to toast

kobemertens:


app/services/signature-service.js
```diff
@@ -49,14 +49,13 @@ export default class SignatureService extends Service {
...
     }
-
     // Prepare sign flow: create preparation activity and send to SH
     const response = await uploadPiecesToSigninghub(signFlows);
     if (!response.ok) {
       for (let signFlow of signFlows) {
         await this.removeSignFlow(signFlow, true);
       }
-      throw new Error('Failed to upload piece to Signing Hub');
+      throw new Error(response.statusText);
```
sergiofenoll: `response.statusText` will contain [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) messages (e.g. `OK` for status code 200, `Internal Server Error` for 500).

In general our services (should) return a JSON body that contains a JSON:API compliant body (see the [`error` helper in the Python template](https://github.com/mu-semtech/mu-python-template/blob/master/helpers.py#L62-L77) or the [`error` handler in the Javascript template](https://github.com/mu-semtech/mu-javascript-template/blob/master/helpers/mu/server.js#L32-L37)).

I think here you should try to get the JSON body and throw that (displaying the whole JSON body is fine, no need to extract the message/detail at this time). But be aware that it may not be JSON (either because the service is misconfigured or because the faulty response doesn't come from the service because e.g. it's down), in which case using `response.statusText` is probably okay (though you could check if it's possible to get the response body and just show that, whatever it may be).

kobemertens: Fixed in [4d77c98](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1881/commits/4d77c986e34c21cce6bad909c4991e12bbe2ad32)

ValenberghsSven (commented):
Had to make some changes after merge with development.
@brenner-company changed some functionality on toaster.

tested on DEV (pressing through disabled button)

Is this right? Could we extract the detail message if there is one?
Need to test this with multiple possible errors so we see what happens in those cases.
image

ValenberghsSven:
The toaster is now `AuAlert`. Seems to be quite small. 
This toaster might be better if it was wider but the only size options in appuniversum are tiny or small (we are using the default now)

ValenberghsSven:
`digital-signing` service is down message:
image

brenner-company:
> The toaster is now `AuAlert`. Seems to be quite small. This toaster might be better if it was wider but the only size options are tiny or small (we are using the default now)

The size of the toaster (and all its toasts within) is defined by CSS as seen here:
 
CleanShot 2023-09-25 at 15 49 14@2x

I could try and fiddle with this (maybe with `min-width` & `max-width` values) in another ticket if needed.

The `@size` options are regarding the general build-up & whitespace of the alert component inside the toaster, not the dimensions.

ValenberghsSven:
made another minor change.
`JSON.stringify()` can throw an error we did not catch which results in "unexpected end of JSON input"
Added try catch in that case.
now shows `Internal server error` when  `digital-signing` is down. Better, but not much of feedback for the users.

image

ValenberghsSven (approved):
I think this is ok. I would argue the size of the alert does not match the size of the message. But that is out of scope.
Should possibly test this with some real-world examples on ACC (f.e. mimic the external SH to not have the correct credentials...)

Extracting the detail message is not really needed as of now.

@sergiofenoll or @tdn second opinion needed?

PR #1884, opened 2023-09-13
add Timezone to jenkins testing

ValenberghsSven:
No ticket.
The mandatee dates are off in jenkins testing but are ok locally.
Checking if this is timezone related and if we can run the tests in our own timezone (Europe/Brussels) on the server.


ValenberghsSven:
closing this PR but keeping branch alive.

PR #1885, opened 2023-09-13
feature/KAS 4291 betreft document names

kobemertens:
- chore: autoformat
- chore: add regex to simplify document name


tdn (commented):
@ValenberghsSven reworked this in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1891 , since there were many test cases where the regex in this PR didn't work.

Instead, we apparently have a util for parsing the VR document name, and can use that to include the correct document names. 

So closing this PR and continuing with https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1891 instead

PR #1887, opened 2023-09-14
feature/KAS 4292 notulen automatisch

kobemertens:
[JIRA ticket](https://kanselarij.atlassian.net/browse/KAS-4292)

- chore: update decision-report-generation to work with minutes
- chore: use decision-report-generation service in minutes
- chore: update secretary in piecePart on secretary change

Related PRs:
- https://github.com/kanselarij-vlaanderen/minutes-report-generation-service/pull/2


app/services/decision-report-generation.js
```diff
@@ -7,13 +7,18 @@ export default class DecisionReportGeneration extends Service {
   @service store;
   @service intl;
 
-  generateReplacementReport = task(async (report) => {
-    const fileMeta = await this.exportPdf.perform(report);
+  generateReplacementReport = task(async (report, type = 'decision') => {
```
sergiofenoll: **suggestion (non-blocking)**: split this up into two functions to clarify the name and prevent passing "strings as enums"

This is just a nitpick and doesn't need to change, but I'd split this up into a function to generate the report (which is specific to beslissingsfiches) and the minutes (which is specific to the meeting). For BFs it's called a report because it's called a "verslag" in the applicatieprofiel.
Beside the naming, it's also nicer not to have to rely on a string argument to choose how the function should behave.
But again, a small nitpick. If you feel like it's okay as is, just leave it.

app/services/decision-report-generation.js
```diff
@@ -7,13 +7,18 @@ export default class DecisionReportGeneration extends Service {
...
   });
 
-  exportPdf = task(async (report) => {
-    const resp = await fetch(`/generate-decision-report/${report.id}`);
+  exportPdf = task(async (report, type = 'decision') => {
+    const typeToUrlBase = {
+      decision: 'generate-decision-report',
+      minutes: 'generate-minutes-report',
+    };
+    console.assert(Object.keys(typeToUrlBase).includes(type));
```
sergiofenoll: **chore**: remove this assert

I don't think we want to keep this in here. If you want proper error handling you should probably raise an exception, otherwise it just shows "Assertion failed" in the user's console which isn't very useful.

app/utils/html-utils.js
```diff
@@ -0,0 +1,21 @@
...
+  htmlString = htmlString.trim();
+  template.innerHTML = htmlString;
+  const element = template.content.querySelector(`#${id}`);
+  if (element) {
+    element.innerHTML = newValue;
+    return template.innerHTML;
+  } else {
+    console.debug(
+      `No element with id "${id}" found, could not set innerHTML`
+    );
```
sergiofenoll: **question**: who/what is this debug for?

I guess this could be useful for a developer, but given the (only?) place we use it this is not a debug I expect a developer to hit (because I doubt they will delete the element with the secretary ID from the minutes template). On the other hand I can imagine users encountering this if they (accidentally) delete the row where we insert the secretary, in which case I don't think a debug statement in the console would be very helpful.

I'd be inclined to either raise an actual exception here that the caller should handle and show a toast for, or just ignore the case (and discuss with the team what we should do, because not finding the html element means we can no longer use this feature, basically).

sergiofenoll (changes requested):
Some remarks, some very small changes needed.

app/components/meeting/edit-meeting-modal.js
```diff
@@ -256,6 +264,22 @@ export default class MeetingEditMeetingComponent extends Component {
...
+          'filter[minutes][:id:]': minutes.id,
+        });
+        const newValue = replaceSecretary(piecePart.value, this.secretary.person.get('fullName'));
+        piecePart.value = newValue;
+        yield piecePart.save();
+        yield this.decisionReportGeneration.generateReplacementReport.perform(
+          minutes,
+          'minutes'
+        );
+      }
```
sergiofenoll: **issue**: re-generating the minutes should occur after the meeting is saved

If we re-generate the minutes here, the secretary linked to the meeting is still the old one, which makes the bottom-right secretary name wrong. We should wait for the meeting to be saved before re-generating the report. This code (and check) should be moved lower in the function.

sergiofenoll (changes requested):

app/components/meeting/edit-meeting-modal.js
kobemertens: Hmm I guess you're right, however, this seems to be working so I'm not really sure how that is possible. Will investigate.

kobemertens: Ah, I just noticed there should also be a signature at the bottom of the page, this is not appearing on my end.

app/utils/html-utils.js
kobemertens: This was intended for a developer, I was foreseeing this function being used in other circumstances as well. I guess raising an error might be a better approach so it can be handled by the caller.

app/services/decision-report-generation.js
kobemertens: Good suggestion, changed in [55c1e17](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1887/commits/55c1e17f55181d14467c29a1ce1c2d8fa771aa8c)

app/services/decision-report-generation.js
kobemertens: You're right, I removed the mapping altogether in [55c1e17](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1887/commits/55c1e17f55181d14467c29a1ce1c2d8fa771aa8c)

app/utils/html-utils.js
kobemertens: Update: I'll just remove it, the caller can check if the substring is present beforehand if it needs to know

app/components/meeting/edit-meeting-modal.js
kobemertens: fixed in [1cd83a7](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1887/commits/1cd83a744dfb82d8ebf1b4246e3d8161e4a01c0a) (replace after save)

tdn (approved):
Pair review. 

Added title change for the secretary (waarnemend vs regular).

Merged development and resolved conflicts. 

Also made `saveReport` a task, so the button disabling works.

PR #1889, opened 2023-09-18
prevent signing of documents from retracted agendaitem

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4280


ValenberghsSven (approved):
Looks good.
Made minor adjustments because we should explicitly wait for the `decisionResultCode` from `decisionActivity` to resolve the getter `isRetracted`.
Also, in case `decisionActivity` is not present, use optional chaining.

PR #1891, opened 2023-09-18
Feature/kas 4291 betreft document names review

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4291

Other PR is already open but is using a new regex and an unsorted list.
instead of just committing I started this new branch.

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision.hbs
+++ b/app/components/agenda/agendaitem/agendaitem-decision.hbs
@@ -107,6 +107,7 @@
               @skin="link"
               @icon="renew"
               class="au-u-padding-left-none"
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
Pair reviewed this with Tom.
Added the try catch in case the given name cannot be parsed.
Added the same thing to VR notulen name (those documents have other names)
Fixed the VR notulen regex because it was too strict (session nr had to be 2 numbers (/02), but in reality short numbers are single digits (/2). Parsing was failing because of this. Not hotfix worthy
Added waiting on the document loader before updating the "betreft", I could update too soon which means the documents were not shown the first update.

PR #1893, opened 2023-09-20
save only on different accesslevel

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4297


app/services/piece-access-level-service.js
```diff
@@ -60,17 +60,19 @@ export default class PieceAccessLevelService extends Service {
...
+        case CONSTANTS.ACCESS_LEVELS.INTERN_SECRETARIE:
+          previousPiece.accessLevel = internSecretarie;
+          break;
+        case CONSTANTS.ACCESS_LEVELS.VERTROUWELIJK:
+          previousPiece.accessLevel = vertrouwelijk;
+          break;
+        default:
+          previousPiece.accessLevel = internRegering;
+      }
+      await previousPiece.save();
```
ValenberghsSven: when uploading a new piece, the accesslevel of the previous piece should be set to "intern regering"
Exceptions to that rule (for now):
- If the new piece is `internSecretarie`, previous piece is also that level.
- same for `vertrouwelijk`.

Your change prevents that defaulting to `internRegering`.
ex: if you upload a BIS to a `publiek` document, we use that accesslevel on our new document and then run this piece of code. so both pieces are then `publiek`. Switch is not entered.

ValenberghsSven (changes requested):

app/services/piece-access-level-service.js
ValenberghsSven: perhaps it would be better to not set the `previousPiece.accessLevel` in the switch.
something like this?
```
    let accessLevelToSet;
    switch (accessLevel.uri) {
      case CONSTANTS.ACCESS_LEVELS.INTERN_SECRETARIE:
        accessLevelToSet = internSecretarie;
        break;
      case CONSTANTS.ACCESS_LEVELS.VERTROUWELIJK:
        accessLevelToSet = vertrouwelijk;
        break;
      default:
        accessLevelToSet = internRegering;
        break;
    }
    if (previousAccessLevel.uri !== accessLevelToSet.uri) {
      previousPiece.accessLevel = accessLevelToSet;
      await previousPiece.save();
    }
```

sergiofenoll: We use the return boolean of the function to determine whether further processing is needed (e.g. if we just uploaded a QUATER and updated the access level of a TER, should we go update the access level of the BIS?)

Your current changes remove that behaviour, because you no longer return true when we update the access level.

Anyway this change doesn't actually resolve the issue detailed in the ticket, here we only prevent the saves when updating the previous piece versions. We're still going to unnecessarily save the access level of the actual piece.

Wouldn't it be a lot easier just to update `app/components/access-level-pill.js` so that `confirmChangeAccessLevel` is a no-op if the access level hasn't actually changed?

ValenberghsSven: we do need a `return true;` at the end of it. just so we keep looping over all versions.
@sergiofenoll is right that the ticket mentions the saving on the actual piece which is not covered.
Did not really look at the ticket during first review as I noticed the current code was not going to work.
But this change can stay because we do saves on previous versions even when it is not needed.


As for the original piece saving, we can try doing the no-op.
We will need to check if the accesslevel has changed (so keep initial .args state?) and then if the accesslevel has changed to a different value (same value also does not need to be saved)

sergiofenoll (approved):
In theory we don't handle the case where a user changes the access level, doesn't save, changes the access level back to the original and then saves. I don't think that matters though, this is fine.

ValenberghsSven (approved):

PR #1894, opened 2023-09-24
Feature/embroider

sergiofenoll:
Enable Embroider build in Kaleidos, which using its static analysis lowers our bundle/`assets/frontend-kaleidos.js` size by 4MB (from 20.8MB to 16.8MB)

Doesn't need to be merged, was more of a try to see how hard it'd be to convert Kaleidos to using Embroider (seemingly not very hard at all).


ValenberghsSven:
Updated it. Unsure if the rdfa editor personal commit was needed for some dependency fix or custom styling
We'll let jenkins and woodpecker do some work

ValenberghsSven:
Woodpecker is not picking up this PR.

sergiofenoll:
> Updated it. Unsure if the rdfa editor personal commit was needed for some dependency fix or custom styling We'll let jenkins and woodpecker do some work

The editor version we were using at the time didn't support Embroider builds, but any newer version should. The bump in this PR is probably unnecessary.

> Woodpecker is not picking up this PR.

Not sure why, will have a look.

ValenberghsSven:
The rdfa editor will be the prosemirror fix + the list styling. Might have to put this on hold until we can bump the rdfa editor first.

ValenberghsSven (changes requested):
@sergiofenoll Can you give this a second shot?
Its more a year behind now so new branch from DEV seems safer (conflicts in the build file)

It tried to throw it in a new branch and it just gave me
```
Build Error (WaitForTrees)

document is not defined
```
during npm run start
I won't push that branch

tdn:
We will take this up in https://kanselarij.atlassian.net/browse/KAS-4935

PR #1898, opened 2023-09-27
decision-report-generation now happens in backend and tab can be clos…

StuyckRuben:
…ed while this is happening

minutes-report-generation updated to be more in backend

https://kanselarij.atlassian.net/browse/KAS-4293

- [x] merge https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/5
- [x] merge https://github.com/kanselarij-vlaanderen/minutes-report-generation-service/pull/3


app/components/meeting/edit-meeting-modal.js
```diff
@@ -246,7 +247,7 @@ export default class MeetingEditMeetingComponent extends Component {
           });
           const pieceParts = yield report?.pieceParts;
           if (pieceParts?.length) {
-            yield this.decisionReportGeneration.generateReplacementReport.perform(report);
+            yield this.decisionReportGeneration.generateReplacementReports.perform([report]);
```
sergiofenoll: **issue**: `generateReplacementReports` should handle the whole batch of reports, not one by one

If we keep doing this one by one, we don't actually solve the problem of letting users close the tab while this work is ongoing. Instead of invoking the generate call one by one, gather all the reports that need to be regenerated in this for loop and then call the function on the list you've gathered.

app/services/decision-report-generation.js
```diff
@@ -1,31 +1,39 @@
...
+      {
+        timeOut: 3 * 60 * 1000,
+      }
+    ); 
+    const reportIDs = reports.map((report) => { return report.id });
+    const data = {
+      reportIDs: reportIDs
+    };
+    const url = `/generate-decision-report/generate`;
+    const resp = await fetch(url, {
```
sergiofenoll: **issue**: this call might get timed out by the browser

Since you're now passing in multiple reports, this call might take a while (e.g. for very large agendas, with 100s of reports to re-generate, this might take multiple minutes). Browsers time out calls that take longer than a minute and that's a setting that we can't change in Kaleidos (it's per-user and set locally). The simplest solution is to implement the job model, like we do for report generation, and let the backend service create a job that we can track in the frontend. This way the initial call should respond quite quickly (it should just create the job that we track and trigger it) and then we can poll the job until completion.

Some context/inspiration can be found in the code to generate reports:
- https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/controllers/publications/overview/reports.js#L37-L67
- https://github.com/kanselarij-vlaanderen/publication-report-service

As well as the code to generate file bundles:
- https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/agenda/agenda-header/agenda-actions.js#L264-L323
- https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/blob/master/app.js

(I wouldn't bother with creating a new service to create job, rather add an endpoint to create the job to the existing decision-report-generation-service).

app/components/meeting/edit-meeting-modal.js
sergiofenoll: **suggestion (out-of-scope, non-blocking)**: link ongoing jobs to meetings/decision activities

You don't need to implement this now, it's very out of scope for this ticket, but it might be interesting to link the ongoing jobs to the meeting/decision activities so that we can load/poll them when we open the agenda. That way if a user closes Kaleidos and re-opens it while a job to re-generate reports is still ongoing, we can show that to the user (and e.g. disable editing of decision activities).

translations/nl-be.json
```diff
@@ -1241,5 +1241,13 @@
...
-  "cannot-sign-because-agendaitem-retracted": "Dit agendapunt is ingetrokken. U kan de ondertekenflow niet opstarten."
+  "cannot-sign-because-agendaitem-retracted": "Dit agendapunt is ingetrokken. U kan de ondertekenflow niet opstarten.",
+  "decision-report-generation--toast-generating--message": "De beslissingen worden aangepast.",
+  "decision-report-generation--toast-generating--title": "Beslissingen Aanpassen",
+  "decision-report-generation--toast-generating-complete--message": "De beslissingen zijn aangepast.",
+  "decision-report-generation--toast-generating-complete--title": "Beslissingen Aangepast.",
+  "minutes-report-generation--toast-generating--message": "De notulen worden aangepast.",
+  "minutes-report-generation--toast-generating--title": "Notulen Aanpassen",
+  "minutes-report-generation--toast-generating-complete--message": "De notulen zijn aangepast.",
+  "minutes-report-generation--toast-generating-complete--title": "Notulen Aangepast."
```
sergiofenoll: **chore**: we normally don't do "Title Casing"

Replace this with "Beslissingen aanpassen", "Beslissingen aangepast", etc.

sergiofenoll (changes requested):

ValenberghsSven (approved):
very biased review.
Sending a full list of reports to backend not implemented.
Follow-up ticket exists.
For now:
- Made it so that we send batches of reports to generate to the service.
- We can track the progress and see what failed in frontend (each failed generation in a batch gets listed)
- bulk generation happens in service so navigation to different pages is possible after editing meeting (no need to wait with a modal open)

PR #1902, opened 2023-10-04
add message to beslissing piece-part when agendaitem is retracted or …

StuyckRuben:
…postponed
https://kanselarij.atlassian.net/browse/KAS-4331


app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -118,6 +120,37 @@ export default class AgendaitemControls extends Component {
     this.deleteItem(agendaitem);
   }
 
+  @task
+  *updateDecisionPiecePart(message) {
+    const report = yield this.store.query('report', {
```
sergiofenoll: **issue**: replace `this.store.query` with `this.store.queryOne`

[`Store.query`](https://api.emberjs.com/ember-data/4.8/classes/Store/methods/query?anchor=query) returns a promise that resolves to an array. Here `report` is actually an array that contains a single report, down the line you do `report.id` which won't work because the array has no id.

We have a custom [`queryOne`](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/store.js#L8-L18) method that automatically unpacks the array when you expect a single record.

sergiofenoll (changes requested):
**issue**: you also need to implement this behaviour for when you change the decision result in the pill in the decisions tab.

image


```js
// example code
// app/components/agenda/gendaitem/agendaitem-decision.js in onDecisionEdit

if (!this.report || !this.beslissingPiecePart) {
  return;
}

let newBeslissingValue = this.beslissingPiecePart.value;
const decisionResultCode = await this.args.decisionActivity
  .decisionResultCode;
switch (decisionResultCode.uri) {
  case CONSTANTS.DECISION_RESULT_CODE_URIS.UITGESTELD:
    newBeslissingValue = this.intl.t('postponed-item-decision');
    break;
  case CONSTANTS.DECISION_RESULT_CODE_URIS.INGETROKKEN:
    newBeslissingValue = this.intl.t('retracted-item-decision');
    break;
  default:
    break;
}
if (newBeslissingValue !== this.beslissingPiecePart.value) {
  const now = new Date();
  const newBeslissingPiecePart = await this.store.createRecord('piece-part', {
    title: 'Beslissing',
    value: newBeslissingValue,
    report: this.report,
    previousPiecePart: this.beslissingPiecePart,
    created: now,
  });
  await newBeslissingPiecePart.save();
  await this.decisionReportGeneration.generateReplacementReport.perform(
    this.report
  );
}
this.loadBeslissingPiecePart.perform();
```

ValenberghsSven:
I'm guessing that we also have to do this if the first creation happens when the decision is already "uitgesteld".
Now it only happens when there is already a generated report, which may not be the case.
Also when pressing "update" manually above the editor
Doing this now

ValenberghsSven (approved):

PR #1892, opened 2023-09-19
Generate minutes - feedback secretarie

StuyckRuben:
on updating secretary, piece-part for minutes mentioning secretary will update secretary
https://kanselarij.atlassian.net/browse/KAS-4303

- [ ] merge https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/4

PR #1904, opened 2023-10-09
Feature/kas 4329 optimalisaties betreft beslissingsfiche

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4329

- [x] merge https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1892

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision.js
+++ b/app/components/agenda/agendaitem/agendaitem-decision.js
@@ -79,7 +79,10 @@ export default class AgendaitemDecisionComponent extends Component {
   });
 
   loadDocuments = task(async () => {
-    let pieces = await this.throttledLoadingService.loadPieces.perform(this.args.agendaitem);
+    let pieces = await this.store.query('piece', {
+      'filter[agendaitems][:id:]': this.args.agendaitem.id,
+      'filter[:has-no:next-piece]': true,
+    });
     pieces = pieces.toArray();
     let sortedPieces;
     if (this.args.agendaitem.isApproval) {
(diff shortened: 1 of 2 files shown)


app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -79,7 +79,10 @@ export default class AgendaitemDecisionComponent extends Component {
   });
 
   loadDocuments = task(async () => {
-    let pieces = await this.throttledLoadingService.loadPieces.perform(this.args.agendaitem);
+    let pieces = await this.store.query('piece', {
+      'filter[agendaitems][:id:]': this.args.agendaitem.id,
+      'filter[:has-no:next-piece]': true,
+    });
```
sergiofenoll, line 85: **note (non-blocking)**: replacing `throttledLoadingService` should be fine, but was probably there to make use of caching

I think the `throttledLoadingService` was added here https://github.com/kanselarij-vlaanderen/frontend-kaleidos/commit/6e53387d212bea6460f0da1f4c2c483bf6438448 to make use of the caching, since we're already loading the pieces for the agendaitem when opening the agenda. Given that there's extra constraints here (we only want the latest versions) it's probably fine to replace it here.

@ValenberghsSven were there any other reasons to use the `throttledLoadingService` here?

sergiofenoll (approved):

app/components/agenda/agendaitem/agendaitem-decision.js
ValenberghsSven: was for caching, but the throttling does not account for BIS and TER on some agendaitems

ValenberghsSven (approved):
will merge and test in base branch

PR #1905, opened 2023-10-10
feature/KAS 4334 date title decisions

kobemertens:
- chore: regenerate pdf on meeting kind change
- chore: update decision report on plannedstart change
- chore: use correct filename based on agendaitem type

https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4334

Moved some code around to check if `meeting.kind` or `meeting.plannedStart` changed first and regenerate the decisionreport pdf files first if needed.

The decision report file name is also set correctly if the agenda item is an announcement.

:warning: Use https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/6 to see changes in the generated pdf files.


kobemertens:
@tdn Currently, the file name/pdf is not updated when the underlying subcase is updated from/to announcement/nota. Is this a requirement? If yes, should this also update the older versions of the file?

app/components/meeting/edit-meeting-modal.js
```diff
@@ -222,41 +222,61 @@ export default class MeetingEditMeetingComponent extends Component {
...
+        currentPlannedStart !== now 
+      ) {
+        this.args.meeting.kind = this.selectedKind;
         this.args.meeting.secretary = this.secretary;
-        const decisionActivities = yield this.store.queryAll('decision-activity', {
-          'filter[treatment][agendaitems][agenda][created-for][:id:]':
-            this.args.meeting.id,
-        });
+        this.args.meeting.plannedStart = this.startDate || now;
+        yield this.args.meeting.save();
```
sergiofenoll: **question (non-blocking)**: any reason we don't move this PDF re-generation logic after we save the meeting?

If we can do it without a double save it'd be great though, since it also impacts the tests.

sergiofenoll (approved):

app/components/meeting/edit-meeting-modal.js
kobemertens: Yeah makes sense, I was being lazy because moving the code up allowed me to reuse existing variables. Will change

kobemertens: fixed in [9553991](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1905/commits/95539913c4dd18d607bf4320e3980fa69a130273)

app/components/subcase/subcase-description-panel/subcase-description-edit.js
```diff
@@ -81,13 +83,49 @@ export default class SubcaseDescriptionEdit extends Component {
...
+  get enableDigitalAgenda() {
+    return (
+      ENV.APP.ENABLE_DIGITAL_AGENDA === 'true' ||
+      ENV.APP.ENABLE_DIGITAL_AGENDA === true
+    );
+  }
+
+  async updateDecisionActivities() {
+    if (this.enableDigitalAgenda) {
+      const decisionActivities = await this.args.subcase.decisionActivities;
```
ValenberghsSven: this behavior is probably not wanted on a postpone/reproposed subcase.
The postponed decision might have already been signed at that point while the reproposed decision is on designagenda.

In theory, there should never be a type change on a subcase that was postponed.
An announcement is usually a different subcase in the same case.
Since changing the type after postpone/reproposed would have an impact on 2 meetings.

So this code is technically fine but we could just query on reports instead since no changes are made to decision-activity?

cypress/e2e/unit/agenda.cy.js
```diff
@@ -346,6 +346,7 @@ context('Agenda tests', () => {
     cy.get(agenda.editMeeting.numberRep.view).should('contain', fullmeetingNumberVV);
     cy.intercept('PATCH', '/meetings/**').as('patchMeetings');
     cy.get(agenda.editMeeting.save).click();
+    cy.wait('@patchMeetings'); // We save the meeting twice when changing types, wait for two saves too
     cy.wait('@patchMeetings');
```
ValenberghsSven: we are no longer saving twice. removing

ValenberghsSven (approved):

PR #1906, opened 2023-10-10
feature/KAS 4288 cleanup failed signflow

kobemertens:
JIRA: https://kanselarij.atlassian.net/browse/KAS-4288

- chore: remove signflow in backend instead
- chore: add pushDeletion method to store
- chore: use pushDeletion to notify store of deletion
- chore: remove unneeded cleanup in frontend

:warning: Needs the following custom services
https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/422
https://github.com/kanselarij-vlaanderen/digital-signing-service/pull/35


app/services/signature-service.js
```diff
@@ -206,57 +203,14 @@ export default class SignatureService extends Service {
...
-      });
-      await signSigningActivities?.map(async (activity) => {
-        await activity.destroyRecord();
-      });
-      await signRefusalActivities?.map(async (activity) => {
-        await activity.destroyRecord();
-      });
-      // destroying signSubcase can throw ember errors. reload fixed that problem.
-      await signSubcase?.reload();
       if (!keepMarkingActivity) {
```
ValenberghsSven: Keeping marking activity was only in case of attempting to prepare (in service) and failing.
This reset is now embedded in the service when preparing.
No other usage of `keepMarkingActivity` in our frontend code.
No dispatcher call needed.
No endpoint needed in the digital signing service.

app/services/signature-service.js
```diff
@@ -206,57 +203,14 @@ export default class SignatureService extends Service {
...
+        await fetch(`/signing-flows/${signFlow.id}`, {
+          method: 'DELETE'
+        });
+      } else {
+        await fetch(`/signing-flows/reset-signflow/${signFlow.id}`, {
+          method: 'POST'
+        });
       }
       await piece.belongsTo('signedPiece').reload();
       await piece.belongsTo('signMarkingActivity').reload();
```
ValenberghsSven: The store still knows some of the models. 
we could unload these to be safe (when looking in the store for signflows we still find these)
```
      await signFlow.unloadRecord();
      await signSubcase.unloadRecord();
      await signMarkingActivity.unloadRecord();
```

ValenberghsSven (changes requested):
We can remove the `keepMarkingActivity`

app/services/signature-service.js
```diff
@@ -66,9 +66,6 @@ export default class SignatureService extends Service {
     // Prepare sign flow: create preparation activity and send to SH
     const response = await uploadPiecesToSigninghub(signFlows);
     if (!response.ok) {
-      for (let signFlow of signFlows) {
-        await this.removeSignFlow(signFlow, true);
-      }
```
ValenberghsSven: forgot to mention, when we set the status/creator in the frontend and reset it in the backend, we will need to refresh those relations here, when the response is not ok or the store will not be up to date
```
     for (let signFlow of signFlows) {
        await signFlow.belongsTo('status').reload();
         await signFlow.belongsTo('creator').reload();
      }
```

ValenberghsSven (approved):
should work fine now.

PR #1907, opened 2023-10-11
decision-activity status approved after making decision-activity

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4289


sergiofenoll:
**question**: no decision result code for announcements, is this intended?

If we go through with the current changes all new agendaitems will have no decision result code until a report is created. But for announcements reports generally aren't generated, which means that users would have to create a report, save it and then delete it have the result code set to "Akte genomen". Is this intentional, or should we instead keep the current behaviour for announcements and only add in the behaviour for new notas? @tdn

app/services/agenda-service.js
```diff
@@ -162,15 +160,6 @@ export default class AgendaService extends Service {
...
 
-    // load code-list item
-    const defaultDecisionResultCodeUri = isAnnouncement
-      ? CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME
-      : CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD;
-    const decisionResultCode = await this.store.findRecordByUri(
-      'concept',
-      defaultDecisionResultCodeUri
-    );
-
```
ValenberghsSven: announcement has to stay `KENNISNAME`
Not setting `GOEDGEKEURD` should be behind feature flag (maybe, users might want this now? feature flag for now)

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -405,6 +405,18 @@ export default class AgendaitemDecisionComponent extends Component {
     beslissingPiecePart,
     betreftPiecePart
   ) {
+    if (!report.id) {
```
tdn: Probably better to check for existence of decisionResultCode instead. Otherwise decisions might get overwritten when saving a report

tdn:
> **question**: no decision result code for announcements, is this intended?
> 
> If we go through with the current changes all new agendaitems will have no decision result code until a report is created. But for announcements reports generally aren't generated, which means that users would have to create a report, save it and then delete it have the result code set to "Akte genomen". Is this intentional, or should we instead keep the current behaviour for announcements and only add in the behaviour for new notas? @tdn

Indeed. Announcements were not part of the ticket and still must have "Akte genomen"

app/components/agenda/agendaitem/agendaitem-decision.js
ValenberghsSven: not having a report id means new file?.
only set `decisionActivity.decisionresultCode` when none exists might be better if a result code was already set manually.

ValenberghsSven (changes requested):

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -405,6 +405,19 @@ export default class AgendaitemDecisionComponent extends Component {
     beslissingPiecePart,
     betreftPiecePart
   ) {
+    const decisionResultCode = await this.args.decisionActivity.decisionResultCode;
+    if (decisionResultCode) {
```
ValenberghsSven: reverse if, only if there no decisionResultCode yet.

Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN ember build -prod
 
 FROM semtech/static-file-service:0.2.0
 
+ENV EMBER_ENABLE_DECISION_GOEDGEKEURD=""
```
ValenberghsSven: I did mean the existing `digital_agenda` flag, I could have been more precise.

app/services/agenda-service.js
```diff
@@ -163,13 +163,20 @@ export default class AgendaService extends Service {
...
-    const decisionResultCode = await this.store.findRecordByUri(
-      'concept',
-      defaultDecisionResultCodeUri
-    );
+    let defaultDecisionResultCodeUri;
+    if (isAnnouncement) {
+      defaultDecisionResultCodeUri = CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME;
+    } else if (ENV.APP.ENABLE_DECISION_GOEDGEKEURD) {
+      defaultDecisionResultCodeUri = CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD;
+    }
```
ValenberghsSven: suggestion
```
} else {
  defaultDecisionResultCodeUri = this.enableDigitalAgenda ? null : CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD;
}
```

tdn (approved):

ValenberghsSven (approved):
approved for now. Wait for cypress in case any test fails on missing decision result.

PR #1908, opened 2023-10-11
remove secretary selector on legacy agendas

StuyckRuben:
add meeting secretary to decision-activity if there is one
https://kanselarij.atlassian.net/browse/KAS-4316


sergiofenoll (approved):

app/components/meeting/edit-meeting-modal.js
```diff
@@ -140,11 +147,12 @@ export default class MeetingEditMeetingComponent extends Component {
...
 
   initializeSecretary = task(async () => {
-    if (this.enableDigitalAgenda) {
+    if (this.enableDigitalAgenda && !this.isPreKaleidos) {
       const secretary = await this.args.meeting.secretary;
       if (isPresent(secretary)) {
         this.secretary = secretary;
-      } else {
+      } else if (this.isNew) {
+        // if a meeting had no secretary yet we don't set a default one automatically
```
ValenberghsSven, line 155: This needs feedback from the users.
I think we should not set a default on editing meeting that have no secretary.
When you edit older meetings (before the release of the secretary feature than that meeting + all agendaitems will suddenly get a secretary.
So I made it that only on new meetings we set a default.
In other cases we use what is on the meeting.

tdn: That's perfect

ValenberghsSven (approved):

PR #1910, opened 2023-10-12
chore: don't make a count call in queryAll, get count from first call

sergiofenoll:
Store#count is really just requesting a single record from mu-cl-resources and using the returned metadata to know how many records there are in total. While this is fine if you only care about the count, since we're going to start fetching the actual records in Store#queryAll, we're better off doing an initial call for the first batch, checking the total count and then deciding if we need to fetch further batches.

While this isn't a giant amount of performance gain, for e.g. the government fields, the uncached count call takes ~200ms, which isn't nothing.

### DIFF
--- a/app/services/store.js
+++ b/app/services/store.js
@@ -19,11 +19,19 @@ export default class ExtendedStoreService extends Store {
 
   async queryAll(modelName, query, options) {
     query = query || {}; // eslint-disable-line no-param-reassign
-    const count = await this.count(modelName, query, options);
     const batchSize = query.page?.size || 100;
+
+    const firstBatch = this.query(modelName, Object.assign({}, query, {
+      'page[size]': batchSize,
+      'page[number]': 0,
+    }));
+
+    const batches = [firstBatch];
+    const result = await firstBatch;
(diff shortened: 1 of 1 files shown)


app/services/store.js
```diff
@@ -19,11 +19,19 @@ export default class ExtendedStoreService extends Store {
...
   async queryAll(modelName, query, options) {
     query = query || {}; // eslint-disable-line no-param-reassign
-    const count = await this.count(modelName, query, options);
     const batchSize = query.page?.size || 100;
+
+    const firstBatch = this.query(modelName, Object.assign({}, query, {
+      'page[size]': batchSize,
+      'page[number]': 0,
+    }));
+
```
kobemertens, line 28: **nitpick:** Never used `Object.assign` so I might be wrong but I think this can be rewritten as
```
this.query(modelName, {
    ...query
    'page[size]': batchSize,
    'page[number]': 0,
}
```

kobemertens (approved):
Looks good to me. Nice optimization

ValenberghsSven (approved):

PR #1911, opened 2023-10-12
chore: use edited on if the report has multiple revisions

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4336

Not sure if using `piece.modified` instead of `piece.created` everywhere might cause problems. Might have overlooked something.


app/components/documents/document-card.js
```diff
@@ -135,12 +135,12 @@ export default class DocumentsDocumentCardComponent extends Component {
         'filter[:id:]': id,
         include: 'document-container,document-container.type,access-level',
       });
-    const loadReportPiecePart = (id) =>
-      this.store.queryOne('piece-part', {
+    const loadReportPieceParts = (id) =>
+      this.store.query('piece-part', {
         'filter[report][:id:]': id,
       });
```
ValenberghsSven: `':has-no:next-piece-part': true,` gets the latest.
then we can check if piecePart.previousPiece exists > edited-on

app/components/documents/document-card.js
```diff
@@ -151,13 +151,22 @@ export default class DocumentsDocumentCardComponent extends Component {
...
         } else if (modelName === 'minutes') {
-          piecePart = yield loadMinutesPiecePart(this.piece.id);
+          pieceParts = yield loadMinutesPieceParts(this.piece.id);
+        }
+        
+        if (pieceParts) {
+          if (pieceParts.length === 1) {
+            this.altLabel = this.intl.t('created-on');
+          } else {
+            this.altLabel = this.intl.t('edited-on');
```
ValenberghsSven: if latestPiecePart.previousVersion

app/components/documents/document-card.js
```diff
@@ -151,13 +151,22 @@ export default class DocumentsDocumentCardComponent extends Component {
...
+          }
+        } else {
+          this.altLabel = null;
```
ValenberghsSven: setting this altLabel to null can go

app/components/documents/document-card.hbs
```diff
@@ -67,7 +67,7 @@
                   }}
                 >
                   {{this.label}}
-                  {{datetime-at this.piece.created}}
+                  {{datetime-at this.piece.modified}}
                   {{#if
```
ValenberghsSven: Not true for most cases.
ex. An agendaitem piece uploaded on monday and the accesslevel gets changed on friday does not mean a new file but will show the modified date.
We can make this value a getter, default it to `piece.created` and only change it to `piece.modified` when handling something with pieceparts.
 
 Could also use a check on pieceParts here, but the getter will be more usuable in other scenarios aswell.

ValenberghsSven (changes requested):

ValenberghsSven:
doing this myself quickly. Small changes

ValenberghsSven:
going ahead with planned changes, but would it be better to work with the `file.created` of the report or `piecePart.created`? @tdn 
The problem with `modified` still exists on reports. you can make the final version on thursday, change the accesslevel on friday.
It will show the "Bijgewerkt op" for friday, but the last change was made on thursday.

ValenberghsSven:
Used the `piecePart.created` for now. trying to think if that would ever be wrong.
Do we ever update the file of a report or minutes without changing the pieceParts?

ValenberghsSven:
also took the liberty to rename that vague "label" variable.

Are there any changes needed in the `document-version-history-item` component?
image

ValenberghsSven (approved):

PR #1912, opened 2023-10-13
chore: redirect to /mock-login when unauthenticated in dev mode

sergiofenoll:
When navigating to a page, if you're unauthenticated, our app will redirect you to a page where you can authenticate and after that's done you are sent back to the page you were first going to.

Locally this doesn't work because you're redirected to the login page, which won't work locally, so you "lose" the first page you were navigating to.

This change makes it so you're redirected to the mock-login route when unauthenticated, but only when the environment is set to development.


ValenberghsSven (approved):
looks fine

PR #1926, opened 2023-10-23
KAS-4306: changed custom footers to use regular footer template when possible

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4306


sergiofenoll (approved):
Working as expected, just added some replacements for test selectors. Waiting on tests now.

ValenberghsSven:
the last failing test I can fix on dev. no need to keep delaying this PR for it. Thx for fixing many of the tests.
Will be moving towards a cypress fixtures file with "Current Mandatees" and using historic mandatees were needed or this will keep happening

PR #1927, opened 2023-10-24
Feature/KAS-4353: Digital Agenda start time flags

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4353

Adds start dates for the digital agenda features that we use in the frontend to determine whether we will show the new digital agenda features on particular agenda.

For agendas before the flags' start date:
- the "Notulen" tab will be hidden on the agenda and the route will redirect to a 404 page
- the "Beslissing" tab on the agenda item will only show the old-style decision document card

The flags are set to the 1st of November and are separate (one for decisions and one for minutes), but both can be merged into one if they're being enabled at the same time (presumably the case).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,10 @@ module.exports = {
   },
   rules: {
     'prettier/prettier': 'off',
+    'no-unused-vars': [
+      'error',
+      { argsIgnorePattern: '^_' },
+    ],
   },
   overrides: [
     // node files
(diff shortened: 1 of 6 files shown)


tdn (commented):
Looks good, can be merged after conflicts have been resolved I think

ValenberghsSven:
tests are going to fail on this (because I changed them to create decision rather than upload.. ) but will fix later
The flags are not implemented in enough places perhaps?

ValenberghsSven (approved):

PR #1930, opened 2023-10-25
Missing p tag caused misalignment when automatically replacing secretary

tdn:


ValenberghsSven (approved):
I see, when updating it in the editor the `<p>` element is added automatically by the editor (not explicit in our code)
When we replace, since we are working with innerHtml replacements that paragraph goes away (since the id we are looking for is on the `<td>` element and we replace everything in it)
This is not the same for the title since that id is on a `<span>` and we replace the value in that element

We are kind of depending on the auto paragraphs now, will that break later in more ways?

ValenberghsSven:
On my review. the fix is not 100% the same as the editor would do it:
1st generation (in editor):

```
<td __rdfaid="fa95795c-2709-4d38-be0e-8e26c7992f30">
  <p data-indentation-level="0">De
    <span
      id="secretary-title"
      __rdfaid="7af3a1e9-3f19-4999-86a6-2af0258ca605"
      __tag="span"
    >secretaris</span>
  </p></td>
<td id="secretary" __rdfaid="00ff256f-2e5d-4897-83aa-bbaec828f284">
  <p data-indentation-level="0">Jeroen Overmeer</p>
</td>
```

after replace:
```
<td __rdfaid="fa95795c-2709-4d38-be0e-8e26c7992f30">
  <p data-indentation-level="0">De
    <span
      id="secretary-title"
      __rdfaid="7af3a1e9-3f19-4999-86a6-2af0258ca605"
      __tag="span"
    >waarnemend secretaris</span>
  </p>
</td>
<td id="secretary" __rdfaid="00ff256f-2e5d-4897-83aa-bbaec828f284">
  <p>Katrien Van Kets</p>
</td>
```
Notice that he 2nd paragraph does not have the "data-indentation-level=0". Does that matter?
@tdn

ValenberghsSven:
Suggestion, we also wrap this with a span (with the id) like we do for the title? Not sure if that makes any difference style wise

tdn:
I'll see what I can optimize about this today, but won't spend too much extra time on this if it works.

sergiofenoll:
I don't know if the lack of that data-attribute really matters.

If don't want to have it removed, what we could do instead is set the `id` on a span inside the `td` tag. Like this:
```diff
modified   app/controllers/agenda/minutes.js
@@ -35,7 +35,7 @@ function renderAttendees(attendees) {
         </tr>
         <tr>
           <td>De <span id="secretary-title">${secretaryTitle}</span></td>
-          <td id="secretary">${mandateeName(secretary)}</td>
+          <td><span id="secretary">${mandateeName(secretary)}</span></td>
         </tr>
       </tbody>
     </table>
```

The editor will auto-insert the wrapping `p` tag with the correct data-attribute. Note that the underlying issue remains, any HTML elements inside the element with the `secretary` id will be removed (e.g. if they bold or italicize the secretary name, that gets lost after replacing it).

ValenberghsSven:
changed the base to development, since we will be doing another release of tickets on dev

ValenberghsSven (approved):

PR #1941, opened 2023-11-03
KAS-4373 Added reports generation in batches

tdn:
Requires [updated decision-report-generation service](https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/13) and https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/430

### DIFF
--- a/app/services/decision-report-generation.js
+++ b/app/services/decision-report-generation.js
@@ -1,5 +1,6 @@
 import Service, { inject as service } from '@ember/service';
 import { task } from 'ember-concurrency';
+import CONSTANTS from 'frontend-kaleidos/config/constants';
 
 export default class DecisionReportGeneration extends Service {
   @service toaster;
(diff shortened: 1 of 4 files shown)


app/services/decision-report-generation.js
```diff
@@ -16,53 +17,55 @@ export default class DecisionReportGeneration extends Service {
...
-          this.intl.t('error-while-generating-report-name-pdf', {
-            name: failure.report.name,
-            error: failure.error.message,
-          })
+          this.intl.t('error-while-generating-report-pdfs')
         );
+      } else {
+        setTimeout(() => {
+          this.pollReplacementReports.perform(job, reports, generatingPDFsToast);
+        }, 2000);
```
sergiofenoll, line 66: **suggestion**: We have a `JobMonitor` service for stuff like this

It's not documented properly (you don't know how to use it without looking at where it's been previously used) but it's here: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/job-monitor.js

An example of usage: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/agenda/agenda-header/agenda-actions.js#L274-L322

Basically, you get a job record, register it to be monitored and pass a callback that will be called once the job is finished (which in practice is just used to replace the toaster).

Actually using it would require some substantial changes in the backend service so I think we can leave this as-is for now, but just something to keep in mind for the future.

sergiofenoll (approved):

ValenberghsSven (approved):

PR #1943, opened 2023-11-06
KAS-4366: Regenerate reports when re-ordering agendaitems

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4366

Regenerates report (so new PDF) and changes name of piece.

:warning: We also need to change the name of the piece when changing a meeting! In my opinion this should go in the `DecisionReportGenerationService`, but I think those changes should occur after the "job generate reports" PR gets merged, to prevent conflicts.

### DIFF
--- a/app/controllers/agenda/agendaitems.js
+++ b/app/controllers/agenda/agendaitems.js
@@ -15,7 +15,6 @@ import {
   all,
   animationFrame
 } from 'ember-concurrency';
-import generateReportName from 'frontend-kaleidos/utils/generate-report-name';
 
 export default class AgendaAgendaitemsController extends Controller {
   queryParams = [
@@ -78,7 +77,6 @@ export default class AgendaAgendaitemsController extends Controller {
 
   @task
   *assignNewPriorities(reorderedAgendaitems, draggedAgendaItem) {
-    const draggedAgendaItemNumber = draggedAgendaItem.number;
     const draggedAgendaItemType = yield draggedAgendaItem.type;
     // reorderedAgendaitems includes all items on the whole page. We only want to re-order within one category (nota/announcement/...)
     const reorderedAgendaitemsOfCategory =  [];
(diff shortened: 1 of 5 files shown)


app/controllers/agenda/agendaitems.js
```diff
@@ -85,6 +89,26 @@ export default class AgendaAgendaitemsController extends Controller {
       }
     }
     yield setAgendaitemsNumber(reorderedAgendaitemsOfCategory, true, true); // permissions guarded in template (and backend)
+    const newDraggedAgendaItemNumber = draggedAgendaItem.number;
+
+    const needNewReports = reorderedAgendaitemsOfCategory.slice(
+      Math.min(draggedAgendaItemNumber, newDraggedAgendaItemNumber) - 1, // number is 1-indexed
+      Math.max(draggedAgendaItemNumber, newDraggedAgendaItemNumber), // slice does not include end index
+    );
```
ValenberghsSven: Would this code would fit better either inside the `setAgendaitemsNumber` method?
Putting this block in that method would also solve the "on deletion of agendaitem" premise of creating new reports since that method is used there.

Or using the return of that method (should return only agendaitems that have been saved?) Perhaps the return doesn't work as expected, we are not using as right now.

ValenberghsSven (commented):
Just wondering if this is the right spot for this code.

app/controllers/agenda/agendaitems.js
sergiofenoll: Moved the logic in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/1943/commits/6cdf59e7703e6c4d47b624fce0559fecf682decf

app/utils/agendaitem-utils.js
```diff
@@ -84,22 +85,42 @@ export const sortByNumber = (groupedAgendaitems, allowEmptyGroups) => {
...
+ * @param {DecisionReportGeneration} decisionReportGeneration The decisionReportGeneration service
+ * @param {Boolean} isEditor When true, the user is allowed to edit the trigger a recalculation of the number.
+ * @param {Boolean} isDesignAgenda When true, the agenda is a designagenda.
  */
-export const setAgendaitemsNumber = async(agendaitems, isEditor, isDesignAgenda) => {
+export const setAgendaitemsNumber = async(agendaitems, meeting, store, decisionReportGeneration, isEditor, isDesignAgenda) => {
   if (isEditor && isDesignAgenda) {
-    return await Promise.all(agendaitems.map(async(agendaitem, index) => {
+    const reports = [];
+    const promises = await Promise.all(agendaitems.map(async(agendaitem, index) => {
```
tdn, line 98: We'll need to test this on large agendas on ACCEPT. Since the operation is a bit more expensive, I wonder how this will handle an agenda of 100s of items

tdn (changes requested):
As discussed yesterday. The ticket included triggering the re-generation of the reports after the meeting number has changed (in the edit-meeting-modal). 

Could be added here on in a separate PR, but at the least should be done before we release to ACCEPT.

Other than that this PR looks good.

Tnx!

sergiofenoll:
> As discussed yesterday. The ticket included triggering the re-generation of the reports after the meeting number has changed (in the edit-meeting-modal).
> 
> Could be added here on in a separate PR, but at the least should be done before we release to ACCEPT.
> 
> Other than that this PR looks good.
> 
> Tnx!

Added functionality to regenerate report names when editing meeting number. A double check of the code would be appreciated :smile:

ValenberghsSven (approved):
Tested by removing agendaitems, reordering agendaitems, changing meeting info, changing types of agendaitem. With and without BIS versions.

Found a new bug in decision-report-generation micro-service and fixed it but this branch looks very sturdy and reports are generated as expected

Out of scope : type changes of agendaitems does not reorder them (double numbers possible in both notas and announcements, but was already the case
That is also a rare use-case and manual workaround (moving the items) works so ok

ValenberghsSven:
Will need to check cypress first, some weird fails are happening

ValenberghsSven:
Ah, perhaps it was the pre dev merge. becuase we removed the `isReportOrMinutes` flag. from backend

ValenberghsSven:
risking the merge :sweat_smile:

PR #1945, opened 2023-11-07
feature/KAS 4380 bulk beslissingen markeren

kobemertens:
:warning: Requires the following PR for signing service: https://github.com/kanselarij-vlaanderen/digital-signing-service/pull/38
- add translations
- add menu button
- add markDecisionsForSigning method


app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -333,6 +333,42 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
 
+  @action
+  async markDecisionsForSigning() {
+    const agenda = this.args.currentAgenda;
+    const agendaitems = await agenda.agendaitems;
+    const reports = await this.store.queryAll('report', {
+      'filter[decision-activity][treatment][agendaitems][:id:]': agendaitems
+        .map((x) => x.id)
+        .join(','),
+    });
```
ValenberghsSven: We do this differently in other places (edit-meeting-modal). Just query all the reports (with piece-parts and no previous version)

```
await this.store.queryAll('report', {
      'filter[:has-no:next-piece]': true,
      'filter[:has:piece-parts]': true,
      'filter[decision-activity][treatment][agendaitems][agenda][created-for][:id:]':
        this.args.meeting.id,
    });
```

This ensures we only get generated reports (not uploaded legacy docs) and in case of BIS we get the BIS report

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -333,6 +333,42 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
     this.args.didApproveAgendaitems();
   }
 
+  @action
+  async markDecisionsForSigning() {
+    const agenda = this.args.currentAgenda;
+    const agendaitems = await agenda.agendaitems;
```
ValenberghsSven: see other comment, we can use meeting for query instead of agenda and the agendaitems.
Regardless: `agenda.agendaitems` is something we should avoid for cache / stale data / concurrency reasons.
We normally do a query to `agendaitems` with a filter on agenda id and a page size (or queryAll would work)

This info is not readily available so I might add it to the agenda.js model with a warning on it. You could not have known this. We only use it in older code and should really be refactored to more specific queries tbh.
Also because normally there is a page size limit on relations. you would only get 20 results from a list this way unless we write an adapter for it (used to be the case, need to verify if that is still a thing)

ValenberghsSven (changes requested):

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -333,6 +333,42 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+    });
+    this.toaster.close(loadingToast);
+    if (!resp.ok) {
+      this.toaster.warning(this.intl.t('error-while-marking-decision-reports-for-signing'));
+      return;
+    } else {
+      this.toaster.success(
+        this.intl.t('decision-reports-are-marked-for-signing')
+      );
+    }
```
sergiofenoll: **suggestion**: We should refresh the route after the fetch all to show the sign flows.

At least if the current route is the "beslissing" of an agendaitem, otherwise it looks like nothing happened.

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -333,6 +333,42 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+  @action
+  async markDecisionsForSigning() {
+    const agenda = this.args.currentAgenda;
+    const agendaitems = await agenda.agendaitems;
+    const reports = await this.store.queryAll('report', {
+      'filter[decision-activity][treatment][agendaitems][:id:]': agendaitems
+        .map((x) => x.id)
+        .join(','),
+    });
+    const loadingToast = this.toaster.loading(
```
sergiofenoll: **suggestion (non-blocking)**: the toasts could be a little more descriptive for edge cases

If there are no reports we might want to show a toast that nothing has been marked instead. We might also want to show a count of reports being marked (not all agendaitems might have a report). Nothing urgent so no need to change for this PR, if users mention it we can make a ticket for it down the line.

ValenberghsSven:
Moved a part of the code. don't like fetch in a component. Should be a util or service.
Added a message when 0 reports are found.
Added flag to only show the option when feature would be enabled (feature flag not needed in that case)
This didn't work. I kept getting 404 from the signing-service.
There was no dispatcher change anywhere, so I made one
Don't think this could have worked without it https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/432/files
(still doesn't, will look again tomorrow)

ValenberghsSven (approved):
went over most of these changes with Sergio.
Added some translations fixes to be singular/plural

PR #1952, opened 2023-11-14
fix: pass signFlow to removeSignFlow in document details panel

sergiofenoll:


tdn (approved):

app/components/documents/document-details-panel.js
```diff
@@ -134,7 +134,12 @@ export default class DocumentsDocumentDetailsPanel extends Component {
   }
 
   verifyDeleteSignFlow = task(async () => {
-    await this.signatureService.removeSignFlow(this.args.piece);
+    const signMarkingActivity = await this.args.piece.belongsTo('signMarkingActivity').reload();
+    if (signMarkingActivity) {
```
tdn: @sergiofenoll apparently you can use `removeSignFlowForPiece` in the service to do exactly this check...

tdn (changes requested):

app/components/documents/document-details-panel.js
tdn: oops, apparently that is only for MARKED signflows.

tdn (approved):

PR #1959, opened 2023-11-21
KAS-4412 Add loader around decision edit button

StuyckRuben:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?assignee=61fa59d71c9bb000727c7cd4&selectedIssue=KAS-4412

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision.hbs
+++ b/app/components/agenda/agendaitem/agendaitem-decision.hbs
@@ -257,13 +257,15 @@
                </h4>
              </Toolbar.Group>
              <Toolbar.Group @position="right" as |Group|>
-               {{#unless 
-                    (and
-                      (not this.isEditingAnnotation)
-                      this.mayEditDecisionReport
-                    )}}
-                  <Auk::Loader />
-                {{else}}
+               {{#if 
+                  (and
+                     this.loadReport.isIdle
+                     (user-may "manage-decisions")
+                     (not this.isEditingAnnotation)
+                     (or this.pieceParts (not this.report))
+                     this.mayEditDecisionReport
+                   )
+                 }}
                    <Group.Item>
                      <AuButton
                        @skin="naked"
@@ -275,7 +277,7 @@
                        {{t "edit"}}
                      </AuButton>
                    </Group.Item>
-                 {{/unless}}
+                 {{/if}}
              </Toolbar.Group>
            </Auk::Toolbar>
          <SanitizeHtml
@@ -339,13 +341,15 @@
                   </h4>
                 </Toolbar.Group>
                 <Toolbar.Group @position="right" as |Group|>
-                  {{#unless 
-                      (and 
-                      (not this.isEditingConcern)
-                      this.mayEditDecisionReport
-                      )}}
-                      <Auk::Loader />
-                  {{else}}
+                  {{#if 
+                      (and
+                        this.loadReport.isIdle
+                        (user-may "manage-decisions")
+                        (not this.isEditingConcern)
+                        (or this.pieceParts (not this.report))
+                        this.mayEditDecisionReport
+                      )
+                  }}
                       <Group.Item>
                         <AuButton
                           data-test-agendaitem-decision-create
@@ -358,7 +362,7 @@
                           {{t "edit"}}
                         </AuButton>
                       </Group.Item>
-                    {{/unless}}
+                    {{/if}}
                 </Toolbar.Group>
               </Auk::Toolbar>
             <SanitizeHtml
@@ -422,13 +426,12 @@
                     </h4>
                   </Toolbar.Group>
                   <Toolbar.Group @position="right" as |Group|>
-                     {{#unless 
-                          (and
-                            (not this.isEditingTreatment)
-                            this.mayEditDecisionReport
-                            )}}
-                        <Auk::Loader />
-                      {{else}}
+                       {{#if
+                      (and
+                        (not this.isEditingTreatment)
+                        this.mayEditDecisionReport
+                      )
+                    }}
                         <Group.Item>
                           <AuButton
                             data-test-agendaitem-decision-create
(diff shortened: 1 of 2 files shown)


ValenberghsSven (changes requested):
Just from the ticket info 'edit button shows for a brief second when not allowed' I'm going to guess that this is not fixing that issue.
If the button is shown it means the "if" part is true for a second
This just adds a loading button in the "else" part. the "if" part will still be true for a second followed by a loader

Also, there are 3 edit buttons (one for each piece part)

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -182,6 +184,15 @@ export default class AgendaitemDecisionComponent extends Component {
...
   });
 
+  loadMayBeEdited = task(async () => {
+    if (this.currentSession.may('manage-decisions') && 
+    !this.isEditingAnnotation && 
+    (this.pieceParts || !this.report) 
+    && this.mayEditDecisionReport) {
+      this.mayBeEdited = true;
+    }
+  });
```
kobemertens: I like the idea of having the "may be edited" conditions in a separate function, however, this does not need to be a `async` task because there is nothing to be `await`ed. This makes the code unnecessarily complicated. I think it would be better to see if this code can be reconciled with the already existing `mayEditDecisionReport` getter and use that one instead. Also, be careful, the condition is not 3 times the same, there is a difference in `isEditingAnnotation` (vs `isEditingTreatment`).

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -56,11 +56,13 @@ export default class AgendaitemDecisionComponent extends Component {
   @tracked decisionViewerElement = null;
 
   @tracked decisionDocType;
+  @tracked mayBeEdited = false;
 
   constructor() {
     super(...arguments);
     this.loadReport.perform();
     this.loadCodelists.perform();
+    this.loadMayBeEdited.perform();
```
kobemertens: This would not be needed when using a getter.

kobemertens (changes requested):
Does this fix the issue? I have not been able to reproduce to test if this is a fix, will update.

Update: the problem most likely lies in this line https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/5c3ffec0032cc751ffe26e6bbdcec8eaf69d276e/app/components/agenda/agendaitem/agendaitem-decision.js#L807 `this.hasSignFlow` is initialized with `false`, which causes the condition to evaluate to `true` until `loadSignatureRelatedData` is finished. This should be initialized to `null` instead, then you can check `this.hasSignFlow === false` instead of `!this.hasSignFlow`.

ValenberghsSven:
>Does this fix the issue? I have not been able to reproduce to test if this is a fix, will update.

I was able to reproduce. It is based on the signing stuff.
You can reproduce by having a signflow with a different status than marked (could do by returning true for `hasSignFlow` and false for `hasMarkedSignFlow`

app/components/agenda/agendaitem/agendaitem-decision.hbs
```diff
@@ -257,15 +257,13 @@
...
-               {{#if
-                   (and
-                     this.loadReport.isIdle
-                     (user-may "manage-decisions")
-                     (not this.isEditingAnnotation)
-                     (or this.pieceParts (not this.report))
-                     this.mayEditDecisionReport
-                   )
-                 }}
+               {{#unless
```
kobemertens: Nice cleanup but this will not work correctly, this condition evaluates to `if (this.isEditingAnnotation || !this.mayEditDecisionReport) { // show loader }`. This means that a loader will always be shown if the user may not edit the decision report. This is not the behavior we want. I would not show any loaders there since they might just resolve to nothing if the user is not allowed to edit. Whereas you would expect a loader to be a placeholder for a UI element.

kobemertens (changes requested):
You can `--proxy` to the development backend and test for [this agendapunt](https://kaleidos-dev.vlaanderen.be/vergadering/6554BAF2BC3B84D2D735935A/agenda/6554BAF4BC3B84D2D735935E/agendapunten/6554BB39BC3B84D2D735936A/beslissingen?anchor=6554BB39BC3B84D2D735936A). Now it shows a loader indefinitely which is not what we want.

app/components/agenda/agendaitem/agendaitem-decision.hbs
ValenberghsSven: Using an `unless` usually means there is no `else` statement, better to use if else instead even though it works
The ticket does not really ask for a loader to be present so doesn't really have to be there afaik.
But if you want to await loading you need to check for:
`this.loadReport.isRunning || this.loadSignatureRelatedData.isRunning`

`loadReport` contains some awaits for tasks (loading the 3 piece parts)
This task also initiates the `loadSignatureRelatedData` task, but does not wait for it.
If we were to wait for that to resolve in `loadReport`, it would have been enough to check `this.loadReport.isIdle` in the template + the values of the `isEditing...` and `hasSignFlow` (should always be a boolean) and `hasMarkedSignFlow` (both of these will be resolved if `loadSignatureRelatedData` has completed)
So we could have left the template alone and just changed the component side.

However the `and` statement in the template is getting a bit much so extracting that to the component will be more readable.
Some of the checks are the same, only the `isEditing..specificPiecePart..` is different.
So suggesting:
- maybe an optional getter to check the loading states (either await it all in loadReport or await both tasks)
Check that getter + the extra requirements to determine if the buttons should be shown.

or

- just make 1 getter that checks if the tasks have been run and the shared requirements are met to edit.
- use that getter in the template in combination with the `isEditing..specificPiecePart..`.


This should result in what you want:
If true: show edit button, if false, show nothing (loader optional)

sergiofenoll (approved):

PR #1960, opened 2023-11-21
Feature/KAS-4409: Fix conflicting signers alert not going away after deselecting conflicting sign flows

sergiofenoll:
Ensure that hasConflictingSigners is reset when changing selection

To make it work (and be more consistent with the programming style of ember-resources), we make the loadSigners resource return the value of hasConflictingSigners instead of mutating it. Mutating it can cause errors: e.g. the resource gets fired as soon as rendering occurs, but we're also reading hasConflictingSigners at that point, if we write to a tracked variable that has already been read in a render cycle, Ember will crash.

Instead the proper way (according to ember-resources) is for a resource not to mutate anything outside of itself and for the resource to return whatever state is needed and that can be depended on.

### DIFF
--- a/app/components/signatures/create-sign-flow/piece.js
+++ b/app/components/signatures/create-sign-flow/piece.js
@@ -19,11 +19,14 @@ export default class SignaturesCreateSignFlowPieceComponent extends Component {
   @tracked approvers = new TrackedArray([]);
   @tracked notificationAddresses = new TrackedArray([]);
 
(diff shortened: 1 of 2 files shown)


StuyckRuben (approved):
Tested code and it works. One failing unrelated to this PR.

ValenberghsSven (approved):
Changed 1 small thing. decisions could not be started anymore with multi select.
Works fine now.

PR #1961, opened 2023-11-23
Feature/KAS-4348: Resend case to VP

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4348

>[!WARNING]
> This PR isn't ready to be merged yet, there's still a bunch of work to be done

Adds the new submitted piece model and makes some small changes to the SendToVpModal:

- Doctype check Decreet, not Ontwerpdecreet
- Take derived/source files into account when checking what's missing
- Show if both signed and Word file is missing (using suffix `XXX mist (ondertekend en in Word)`

Things that should still be changed here to fulfil the ticket requirements:
- [x] An "opmerkingenveld" needs to be added to the send to VP modal and it should be sent to the backend
- [x] ~~Make sure the button and modal are only shown to admin/secretarie/kabinetsdossierbeheerder of the submitting mandatee~~ (this is now part of https://kanselarij.atlassian.net/browse/KAS-4433)
- [x] The "Send to VP modal" should know what files we're actually sending (on a re-send we don't actually send all the files, the current frontend implementation doesn't know that)

Nice to have things for this feature (but that don't need to be implemented for this ticket specifically):
- [ ] The "Send to VP modal" should probably get the files its sending from the backend, to make sure it accurately reflects what's going to be sent
- [x] When the "Send to VP modal" displays the "Deze bestanden ontbreken nog" message, it should be aware of files it has already previously sent (i.e. if we're just sending a signed PDF but everything else has already been sent, it shouldn't complain about missing all other files)
- [ ] Other stuff?

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/438
- [ ] https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/4


app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -36,12 +36,23 @@ export default class AgendaitemControls extends Component {
     super(...arguments);
 
     this.loadAgendaData.perform();
-    this.loadDecisionActivity.perform();
-    this.loadCanSendToVP.perform();
+    this.loadDecisionActivity.perform()
+        .then(() => {
+          this.loadCanSendToVP.perform();
```
tdn: Do we want to wrap this in the feature flag?

tdn: Nevermind, the function itself checks it.

app/components/agenda/agendaitem/send-to-vp-modal.js
```diff
@@ -38,20 +51,104 @@ export default class SendToVpModalComponent extends Component {
...
+          NOTA.id,
+          ADVIES.id,
+        ].join(','),
+        'filter[access-level][:id:]': [
+          INTERN_OVERHEID.id,
+          PUBLIEK.id,
+        ].join(','),
+        'filter[submission-activity][subcase][:id:]': subcase.id,
+        sort: 'name',
+      })).slice();
```
tdn, line 129: Why is this `slice()` here?

app/components/agenda/agendaitem/send-to-vp-modal.js
```diff
@@ -38,20 +51,104 @@ export default class SendToVpModalComponent extends Component {
...
+
+          if (submittedPiece.unsignedFile === pdf.uri) {
+            hasSentPdf = true;
+          }
+          if (submittedPiece.wordFile === word?.uri) {
+            hasSentWord = true;
+          }
+          if (submittedPiece.signedFile === signed?.uri) {
+            hasSentSigned = true;
+          }
```
tdn, line 176: I'm always reluctant to write something like `if (somestatement) { someBoolean = true } else { someBoolean = false }` unless it's inside a loop where you only set the value of `someBoolean` once and then break the loop.

These statements can be rewritten as `someBoolean = someStatement`, so in this case, I'd prefer

```
let hasSentPdf = (submittedPiece.unsignedFile === pdf.uri);
let hasSentWord = (submittedPiece.wordFile === word?.uri);
let hasSentSigned = (submittedPiece.signedFile === signed?.uri);
```

which replaces 12 lines of code by 3.

app/components/agenda/agendaitem/send-to-vp-modal.js
```diff
@@ -96,25 +215,31 @@ export default class SendToVpModalComponent extends Component {
...
-      CONSTANTS.CONCEPT_SCHEMES.DOCUMENT_TYPES
-    );
+  @action
+  onChangeComment(event) {
+    this.comment = event.target.value;
+  }
 
-    const BESLISSINGSFICHE = this.documentTypes.find(
-      (type) => type.uri === CONSTANTS.DOCUMENT_TYPES.BESLISSINGSFICHE
+  async loadDocumentTypes() {
```
tdn: This loads the same (or at least an overlapping set of) document types as the `loadData` function. I'd rather see this in a separate task, with the constants then being accessible for the whole component all at once.

app/models/parliament-submission-activity.js
```diff
@@ -4,7 +4,7 @@ export default class ParliamentSubmissionActivity extends Model {
   @attr('date') startDate;
   @attr('date') endDate;
 
-  @belongsTo('user', { async: true }) submitter;
+  @belongsTo('user', { inverse: null, async: true }) submitter;
 
-  @hasMany('piece', { async: true }) pieces;
+  @hasMany('submitted-piece', { inverse: null, async: true }) submittedPieces;
```
tdn, line 15: Just a question, since I know this was a point of contention while creating the story. We did land 100% on not persisting the `comment` that goes with the `submission-activity`?

tdn (changes requested):
Some minor code style/clarification comments, but otherwise looks good at first glance.

Pending these changes and a few tests tomorrow, I think this is close to being mergeable.

app/components/agenda/agendaitem/send-to-vp-modal.js
sergiofenoll: The result of `query{All}` and other Ember Data store related stuff (`await record.thisIsAHasManyRelationship`) is a special kind of Ember array object. `ArrayProxy` or `ManyArray` or `RecordArray`. They may not always support all array-like methods/be iterable. 

When you call `slice()` on it you get a real native JS array. I think in this specific case I was getting errors when iterating over the result on line 99 when no report was present.

app/components/agenda/agendaitem/send-to-vp-modal.js
sergiofenoll: The problem is that we only want to assign in the positive case.

When we submit a piece we only send what hasn't been sent. Take the following case:

- Piece A has PDF and DOCX, we send both to VP
- Piece A gets signed
- Signed version of Piece A gets sent to VP, original PDF and DOCX won't be sent because they haven't changed
- We want to send to VP again and now calculate which pieces have been sent (done in this loop)

In the current implementation, in the first iteration we'll get the submitted piece with PDF and DOCX, mark `hasSentPdf` and `hasSentWord` as true. In the next iteration we'll get the signed piece and mark `hasSentSigned`  as true. Once done, we know all three have been sent and that the piece no longer needs to be sent to the VP.

If we assign like you suggest, in the second iteration we'll find:

```js
let hasSentPdf = (submittedPiece.unsignedFile === pdf.uri); // False, because 2nd submitted piece has no unsignedFile
let hasSentWord = (submittedPiece.wordFile === word?.uri); // False, because 2nd submitted piece has no wordFile
let hasSentSigned = (submittedPiece.signedFile === signed?.uri); // True
```

Which will cause the piece to be erroneously resent.

app/models/parliament-submission-activity.js
sergiofenoll: No idea, that was said as such in the story
> Wij hoeven (voorlopig toch) dat veld NIET bij te houden in onze data. M.a.w. de inhoud van het opmerkingen veld wordt meegestuurd naar de vlaams-parlement-sync-service bij indiening, maar daarna niet bijgehouden in het datamodel.

and I didn't know who to follow up with about it so I just implemented it as stated.

tdn: so as discussed in standup, I'm adding the comment part to the model anyway, since it was decided while I was away.

PR #1966, opened 2023-11-30
add checks before regenerating pdf's

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4374

Adds 2 checks in `decision-report-generation.js` to make sure the following conditions are checked before regenerating the decision-reports and minutes.
- Vergadering is afgesloten (laatste agenda is afgesloten)
- Beslissing is vrijgegeven
- Proces voor ondertekenen is opgestart voor beslissingsfiche of notulen

### DIFF
--- a/app/components/meeting/edit-meeting-modal.js
+++ b/app/components/meeting/edit-meeting-modal.js
@@ -365,6 +365,9 @@ export default class MeetingEditMeetingComponent extends Component {
 
   async regenerateMinutes() {
     const minutes = await this.args.meeting.minutes;
+    if (! (await this.decisionReportGeneration.canReplaceMinutes(minutes))) {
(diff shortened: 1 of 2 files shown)


sergiofenoll (changes requested):
Checks for regenerating PDFs look good, but you still need to block off renaming the report as well. If e.g. the meeting number changes, we separately rename the report & regenerate the PDF. The former also needs to be blocked off.

sergiofenoll (approved):

sergiofenoll:
Tests seem unrelated to changes here (related to dropdowns instead) guessing it is/was an issue but shouldn't stop merging this IMO.

PR #1971, opened 2023-12-07
 Feature/kas 1134 generate reports bundle

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-1134

- [ ] https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/17


app/components/agenda/agenda-header/agenda-actions.hbs
```diff
@@ -54,14 +54,32 @@
...
-    <AuButton
+  <AuButton
     data-test-agenda-actions-download-decisions
     @skin="link"
     {{on "click" this.openDownloadDecisions}}
     role="menuitem"
   >
     {{t "download-decisions"}}
   </AuButton>
+  {{#if this.allReports.length}}
```
kobemertens: **question**: Should this action be restricted to certain profiles, since it adds to documents?

kobemertens (approved):
works great on my end :ok_hand:

app/components/agenda/agenda-header/agenda-actions.hbs
tdn: If it writes to documents, only secretarie and admin. We'll fix this during review

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -193,6 +198,48 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+    this.signedReports = await this.store.queryAll('report', {
+      'filter[:has-no:next-piece]': true,
+      'filter[:has:signed-piece-copy]': true,
+      'filter[access-level][:id:]': [INTERN_OVERHEID.id, PUBLIEK.id].join(','),
+      'filter[decision-activity][treatment][agendaitems][agenda][created-for][:id:]':
+        this.args.meeting.id,
+      sort: 'name',
+    });
+  });
+
```
tdn: Two remarks here (this was not clear in the story, so understandable that it wasn't done).

1. both of these need to be filtered on INTERN_OVERHEID
2. This check would be better in de backend.

In fact, as was perhaps not clear in the story, we don't even need the unsigned versions, just the flattened.

sergiofenoll: > both of these need to be filtered on INTERN_OVERHEID

Double check with Johan. The initial implementation only had 1 button to generate a bundle of reports (up to, so including `PUBLIEK`) `INTERN_OVERHEID`, but Johan told me that the users really want 2 things:

- A bundle that only contains up to `INTERN_OVERHEID` signed (flattened) BFs, which is the thing they currently upload in Kaleidos
- A bundle that contains all unsigned BFs (including `VERTROUWELIJK` and other access levels) that the secretary uses during the meeting as a reference. And they have to be unsigned versions because the signing happens after the meeting IIRC.

> This check would be better in de backend.

Yes, but I added it here to keep the API of the service somewhat equivalent (send a list of reports to be bundled/generated together). But it's fine if the report collection is moved to the backend

ValenberghsSven (approved):
Reworked with @tdn 
Perhaps we should do a `queryOne` to check if there is at least 1 signed report before showing the action. The action is always at the user's discretion. They have to ensure that all `intern overheid` decision reports have been signed before using the action.
It is possible to enforce the check in the backend, but might be too strict in exceptional cases.

app/components/agenda/agenda-header/agenda-actions.js
ValenberghsSven: > * eting as a reference. And they have to be unsigned versions because the signing happens after the meeting IIRC.

@tdn This does not match our refactor...

ValenberghsSven: > > both of these need to be filtered on INTERN_OVERHEID
> 
> Double check with Johan. The initial implementation only had 1 button to generate a bundle of reports (up to, so including `PUBLIEK`) `INTERN_OVERHEID`, but Johan told me that the users really want 2 things:
> 
>     * A bundle that only contains up to `INTERN_OVERHEID` signed (flattened) BFs, which is the thing they currently upload in Kaleidos
> 
>     * A bundle that contains all unsigned BFs (including `VERTROUWELIJK` and other access levels) that the secretary uses during the meeting as a reference. And they have to be unsigned versions because the signing happens after the meeting IIRC.
> 
> 
> > This check would be better in de backend.
> 
> Yes, but I added it here to keep the API of the service somewhat equivalent (send a list of reports to be bundled/generated together). But it's fine if the report collection is moved to the backend

@tdn our refactor does not support the second bundle

tdn: We decided to postpone the 2nd functionality due to it being too unclear and risky at the moment.
For example, other users than secretarie would also see this file, but cannot access it. 

With the current priorities in mind, this shouldn't have made it to implementation. Sorry for wasting your time with this!

tdn (approved):

PR #1973, opened 2023-12-08
Feature/kas 4433 send to vp role permissions

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4433

"Doorsturen naar het Vlaams Parlement" button is now shown to Admin (& secretarie & kanselarij) & Kabinetdossierbeheerders profiles. Other profiles can see the sending to VP (with the green banner) but cannot send to VP.

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/439
- [ ] https://github.com/kanselarij-vlaanderen/cache-warmup-service/pull/14

### DIFF
--- a/app/config/permissions.js
+++ b/app/config/permissions.js
@@ -45,6 +45,8 @@ const {
 // - view-only-specific-confidential-documents: allow the viewing of a restricted selection of confidential documents.
 // - search-confidential-cases: allow searching of cases that have at least 1 confidential subcase
 // - search-confidential-documents: allow searching of documents that have vertrouwelijk access level
+// - send-cases-to-vp: allow sending a case's documents to the VP (Flemish Parliament).
+// - send-only-specific-cases-to-vp: allow sending a restricted selection of cases' documents to the VP (Flemish Parliament).
 // - impersonate-users: Use the app as if you were a different user, without logging it with their credentials
 
 const groups = [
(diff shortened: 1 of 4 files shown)


app/config/permissions.js
```diff
@@ -110,6 +113,7 @@ const groups = [
       'view-decisions-before-release',
       'search-confidential-cases',
       'search-confidential-documents',
+      'send-cases-to-vp',
```
tdn: I was under the impression that secretarie/kanselarij would never send these, unless that changed while I was on vacation?

app/config/permissions.js
```diff
@@ -138,6 +142,7 @@ const groups = [
       'view-decisions-before-release',
       'search-confidential-cases',
       'search-confidentnial-documents',
+      'send-cases-to-vp',
```
tdn: I was under the impression that secretarie/kanselarij would never send these, unless that changed while I was on vacation?

app/config/permissions.js
```diff
@@ -168,6 +173,7 @@ const groups = [
       'view-decisions-before-release',
       'search-confidential-cases',
       'search-confidential-documents',
+      'send-cases-to-vp',
```
tdn: OVRB can send cases to VP? That doesn't seem right to me...

tdn (changes requested):
We need to double check during/after the standup tomorrow whether these permissions are correct, because they are different than I remember putting into the story...

tdn (approved):

PR #1975, opened 2023-12-10
KAS-4432: separately updates notas and announcements

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4432

- [ ] merge https://github.com/kanselarij-vlaanderen/minutes-report-generation-service/pull/11


sergiofenoll (changes requested):
These `div`s you added are getting stripped by the editor and the rest of the logic in this PR doesn't work, you never get the separate buttons to update the notas or mededelingen.

The initial "Update content" correctly inserts them but I think the editor is stripping them the next time around, which causes the buttons to not appear.

You should check the actual root cause, the above is just a hypothesis, but if it's the case that the editor is stripping the divs you might have to look intohow prose mirror handles schemas (which are used to strip certain elements).

@brenner-company might know more about this at this point.

kobemertens:
While testing this yesterday, I got the same behavior.
I talked about this with Arne and he suggested using custom ProseMirror Nodes. As I understand, that's the idiomatic way to do things like this. I think at this point we should dive into the ProseMirror docs a little more, so we are not shooting ourselves in the foot by assuming things.

brenner-company:
I think the 'custom ProseMirror Nodes' route @SchrodingersCat00 suggests is the way to go here. Would love the explore the ProseMirror docs a bit more myself, but I'm currently missing extra time to do this 😅

PR #1976, opened 2023-12-14
fix: add inverse relation of signed piece copy

sergiofenoll:
Otherwise when doing `signedPieceCopy.destroyRecord()`, the link from piece to the copy will be left intact.

Related: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/440

### DIFF
--- a/app/models/piece.js
+++ b/app/models/piece.js
@@ -37,11 +37,17 @@ export default class Piece extends Model {
   })
   unsignedPiece;
   @belongsTo('piece', {
-    inverse: null,
+    inverse: 'signedPieceCopyOf',
     async: true,
     polymorphic: true,
   })
   signedPieceCopy;
+  @belongsTo('piece', {
+    inverse: 'signedPieceCopy',
+    async: true,
+    polymorphic: true,
+  })
+  signedPieceCopyOf;
 
   // resources with pieces linked:
 


tdn (approved):
Ok, makes sense. We don't need to mirror this in resources for it to work?

EDIT: skimmed over de "related" PR mention in de description... 🤦‍♂️  so never mind my comment above

ValenberghsSven:
> Otherwise when doing `signedPieceCopy.destroyRecord()`, the link from piece to the copy will be left intact.

hmm, I guess it's possible in these cases (not verified)
-1 the deletion order is wrong (in signature-service)
current
```
      // delete in reverse order of creation
      await signedFile?.destroyRecord();
      await signedPiece?.destroyRecord();
      await signedPieceCopyFile?.destroyRecord();
      await signedPieceCopy?.destroyRecord();
```
should be 
```
      // delete in reverse order of creation
      await signedPieceCopyFile?.destroyRecord();
      await signedPieceCopy?.destroyRecord();
      await signedFile?.destroyRecord();
      await signedPiece?.destroyRecord();
```
`await signedPiece?.destroyRecord();` should cleanup all relations (including to the destroyed signedPieceCopy)

-2 we are only deleting the signedPieceCopy and not the signedPiece

Ok to add the inverse.

sergiofenoll:
@ValenberghsSven signed piece copy is not related to signed piece (piece is linked to signed piece copy only, there's no link between singed piece copy and signed piece). Order of deletes should not matter, `await signedPiece?.destroyRecord()` should not clean up any relationship related to the signed piece copy because it has no relations to that model.

ValenberghsSven (approved):
> @ValenberghsSven signed piece copy is not related to signed piece (piece is linked to signed piece copy only, there's no link between singed piece copy and signed piece). Order of deletes should not matter, await signedPiece?.destroyRecord() should not clean up any relationship related to the signed piece copy because it has no relations to that model.

:man_facepalming: jep, was thinking incorrectly.
never mind

PR #1977, opened 2023-12-15
add casePrefix in withoutDate function as well

kobemertens:
https://kanselarij.atlassian.net/browse/KAS-4429


ValenberghsSven (approved):
@SchrodingersCat00 
Added a trim to the casePrefix.
When VV is present, it resolves to ` VV` (with leading space) causing double spacing in the betreft.
`VV DOC.0001/1 en  VV DOC.0001/2` instead of `VV DOC.0001/1 en VV DOC.0001/2`

Tested this with VV and with normal agendaitems and seems to work nicely

PR #1981, opened 2023-12-20
KAS-4470 upgrade ember-file-upload

ValenberghsSven:
### This PR blocks the ACC release because some users primarily drop files on the uploader in daily workflow.

https://kanselarij.atlassian.net/browse/KAS-4470

Upgraded ember-file-upload to 7.4.0 (fixed) to overcome issues with ember-appuniversum's 7.0.3 dependency.
- not many changes needed (there was a migrations guide, but we are not using any deprecated methods/args it seems)
- The response from uploading changed, needed to transform the response to json.
- dropping files works again
- dropping multiple files was possible on replacing files / adding BIS/TER so added `@multiple={{false}}` where needed. You can still do it but only 1 file will be sliced and uploaded.

potential issue:
The dropzone can't be disabled out of the box so it is possible to drop files when a file selection upload with button is not possible. (might have already been the case prior to breaking)

Don't forget to pull this back into DEV aswell

### DIFF
--- a/app/components/auk/file-uploader.hbs
+++ b/app/components/auk/file-uploader.hbs
@@ -1,3 +1,4 @@
+{{! multiple defaults to true, false means 1 file will be sliced if multiple files are dropped}}
 <FileDropzone
   @queue={{this.fileQueue}}
   @multiple={{@multiple}}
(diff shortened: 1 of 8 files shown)


tdn (approved):
works locally

PR #1982, opened 2023-12-21
KAS-4461: added reloading of parliamentFlow.status after sending to VP and added refresh of route to update alert

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4461

- [ ] merge https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/5


app/controllers/agenda/agendaitems/agendaitem/index.js
```diff
@@ -78,7 +78,11 @@ export default class IndexAgendaitemAgendaitemsAgendaController extends Controll
       this.currentSession.may('manage-agendaitems'),
     );
   }
-
+  @action 
+  refresh() {
+    this.router.refresh(this.router.currentRouteName);
```
tdn: The story specified explicitly to not refresh the page... I'm not sure this is the right way to do this

tdn (approved):
Implemented a smoother user experience during review. Merging as-is now

PR #1989, opened 2024-01-08
fix changing startdate year does not auto update number

ValenberghsSven:
no ticket, noticed this while looking at failed tests.

image

The `numberRepresentation` does not update when changing the agenda startdate from 2024 to 2023.
This is what happens in the test for the search.cy.js setup.
Meaning that we expect to see a generated decision report with title `VR PV 2023`, but we actually get `VR PV 2024`.

Users would not have this issue as they don't really create agenda's for previous/next years.
And if they did, they can manually adjust the numbers.

Adapted the test to only check `VR PV 20` which should work (committed on DEV).
So this extra PR is either "sure" or "nah" to make it automatic (if tests are ok).
Any input welcome.


ValenberghsSven:
> Adapted the test to only check VR PV 20 which should work (committed on DEV).

This didn't work. search doesn't like half numbers.
changed to "VR PV 2024" on DEV to fix the test, but after this PR, the date should be "VR PV 2023"

sergiofenoll (approved):

PR #1996, opened 2024-01-16
Feature/kas 4485 nieuwe manier dossier en procedurestap aanmaken

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4485


brenner-company:
@StuyckRuben Ik ben er even door gegaan en heb de specifieke classes/styling vervangen door een aantal object- (waaronder `au-o-region` & `au-o-layout`) & utility-classes van `ember-appuniversum` ✅

app/components/cases/cases-header.hbs
```diff
@@ -18,6 +18,16 @@
             >
               {{t "add-case"}}
             </AuButton>
+            {{#if this.enableNewCaseCreation}}
+              <AuButton
+                data-test-cases-header-add-case
```
ValenberghsSven: don't copy test selectors, duplicates can break tests. (also no new unused selectors)

app/components/cases/cases-header.hbs
```diff
@@ -18,6 +18,16 @@
             >
               {{t "add-case"}}
             </AuButton>
+            {{#if this.enableNewCaseCreation}}
```
ValenberghsSven: We might as well put this behind a permission instead of putting this button inside the previous permission check. It makes it easier to remove the old one later.
It also prevents accidentally releasing this without permissions.
`(user-may "manage-cases")`  is fine for now.

app/components/cases/cases-header.js
```diff
@@ -1,21 +1,40 @@
...
   @action
-  saveNewCase() {
+  saveNewCase(decisionmakingFlow) {
     this.toggleIsOpenNewCaseModal();
-    this.args.didCreateNewCase(...arguments);
+    if(this.shouldRedirectToSubcase) {
+      this.router.transitionTo('cases.case.subcases.add-subcase', decisionmakingFlow.id);
+    } else {
+      this.args.didCreateNewCase(decisionmakingFlow);
+    }
```
ValenberghsSven: this is adding new code to old code.
We could separate this more so the "switch" from old to new will be easier later.

ValenberghsSven (commented):
limited review for now, getting late. will test more in depth tomorrow

app/components/cases/new-subcase-form.js
```diff
@@ -0,0 +1,195 @@
...
+    this.subcaseName = subcaseType.label;
+  }
+
+  @action
+  onChangeAgendaItemType(selectedAgendaItemType) {
+    this.agendaItemType = selectedAgendaItemType;
+  }
+
+  @task
+  *loadAgendaItemTypes() {
```
sergiofenoll, line 53: I think we should set a default agenda item type too, otherwise users can create a subcase with no type.

Something like this should work:
```js
this.agendaItemType = this.agendaItemTypes.find(
  (type) => type.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA
);
```

app/components/cases/new-subcase-form.hbs
```diff
@@ -0,0 +1,191 @@
...
+                        </AuRadioGroup>
+                      </Fieldset.content>
+                    </AuFieldset>
+                  </AuFormRow>
+                  <AuFormRow>
+                  <AuButton
+                    @skin="secondary"
+                    @icon="copy-paste"
+                    {{on "click" (fn this.saveCase true)}}
+                    @disabled={{this.saveCase.isRunning}}
```
sergiofenoll: `this.saveCase` is a function, not a task, so it will never have an `isRunning` property, so this disabled doesn't work. You can currently click it a bunch of times and it will create a bunch of empty subcases.

sergiofenoll: Perhaps we should conditionally show this button. I don't think it makes a lot of sense to show "Kopieer voorgaande stap" when you're creating the first one.

app/components/cases/new-subcase-form.hbs
```diff
@@ -0,0 +1,191 @@
...
+    <Group>
+      <AuLink
+        @skin="button-secondary"
+        @route="cases.case.subcases"
+      >
+        {{t "cancel"}}
+      </AuLink>
+      <AuButton
+        @skin="primary"
+        @disabled={{this.saveCase.isRunning}}
```
sergiofenoll, line 188: Not a task, this doesn't do anything. Should replace `saveCase` with a task instead.

app/utils/feature-flag.js
```diff
@@ -7,4 +7,11 @@ function enableVlaamsParlement() {
   );
 }
 
-export { enableVlaamsParlement };
\ No newline at end of file
+function enableNewCaseCreation() {
```
sergiofenoll: Perhaps rename to `isEnabledNewCaseCreation`?

`enableXXX` makes it sound like the method will turn on the feature/flag.

translations/nl-be.json
```diff
@@ -5,6 +5,7 @@
   "add-attachments": "Documenten opladen",
   "add-attachment": "Document opladen",
   "add-case": "Nieuw dossier aanmaken",
+  "add-case-new": "Nieuw dossier aanmaken (Nieuw)",
```
sergiofenoll: Lowercase `nieuw` like for `Procedurestap toevoegen (nieuw)`?

sergiofenoll (changes requested):

app/components/cases/cases-header.hbs
```diff
@@ -9,7 +9,7 @@
       </Toolbar.Group>
       <Toolbar.Group @position="right" as |Group|>
         <Group.Item>
-          {{#if (user-may "manage-cases")}}
+          {{#if (and (user-may "manage-cases") this.isEnabledNewCaseCreation)}}
```
sergiofenoll: If `this.isEnabledNewCaseCreation` is false, we'll show no "Nieuw dossier aanmaken" buttons, which doesn't seem intentional.

app/components/cases/new-subcase-form.hbs
```diff
@@ -0,0 +1,194 @@
...
+  <AuToolbar
+    @reverse={{true}}
+    @size="large"
+    @border="top"
+    as |Group|>
+    <Group>
+      <AuLink
+        @skin="button-secondary"
+        @route="cases.case.subcases"
+      >
```
sergiofenoll, line 183: Also disable the cancel button when saving the subcase.

app/components/cases/cases-header.hbs
ValenberghsSven: fixed

app/components/cases/new-subcase-form.hbs
ValenberghsSven: fixed

ValenberghsSven (approved):

PR #1997, opened 2024-01-16
KAS-4452: Replace `store.query` with `store.queryAll`

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4452

Where applicable and necessary calls to `store.query` are replaced with calls to `store.queryAll`. In most of the places this is done just for consistency's sake, though a couple of cases were found where `store.query` was used which could lead too little results being shown or linked.

The ticket contains an attachment with a spreadsheet where each case is listed, specified what was done and why.

### DIFF
--- a/app/components/subcase/subcase-description-panel/subcase-description-edit.js
+++ b/app/components/subcase/subcase-description-panel/subcase-description-edit.js
@@ -102,7 +102,7 @@ export default class SubcaseDescriptionEdit extends Component {
 
   async updateDecisionReports() {
     if (this.enableDigitalAgenda) {
-      const reports = await this.store.query('report', {
+      const reports = await this.store.queryAll('report', {
         'filter[decision-activity][subcase][:id:]': this.args.subcase.id,
         'filter[:has-no:next-piece]': true,
         sort: '-created',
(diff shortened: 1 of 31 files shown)


app/components/subcase/subcase-description-panel/subcase-description-edit.js
```diff
@@ -102,7 +102,7 @@ export default class SubcaseDescriptionEdit extends Component {
 
   async updateDecisionReports() {
     if (this.enableDigitalAgenda) {
-      const reports = await this.store.query('report', {
+      const reports = await this.store.queryAll('report', {
         'filter[decision-activity][subcase][:id:]': this.args.subcase.id,
```
ValenberghsSven, line 106: image

Will add missing `slice()`

ValenberghsSven (approved):
Added 1 missing slice. Changed a test where we had page size in the intercept.
Wait for tests to merge. Unsure if the failed newsitems test is from this PR

ValenberghsSven:
Update: There is an ongoing thread on mu-semtech.
Query to agendaitems is returning different sorting when `page[size]` is removed. (1 `news-item.cy.js` test failed because sorting is different than before)
Could be a bug related to queries that have `LIMIT` (page size present) or not have `LIMIT` (no-pagination-defaults on agendaitem model)
For now, restoring the `page[size]` on queries to agendaitem so we can move this PR along

ValenberghsSven:
Since this was the only failed test (ignoring existing flaky test). i will go ahead and merge

PR #1998, opened 2024-01-19
Feature/kas 4486 ministers en beleidsdomeinen aanmaken bij aanmaken procedurestap

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4486


app/components/cases/mandatee-selector-panel.hbs
```diff
@@ -0,0 +1,45 @@
...
+          <Auk::Toolbar @auto={{true}} as |Toolbar|>
+            <Toolbar.Group @position="left">
+              <h4 class="auk-toolbar__title">
+                {{t "ministers"}}
+              </h4>
+            </Toolbar.Group>
+          </Auk::Toolbar>
+        </Auk::Panel::Header>
+        <Auk::Panel::Body>
+          <Auk::CheckboxTree
```
sergiofenoll: Maybe wrap this in a `Utils::LoadableContent` with `@isLoading={{this.prepareCurrentMinisters.isRunning}}`?

Otherwise when the initial load for ministers is going on you see this, which is a little weird:
image

app/components/cases/mandatee-selector-panel.hbs
```diff
@@ -0,0 +1,45 @@
...
+            @layout="grid"
+          />
+          {{#if this.selectedCurrentMinisters}}
+          <AuHr/>
+          {{/if}}
+          <AuList
+            @divider={{true}}
+            as |Item|
+          >
+            {{#each this.selectedCurrentMinisters as |minister index|}}
```
sergiofenoll: Should this list be sorted (either here in the template or in the component class) to sort the mandatees based on their priority (MP first, Vice MP then, ...)?

app/components/cases/government-areas-panel.hbs
```diff
@@ -0,0 +1,43 @@
...
+        >
+          <Auk::Toolbar @auto={{true}} as |Toolbar|>
+            <Toolbar.Group @position="left">
+              <h4 class="auk-toolbar__title">
+                {{t "government-areas-and-fields"}}
+              </h4>
+            </Toolbar.Group>
+          </Auk::Toolbar>
+        </Auk::Panel::Header>
+        <Auk::Panel::Body class="au-c-panel-grid">
```
sergiofenoll: This could also be wrapped in a `Utils::LoadableContent`

app/components/cases/new-subcase-form.js
```diff
@@ -98,9 +106,10 @@ export default class NewSubcaseForm extends Component {
   }
 
   @action
-  async saveCase(fullCopy) {
+  async saveSubcase(fullCopy) {
+    this.saveDisabled = true;
```
sergiofenoll: Better to use a `task` to check if the function is running than to store state in the component.

sergiofenoll (changes requested):
The panels should be disableable, right now you can click on "Procedurestap aanmaken" and start clicking on extra mandatees and government fields and they will be assigned to the new subcase.

It's better to block off interaction with the form once the process of persisting the subcase starts, because this could easily cause unexpected changes for users.

app/components/cases/mandatee-selector-panel.hbs
StuyckRuben: I believe the list is sorted based on priority?

sergiofenoll: The bottom list is just "order of selection":

image


This may be okay though, perhaps it's not important here.

ValenberghsSven (approved):
Samen met Ruben bekeken en kleine aanpassingen gedaan.

app/components/cases/mandatee-selector-panel.hbs
ValenberghsSven: fixed

PR #2005, opened 2024-01-25
Cypress/kas 4441 new tests automatic update decisions

MathiasZabot:


app/routes/agenda/minutes.js
```diff
@@ -11,8 +11,9 @@ export default class AgendaMinutesRoute extends Route {
   @service signatureService;
 
   async getMandatees() {
+    const { meeting } = this.modelFor('agenda');
     const currentMandatees = await this.mandatees.getMandateesActiveOn.perform(
-      startOfDay(new Date()),
+      startOfDay(meeting.plannedStart),
```
ValenberghsSven, line 16: @tdn During tests I noticed this.
When we calculate the minutes we use the current mandatees (today) even though the `meeting.plannedStart` can be in the past or in the future.
This would only be a problem if editing legacy (won't happen for minutes I think) or every 4 years when the mandatees change.
So checking the `meeting.plannedStart` should be fine and will mostly be for testing purposes (after June, all existing tests still need Overmeer, not the new secretaris)

ValenberghsSven (approved):
Wait for cypress test. Possible flakiness on closing the toasts

app/routes/agenda/minutes.js
tdn: current mandatees is fine, since legacy won't be edited using digital minutes

ValenberghsSven (approved):

PR #2009, opened 2024-01-29
KAS-4315: Subcase submission via the backend

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4315

All the logic for submitting subcases to agendas is now moved to the new backend service.

The idea is to provide basic data to the backend service, the subcase we want to submit and what meeting we want to submit it to, so that the service then determines what data needs to be created.

The service has added checks to ensure you can't submit a subcase that is already on an agenda or submit a subcase on a closed meeting, although the frontend already blocks this off.

Related PRs:
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/448
- [ ] https://github.com/kanselarij-vlaanderen/agenda-submission-service/pull/1


ValenberghsSven (approved):
tested it. Rarely the cache time out was not enough.
Need to check if we might need to up it locally + for cypress.
Merging as-is for now.
Works very good

PR #2012, opened 2024-02-01
KAS-4505: show mandatee fullDisplayName + mandatee start/end for admin/secretarie in subcase

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4505


app/components/mandatees/mandatees-panel/mandatees-panel-view.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+
+export default class MandateesMandateesPanelEditComponent extends Component {
+  @service currentSession;
+
+  get isAdminOrSecretarie() {
+    return this.currentSession.isAdmin || this.currentSession.isSecretarie;
```
sergiofenoll: Should the `Kanselarij` users also have this? Perhaps even `OVRB` (I think they also see mandatees and might be interested in knowing which one is selected)?

Also it might be easier to add this as a permission (a string in `app/config/permissions.js`, e.g. `view-mandatees-with-range`) added for all the profiles and then here you would do `this.currentSession.may('view-mandatees-with-range')`.

sergiofenoll: @tdn Perhaps you know about other profiles that need this.

Other than that, the PR looks great!

ValenberghsSven (changes requested):
`currentSession.isAdmin` should not be used anymore
Permissions is the way to go.

Unsure who needs to see/allowed to see.
I would say:
admin
secretarie
kanselarij


maybe ovrb @tdn  ??  
Their views have their own panel 
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/publications/publication/case/mandatees/mandatees-panel.hbs#L40

In any case, ovrb would be out of scope of what is requested in the ticket

ValenberghsSven (approved):

PR #2015, opened 2024-02-06
KAS-4104 replace toArray with slice

ValenberghsSven:

replace all occurrences of `toArray()` to `slice()`


ValenberghsSven:
Tests are flaky because of backend submitting not clearing cache fast enough.
Suite ran twice, different tests failed so this is PR is fine

PR #2016, opened 2024-02-06
KAS-4104 replace mapBy with map

ValenberghsSven:

replace all occurrences of `mapBy('property')` to `map((x) => (x.property))`
replace 1 occurrence of `filterBy` in combination of `mapBy`  (`filterBy` only used 3 times, not worth its own PR)


ValenberghsSven:
Test are flaky because of backend submitting not clearing cache fast enough.
Suite ran twice, different tests failed so this is PR is fine

PR #2017, opened 2024-02-07
Feature/kas 4104 array like sort by

ValenberghsSven:

replace all occurrences of `sortBy('property')` to `sort((a1, a2) => a1.property - a2.property)` for numbers or `sort((a1, a2) => a1.property.localeCompare(a2.property))` if property is a string
replaced some of `firstObject` and `lastObject` because it was the same line.
replaced some use of async getters with `sortBy` to `@use` resources.


ValenberghsSven:
only existing flaky test.
Manually checked many of the sorts. Looks good

PR #2019, opened 2024-02-07
KAS-4104 explicit polymorphic relations, fix 1 typo

ValenberghsSven:

Make polymorphic relation explicit.
example: if we specify an inverse relation ship with a polymorphic type like `pieces`,  it could be `piece`, `report`, or `minutes`.
This has to be explicit.
Also fixed 1 typo.


ValenberghsSven:
ignoring flaky tests

PR #2022, opened 2024-02-08
KAS-4496: reordering agendaitems now doesnt automatically save after each move + added sum helper

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4496


app/templates/agenda/agendaitems.hbs
```diff
@@ -38,7 +38,11 @@
       <Agenda::AgendaitemSearch
         @searchText={{readonly this.filter}}
         @onSearch={{this.searchAgendaitems}}
-      />
+        @showReorderButton="true"
```
sergiofenoll:
```suggestion
        @showReorderButton={{not this.showModifiedOnly}}
```

When we show modified only we don't allow re-ordering (so e.g. we hide the `ChangesAlert` component), if you still click this button while in this state you get stuck: the button becomes greyed out and the alert that has buttons to cancel is not shown.

Better to just hide this button when in a state we don't support.

app/components/agenda/agenda-overview.hbs
```diff
@@ -1,9 +1,11 @@
 <div class="vlc-agenda-items" ...attributes>
-  {{#if @isEditingOverview}}
+  {{#if (and @isEditingOverview this.canMoveAgendaitems)}}
```
sergiofenoll:
```suggestion
  {{#if this.canMoveAgendaitems}}
```
`this.canMoveAgendaitems` already contains `this.args.isEditingOverview`

app/components/agenda/agenda-overview.hbs
```diff
@@ -105,9 +125,9 @@
     </h3>
   </div>
   <div class="auk-u-mb-4">
-    {{#if @announcements.length}}
+    {{#if this.announcements.length}}
       <ul {{sortable-group onChange=@onReorderAgendaitems}}>
-        {{#each @announcements as |announcement indexitems|}}
+        {{#each this.announcements as |announcement indexitems|}}
```
sergiofenoll:
```suggestion
        {{#each this.announcements as |announcement indexItems|}}
```
Casing here is different, better to match `camelCasing`

app/components/agenda/agenda-overview.hbs
```diff
@@ -127,7 +147,8 @@
               @allowReorderItems={{this.canMoveAgendaitems}}
               @onMove={{this.move}}
               @isFirst={{eq indexitems 0}}
```
sergiofenoll:
```suggestion
              @isFirst={{eq indexItems 0}}
```

app/components/agenda/agenda-overview.hbs
sergiofenoll, line 1: There's a longstanding bug here where when re-ordering in an agenda with notas and announcements, the two lists will get merged. Before it didn't matter because as soon as a user stopped dragging we would save the changes and reset the view, but now it's important to get this working.

Screencast:
[Screencast from 2024-02-12 10-18-52.webm](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/assets/22411874/98a3d14f-e176-4ddb-8ed1-d760a3974923)

This seems to be a known thing with the addon we use, see [this part of their README](https://github.com/adopted-ember-addons/ember-sortable/tree/v4.0.3?tab=readme-ov-file#multiple-ember-sortables-renders-simultaneously)

Here we can fix it by passing in an extra `groupName` argument to all the `sortable-group` and `sortable-items` helpers. So e.g.

```hbs
<ul {{sortable-group groupName="notas" onChange=this.setNotas}}>
  {{each ... }}
    <li
      ...
      {{sortable-item
        groupName="notas"
      }}
```

And `announcements` for the announcements bits.

sergiofenoll (changes requested):

app/helpers/sum.js
sergiofenoll, line 1: We already include the `ember-math-helpers` dependency that implements a helper with the same name and function, this custom helper can be removed.

https://robbiethewagner.github.io/ember-math-helpers/docs/api/helpers/sum

app/components/agenda/agenda-overview/agenda-overview-item.hbs
sergiofenoll, line 1: @brenner-company in deze commit heb ik een paar CSS changes gemaakt zodat de selector rechts staat en om ervoor te zorgen dat de aflijning van alle items nog in orde blijft.

- Voeg `auk-o-flex--vertical-center` klasse toe aan wrapper div in `AgendaOverviewItem` om alles op één lijn te centreren
- Bottom margin verwijderen uit `.vlc-agenda-items__subject` (centreren gebeurt nu via de flex property op de parent div)
- Voeg `.auk-u-flex-grow` helper klasse toe en gebruik deze om de selector wrapper zo breed mogelijk te maken

Laat maar weten als dat zo gemerged mag worden of als je liever hebt dat dit anders aangepakt wordt. 

image

brenner-company: Ziet er top uit ✅

ValenberghsSven (changes requested):
Going to review this PR and make some changes based on what i've found after a small test:
- changing the order of announcements by dragging patches immediately (notas looks ok)
- it looks possible to search and reorder with the filtered items only

possibly more, not mergable as-is

ValenberghsSven (approved):
Changed:
- Cancel reset local variables
- Search agendaitems not possible when changing order
- Changing order button not possible on typing anything in the search bar
- Removed some unnecessary args
- Announcement dragging no longer saves immediately
- Hide newsletter status while reordering
- Renamed changes alert close > confirm so it made more sense with the added cancel method
- Updated cypress test with moved/renamed selectors


Wait for tests before merging. Many tests have changed with the selector update.

PR #2026, opened 2024-02-08
Feature/kas 4493 status op website aanpassen bij mededeling

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4493

- [ ] https://github.com/kanselarij-vlaanderen/agenda-submission-service/pull/3


app/components/agenda/agendaitem/agendaitem-decision-edit.js
```diff
@@ -2,13 +2,20 @@ import Component from '@glimmer/component';
...
 export default class AgendaitemDecisionEditComponent extends Component {
   @tracked decisionActivity = this.args.decisionActivity;
 
   @action
   changeDecisionResultCode(resultCode) {
     this.decisionActivity.set('decisionResultCode', resultCode);
+    if (resultCode === CONSTANTS.DECISION_RESULT_CODE_URIS.UITGESTELD ||
+      resultCode === CONSTANTS.DECISION_RESULT_CODE_URIS.INGETROKKEN)
+      {
+        this.args.onPostponedOrRetracted();
```
sergiofenoll: This action will be triggered as soon as a user selects the uitgesteld/ingetrokken decision in the dropdown (i.e. before saving). We only want to trigger the action when saving.

We should handle this outside of this component, you can just add the check to the `onDecisionEdit` task in `AgendaitemDecision`. Get the decision result code there (it's already been saved here) and then just remove the newsletter if needed.

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
sergiofenoll, line 1: We might need this same logic in the `Subcase::SubcaseTitlesPanel::SubcaseTitlesPanelEdit` component (you can also make subcases/agendaitems confidential there). @tdn is this necessary?

If so, you can base yourself off the following query to find the pertinent agendaitems (from the `agenda-and-subacse-properties-sync` service):
```js
      const agendaitemsOnDesignAgendaToEdit = await this.store.query('agendaitem', {
        'filter[agenda-activity][subcase][:id:]': item.id,
        'filter[agenda][status][:uri:]': CONSTANTS.AGENDA_STATUSSES.DESIGN,
      });
```

sergiofenoll, line 1: Toggling the `inNewsletter` attribute of `this.newsItem` here is fine, the problem is that in this view, the sidebar contains a pill for it (Op de website/Niet op de website). When changing confidentiality, that pill doesn't update because of bad data loading in that component.

I think the simplest route here is to add a `this.router.refresh()` at the end of the save action of this component.

sergiofenoll (changes requested):

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
```diff
@@ -79,5 +86,6 @@ export default class AgendaitemCasePanelEdit extends Component {
       }
     }
     this.args.onSave();
+    this.router.refresh();
```
ValenberghsSven: router refresh to update 1 icon in the side bar is very load heavy.
Let me see if I can fix it to update better

app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
```diff
@@ -24,6 +26,25 @@ export default class SubcaseTitlesPanelEdit extends Component {
     this.args.onCancel();
   }
 
+  @task
+  *removeFromNewsletter() {
+    const agendaitem = yield this.store.queryOne('agendaitem', {
+      'filter[agenda-activity][subcase][:id:]': this.args.subcase.id,
+    });
```
ValenberghsSven: This will get all agendaitem versions and return a random one.
Technically fine, you could have postponed subcases but not typically with announcements. 
Also it wouldn't really make sense to change confidentiality very late in the progress.
But we could query for newsletters directly or at least make sure we only get the current latest newsletter

app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -145,12 +145,26 @@ export default class AgendaitemControls extends Component {
...
+
+  @task
+  *removeFromNewsletter() {
+    const agendaitemType = yield this.args.agendaitem.type;
+    if (agendaitemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT) {
+      const treatment = yield this.args.agendaitem.treatment;
+      const newsItem = yield treatment?.newsItem;
+      if (newsItem) {
+        newsItem.inNewsletter = false;
+      }
```
ValenberghsSven: missing save?

app/components/agenda/agendaitem/agendaitem-decision.js
```diff
@@ -105,6 +105,19 @@ export default class AgendaitemDecisionComponent extends Component {
...
 
+  @task
+  *onPostponedOrRetracted() {
+    const agendaitemType = yield this.args.agendaitem.type;
+    if (agendaitemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT) {
+      const treatment = yield this.args.agendaitem.treatment;
+      const newsItem = yield treatment?.newsItem;
+      if (newsItem) {
+        newsItem.inNewsletter = false;
+        yield newsItem.save();
```
ValenberghsSven: If the `inNewsletter` was already on false we don't have to save it.
so `newsItem?.inNewsletter` would be better.
Smaller use case but it could avoid concurrency

app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
```diff
@@ -24,6 +26,25 @@ export default class SubcaseTitlesPanelEdit extends Component {
...
+      if (agendaitemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT) {
+        const treatment = yield agendaitem.treatment;
+        const newsItem = yield treatment?.newsItem;
+        if (newsItem) {
+          newsItem.inNewsletter = false;
+          yield newsItem.save();
+        }
+      }
+    }
+  }
```
ValenberghsSven: we have 4 copies of the same code. A service call could be beneficial here?

app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
```diff
@@ -49,6 +70,7 @@ export default class SubcaseTitlesPanelEdit extends Component {
     if (this.confidentialChanged && this.args.subcase.confidential) {
       yield this.pieceAccessLevelService.updateDecisionsAccessLevelOfSubcase(this.args.subcase);
       yield this.pieceAccessLevelService.updateSubmissionAccessLevelOfSubcase(this.args.subcase);
+      yield this.removeFromNewsletter.perform();
```
ValenberghsSven: bit of insider knowledge. 
The term "newsletter" in terms of the actual model was replaced to "news-Item / newsItem" a while ago.
some routes and services still have the name but used mainly in reference to the entire collection of newsItems for a meeting.
so a "newsletter" has a bunch of "newsItems"
There are still some "incorrect" names floating around, but we should avoid reintroducing more.

ValenberghsSven (changes requested):
Doing these changes as I'm reviewing.
Will be pushing the updates soon.

We will also might need a change to the `agenda-submission` service.
As we create the announcement newsItem there.
We always default to "true", even if the subcase is already confidential

app/components/subcase/subcase-titles-panel/subcase-titles-panel-edit.js
ValenberghsSven: I changed the duplicate code to a service call in the newsletterService (in favor of a util since utils should not update models imo)
called the method `updateNewsItemVisibility` (technically you could update to true or false here, but in our case we only want to update to false)

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
ValenberghsSven: Managed to fix this quite easily without a full page refresh

ValenberghsSven (approved):
pair review with Tom

PR #2027, opened 2024-02-08
KAS-4104 replace firstObject, lastObject, objectAt and findBy

ValenberghsSven:

replace all occurrences of `findBy('property')`
replaced the remainder of `firstObject`, `lastObject` and `objectAt`


ValenberghsSven:
Going to yolo this one, I was only worried about 1 test (deleting BIS from subcase docs view needs to restore pieces on agendaitel), but with the new subcase view that feature has been removed (batchEdit)
Should we revive that feature, I suggest not offering the BIS removal there, only a full container removal with all versions
Calcuating which agendaitems have to be updated is tedious from this view

PR #2029, opened 2024-02-09
KAS-4386: Refactor `Agenda::Agendaitem::AgendaitemDecision`

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4386

The component is split into two, one for legacy agendas and one for the new "digital" agendas.

Down the line we should further split up the digital component, as all the bits for editing the piece parts can be de-duplicated and placed in separate parametrized components. I chose not to make this change here to keep this PR simple. 

The current refactor is purely structural, code was moved but the logic remains the same. I didn't want to introduce possible logic changes (caused by refactoring the piece part logic) here. If this gets merged, we can start looking at moving off more stuff, without having to be worried about this refactor having broken stuff.


app/components/agenda/agendaitem/decision.hbs
```diff
@@ -0,0 +1,16 @@
+{{! @param {Agendaitem} agendaitem
+    @param {DecisionActivity} decisionAcitivity
+    @param {Object} agendaContext }}
```
kobemertens, line 3: Love the param annotations!

kobemertens (approved):
Tested locally, works great.

ValenberghsSven (approved):
Looks good
New code was mostly auto merged in digital, which makes it easier
I did not apply all new code to legacy (like hiding newsitem based on decision result)
Because legacy is not active edit view. Many of the automations may have to go to prevent any side effects should users attempt to edit legacy

Fixed 1 typo that needed to be fixed in both templates but auto merge only fixed it in digital

Wait for jenkins, but should be ok to merge if nothing strange comes out of that.

PR #2034, opened 2024-02-13
fix: save publication decision in info panel

sergiofenoll:
Otherwise we will never actually save the decision end date.


ValenberghsSven (approved):

ValenberghsSven:
changed base branch to acceptance so we can test before doing the hotfix since we need to test and bump VP as well this week

PR #2041, opened 2024-02-22
KAS-4527: when navigating to subcase now shows a sidebar with other subcases + removed feature flag for new case creation

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4527


app/components/subcases/side-nav/subcase.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @task
+  *loadData() {
+    this.decisionActivity = yield this.store.queryOne('decision-activity', {
+      'filter[treatment][agendaitems][agenda-activity][subcase][:id:]': this.args.subcase.id
+    })
+    this.decisionResultCode = yield this.decisionActivity?.decisionResultCode;
+
+  }
```
kobemertens: I would suggest using the new syntax for tasks which is:
```
loadData = task(async () => { // replace yield with await here })
```
This allows for easier conversion between tasks <-> regular async functions.

kobemertens (changes requested):
Looks good for the most part, just a few functional remarks:
- T̶h̶e̶r̶e̶ i̶s̶ n̶o̶ p̶a̶d̶d̶i̶n̶g̶ a̶t̶ t̶h̶e̶ b̶o̶t̶t̶o̶m̶ o̶f̶ t̶h̶e̶ s̶u̶b̶c̶a̶s̶e̶ r̶o̶u̶t̶e̶. (edit: not needed in the ticket)
- This is not implemented from the ticket: "Als er stappen bij zijn met het type “Principiële goedkeuring”, dan willen we daar ook een telling bij maken, zoals in de mockup “Eerste principiële goedkeuring”, “tweede …” enz."

app/routes/cases/case/subcases.js
```diff
@@ -24,13 +14,15 @@ export default class CasesCaseSubcasesIndexRoute extends Route {
...
-      page: {
-        number: params.page,
-        size: params.size,
-      },
-      sort: '-created',
+      sort: 'created',
+      include:'type'
     };
-    return this.store.query('subcase', queryParams);
+    return this.store.queryAll('subcase', queryParams);
```
sergiofenoll, line 20: We need to re-run this model hook after creating a new subcase, otherwise the new subcase won't show up in the sidebar.

sergiofenoll: I think the reason the sidebar doesn't auto-update is because this `queryAll` is just a collection of subcases. If instead we were returning a relationship, i.e. `decisionmakingFlow.subcases`, it would auto-update because the creation of the new subcase also udpates that relationship in the Ember Data store. With a `queryAll` ED has no way to know that the new subcase is meets the query's filter without re-running the whole filter.

app/components/subcases/side-nav.js
```diff
@@ -0,0 +1,21 @@
...
+    const subcase = subcases.find((subcase) => subcase.id === subcaseId);
+    const number = subcases.indexOf(subcase);
+    const subcaseType = await subcase.type;
+    if (subcaseType) {
+      const label = await subcaseType.label;
+      return label;
+    } else {
+      return this.intl.t("step-and-number", {number: number+1})
+    }
+  }
```
sergiofenoll: Is there a specific reason for calculating the label here instead of inside the `Subcases::SideNav::Subcase` component?

app/components/subcases/side-nav/subcase.js
```diff
@@ -0,0 +1,33 @@
...
+    super(...arguments);
+    this.loadData.perform();
+  }
+
+  @task
+  *loadData() {
+    this.decisionActivity = yield this.store.queryOne('decision-activity', {
+      'filter[treatment][agendaitems][agenda-activity][subcase][:id:]': this.args.subcase.id
+    })
+    this.decisionResultCode = yield this.decisionActivity?.decisionResultCode;
```
sergiofenoll: Not the biggest deal but as with other parts of the app, if we load data this way we lose reactivity.
If you have a subcase with agendaitem-type mededeling and you put it on an agenda, it immediately gets the decision result code "Akte genomen", but we don't see this in the sidebar until you refresh.

app/routes/cases/case/subcases.js
kobemertens: Won't this only display up to 20 subcases then? Or did we override this default somewhere?

sergiofenoll: `queryAll` queries all the subcases. Not suggesting we change it to use the relationship, just pointing out that using the result of a `store.query(All)` means you don't have reactivity, even if it's a direct relationship.

sergiofenoll:
> This is not implemented from the ticket: "Als er stappen bij zijn met het type “Principiële goedkeuring”, dan willen we daar ook een telling bij maken, zoals in de mockup “Eerste principiële goedkeuring”, “tweede …” enz."

From our last sprintplanning, we're not implementing this, unless that has changed again somewhere else.

app/components/subcases/side-nav.js
kobemertens: I'm assuming it's because of the index calculation that needs access to the `subcases` list. Might change the `Subcases::SideNav::Subcase` to accept an index as well.

kobemertens: Fixed in [e3e52d0](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2041/commits/e3e52d0870926eb57316ebcf005d4594c6966166)

sergiofenoll:
Making heads or tails out of all the related changes is getting tricky. Some underlying behaviour here was broken but only gets fixed in the subsequent PR. Re-fixing it here will probably lead to merge conflicts in the later PR.

Instead, I'm just merging this into the `feature/dossiers-revamp` branch and slowly fixing that to the best of my abilities.

PR #2043, opened 2024-02-23
cypress: attempt to reduce flakyness by adding afterEach logout

ValenberghsSven:
Profile-kort-bestek.cy.js and profile-ovrb.cy.js are flaky.
These tests only look at data so they really shouldn't be.
It might be an issue with login info.

From experience, too many tests in 1 file with no `afterEach` where we logout can cause issues.
Because we do login in the `beforeEach`.
This is because each test should be able to run in isolation so sessions should be removed.
This might not always happen correctly in larger files.
So just checking if logging out after each reduces flakyness in this case.


ValenberghsSven:
Well, test still failed on get /500. Need to look into it further.
Perhaps the editing of newsitems in other profile specs is causing issues.
Maybe it's an issue with switching routes too fast, where we grab params from route 1 and try to apply them on route 2
Closing for now

PR #2045, opened 2024-02-26
feature/KAS 4526 nieuw dossier overzicht

kobemertens:
- remove cases loading page
- improve cases header
- add cases status pill
- update dossier page
- add submitters component
- add submitters filter
- use formatlist for submitters
- fix title filter

https://kanselarij.atlassian.net/browse/KAS-4526


app/components/cases/submitters.hbs
```diff
@@ -0,0 +1,11 @@
+{{#if this.loadSubmittersText.isRunning}}
+  <AuLoader />
+{{else}}
+  {{#if this.submitters.length}}
+    <FormatList @list={{this.submitters}} @locale="nl-be" as |formattedPart|>
+      {{formattedPart.value}}
+    </FormatList>
+  {{else}}
+    <p class="au-u-text-center">-</p>
```
sergiofenoll: The dashes look weird. They're correctly centered, it's just that the columns are a lot wider than they need to be, so the dashes don't end up in the middle of the text. I'd remove the center-alignment or try to figure out why the columns are so wide.

image

app/components/cases/cases-header.js
```diff
@@ -28,4 +29,17 @@ export default class CasesHeader extends Component {
     this.isOpenNewCaseAddSubcaseModal = false;
     this.router.transitionTo('cases.case.subcases.add-subcase', decisionmakingFlow.id);
   }
+
+  debouncedSetFilter = restartableTask(async (event) => {
+    if (event.target.value === "") {
+      this.clearFilter();
+      return;
+    }
+    await timeout(350);
```
sergiofenoll: There's an app-wide debounce time in `config/config.js` called `LIVE_SEARCH_DEBOUNCE_TIME`.

app/components/cases/cases-header.hbs
```diff
@@ -1,36 +1,57 @@
...
+            @hideText={{true}}
+            {{on "click" this.clearFilter}}
+          >
+            {{t "clear-filter"}}
+          </AuButton>
+        {{/if}}
+        <AuInput
+          @width="block"
+          @icon="search"
+          @value={{@nameSearchText}}
```
sergiofenoll: `@value` is going to do a two-way binding of the value, so that `@nameSearchText` gets immediately updated as soon as the input changes, which bypasses the `{{on "input" ... }}`.

We don't notice this because we don't actually use the buffer for anything, what actually causes the search to fire is the `setNameSearchText` method we pass in to this component. We can remove the buffer altogether by removing this two-way binding and using `value={{@nameSearchText}}`. This way the initial value is set, but it doesn't auto-update.

Removing the buffer is also beneficial because otherwise we need to add a `setupController` hook to set the `searchTextNameBuffer` field (otherwise when you e.g. reload the page when a filter is set, the input field won't be pre-filled with the right value because the buffer would be empty). By directly using the query param we don't need to take all of that into account.

sergiofenoll: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2045/commits/2924ea0abf3285b13c9c406d3899154e1d094482

app/components/cases/cases-header.js
sergiofenoll: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2045/commits/56f833b0a303ea63b78e9f235f5adffba9b6bf3e

app/components/cases/submitters.hbs
sergiofenoll: Removed the centering. We can always bring it back later.
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2045/commits/cb5b71ae046c402a5a7b1eca74c99f08e042c8b1

PR #2046, opened 2024-02-27
Remove "fill in search term" messages

kobemertens:
When no search results are found, a message is displayed to fill a search term. Since we can search with empty search field, this is not needed anymore. This change removes this message.


ValenberghsSven (approved):
we had 1 TODO in cypress tests where we knew the text was incorrect.

PR #2053, opened 2024-03-04
change incorrect wording

kobemertens:
In the agenda detail view, the agenda header says "Agenda van", this should be "Vergadering van". An agenda does not have a date.

### DIFF
--- a/app/components/agenda/agenda-header.hbs
+++ b/app/components/agenda/agenda-header.hbs
@@ -8,7 +8,7 @@
               class="auk-toolbar-complex__title"
               data-test-agenda-header-title
             >
-              {{t "agenda-of"}}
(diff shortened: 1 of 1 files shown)


ValenberghsSven (commented):
Don't merge until we get some user feedback.

1 comment. the `session-of` is old wording (We used to call a meeting "a session" which was confusing versus the session of a logged in user)
We had `currentSession` for the displayed meeting and `currentUserSession` for the users' session.
`meeting-of` would be more model appropriate these days.
But I realize we still use `session` in many places where we mean the meeting.
We could consider this "out of scope" and have 1 ticket to rename the wording `session` to `meeting` (internal logic only)

kobemertens:
Needs more user feedback

PR #2055, opened 2024-03-06
add decision-result-pill loading state

kobemertens:
Adds a separate loading state to the decision pills. This is better than the current behavior that shows "No decision yet" while loading. This is not what we want.

### DIFF
--- a/app/components/decision/decision-result-pill.hbs
+++ b/app/components/decision/decision-result-pill.hbs
@@ -1,22 +1,33 @@
-{{#if (not @isEditable)}}
-  <AuPill
-    @size="small"
-    @skin={{this.skin}}
-    @icon={{this.icon}}
-    data-test-decision-result-pill
-  >
-    {{this.label}}
-  </AuPill>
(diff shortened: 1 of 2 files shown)


ValenberghsSven:
Yes to the loader, no to the extra dependency needed.
We have to assume the `DecisionResultCode` can be a promise, and treat it as such.
Load it in the constructor with a task. Use the task in the template to check the status.
Works fine locally.
Changed this on development branch: see commit https://github.com/kanselarij-vlaanderen/frontend-kaleidos/commit/714f905c5cb09e27fddefa515ea7558dd381024f

ValenberghsSven:
Solved differently on DEV.
PR no longer needed.

kobemertens:
@ValenberghsSven I don't like how you close this with no further discussion. I agree that the dependency is not strictly needed here. But your approach loses reactivity which is definitely needed for a component as general as this. We have enough cases where we have to reload stuff manually because we don't use reactivity correctly.

PR #2057, opened 2024-03-07
KAS-4567: add `process/browser` webpack polyfill

elpoelma:
### Overview

Adds a `process/browser` polyfill which should be a solution to the error `process is undefined` which shows up when opening the `ember-rdfa-editor`.

##### connected issues and PRs:
KAS-4567

### How to test/reproduce

- Start-up the application
- Open a page which contains the editor
- Notice that the `process is undefined` error is gone and the editor is usable again

### Challenges/uncertainties

Unsure why this error pops-up now, our projects using the editor do not require this polyfill. This can, of course, be caused by a transitive dependency.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -1,6 +1,7 @@
 /* eslint-disable */
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const Funnel = require('broccoli-funnel');
+const webpack = require('webpack');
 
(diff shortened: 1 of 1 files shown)


ValenberghsSven (approved):

ValenberghsSven:
Going to merge this prior to tests being fully run.
The first few tests that actively use the editor have passed.

PR #2060, opened 2024-03-14
feature/KAS 57 terugkerende dossiers

kobemertens:
Implements https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4535


app/components/utils/government-areas/government-areas-panel.hbs
```diff
@@ -24,8 +24,21 @@
...
+            @list={{@subcase.parliamentRetrievalActivity.themes}}
+            @locale="nl-be"
+            as |formattedPart|
+          >
+            {{formattedPart.value}}
+          </FormatList>
+        </p>
+      </div>
+    {{/if}}
+    {{#if @governmentAreas.length}}
```
ValenberghsSven: doing this in the template will now try `subcase.parliamentRetrievalActivity` on every subcase where this util is used. but you only really need it for 1 subcase type.
It would make more sense to get these themes in the parent component and passing those down instead.
To avoid extra calls/load to the frontend

This is also an unspoken "await" in the template.
Better to actively load the data in the `afterModel` in the route (only for 1 type)

Then it would be ok to do:
```
{{#if (and @subcase.isBekrachtiging  @subcase.parliamentRetrievalActivity.themes)}}
```

So you could avoid having another named param

ValenberghsSven (commented):
Small comment before merge.
The rest should be ok

app/components/subcase/detail/regular.hbs
```diff
@@ -0,0 +1,119 @@
...
+    </div>
+    <div class="auk-u-mb-4">
+      <Mandatees::MandateesPanel
+        @referenceDate={{if @meeting @meeting.plannedStart @subcase.created}}
+        @mandatees={{@mandatees}}
+        @submitter={{@submitter}}
+        @emptyStateMessage={{if
+          this.showMandateesNotApplicableMessage
+          (t "not-applicable")
+        }}
```
ValenberghsSven: This was a getter on the controller, this component has no getter / controller does not pas down the result
But we can change this by checking `@subcase.isBekrachtiging` instead of getter
Reasoning: There are normally no mandatees needed for this type of subcase so we want to show "Niet van toepassing" ipv "nog geen ministers toegevoegd"

app/routes/cases/case/subcases/subcase/index.js
```diff
@@ -52,10 +52,13 @@ export default class CasesCaseSubcasesSubcaseIndexRoute extends Route {
       sortedPieces = sortPieces(pieces);
     }
 
+    const subcaseType = await subcase.type;
+
```
ValenberghsSven: you added this but never used it. Unless I overlooked it.
It would be fine just loading this so the promise is at least resolved and always check `model.subcase.isBekrachtiging` (this getter in model works best when the relation is loaded before calling it)

So adding to the model is not really needed

ValenberghsSven (commented):
Caught some stuff on second look

ValenberghsSven:
Forgot to approve, PR reviewed

PR #2061, opened 2024-03-14
Feature/dossiers revamp

ValenberghsSven:
TODO fill out

Combines multiple PR's of the same views for easier test fixing.
All or nothing kind of deal.

KAS-4527
KAS-4529


ValenberghsSven:
4 days spent on fixing tests. good to go now.
There might be 1 more issue that I will look into on DEV:
When switching `agendaitemType` in subcase description edit we should remove/create newsitems depending on the change.
This only failed once in 5 runs where the POST call of a newsitem did not happen.
The use case is quite small, users don't really switch types that often when already on agenda.
So this should be fine and might just be a "cypress too fast" issue

PR #2062, opened 2024-03-15
Feature/kas 4529 cypress test fixes

ValenberghsSven:
Large changelog.
Mostly cypress + some fixes that were needed
Need to expand on those later


ValenberghsSven:
Of the final 6 tests (4 flaky)
1 issue was due to decision report access-level pill being editable before loading the data so we would override new value with the old value.
1 issue was due to case filtering not working in all scenarios (used 1-way binding instead of 2-way binding)
1 issue was page loading timeout
1 issue was trying to click an action that needed a dropdown to be opened first.

Running in jenkins and will most likely merge right after.
Followed by a merge of the base branch into DEV.
Should any other suddenly tests fail it should be fine to just fix on DEV.

ValenberghsSven (commented):
some flakyness exists, merging for now

PR #2065, opened 2024-03-22
KAS-4572: documents with doctype bijlage ter inzage not visible in generateBetreft

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4572

- [ ] merge https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/457


app/utils/decision-minutes-formatting.js
```diff
@@ -36,20 +37,29 @@ function formatDocuments(pieceRecords, isApproval) {
...
   isApproval,
   documents,
   subcaseName = null
 ) {
+  const documentsWithoutBijlageTerInzage = await Promise.all(documents.map(async (document) => {
+    const documentContainer = await document.documentContainer;
+    const type = await documentContainer.type;
+    if (type.uri !== CONSTANTS.DOCUMENT_TYPES.BIJLAGE_TER_INZAGE) {
+      return document;
+    }
```
ValenberghsSven: Type is optional, so if type is not set, we don't return a document.
Better to check
```
if (type?.uri === CONSTANTS.DOCUMENT_TYPES.BIJLAGE_TER_INZAGE) {
return null;
}
```
So we only return null if there is a type and that type is the new "Bijlage ter inzage"

ValenberghsSven (approved):
Going to implement my comment.

Typeless pieces are added
"Bijlage ter inzage" pieces are not

Rest is OK

app/utils/decision-minutes-formatting.js
ValenberghsSven: Fixed it & tested it

PR #2067, opened 2024-03-22
KAS-4540 implement searching old mandatees to add on subcase form

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4540

- Changed the logic to work more with mandatees instead of ministers.
- Added open search possibility
- Added new permission for this open search
- Used 2 separate lists to keep track of current and past mandatees, but template uses a concatenated list
- refactored the hardcoded filter on Jambon's Minister role. We now build a list with Minister-President + all active ministers that have a different person. Should be fine unless we start adding Vice-minister-President role to the `DEFAULT_VISIBLE_ROLES` in the `mandatees` service
- Had to adapt `open-search` because I wanted to clear the list when I added the selected one while keeping the modal open
- simplified the submitter selection since the fix from a few weeks ago made sure we no longer use args `mandatees` and `submitter` directly, as they could be inactive. The preparing tasks clean up those

As i'm writing this I realize: the users may have asked for this view to automatically select the new mandatees from the old mandatees that were present.
And I suggested that we instead show a list "these were the old mandatees" because of mapping issues.
You can map "Jambon" to "Jambon" to the same person, but not "Somers" to "Rutten". 
So the automation would not work in some cases.
Easier to show a list instead.
But I have not done this (yet?) since I just remembered... 
In 99% of cases, we want the new mandatees, so adding them as `past mandatees` is a no go.
Users would then have to select the new mandatee and remove the old mandatee. very error prone

@tdn


tdn (approved):
Pair review 👍

PR #2068, opened 2024-03-25
KAS-4589: edit translation key + accessLevelPill visibility

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4589


sergiofenoll (commented):
image

We're still showing these pills here, is this intended?
@ValenberghsSven, do you know? Otherwise we can ask Johan tomorrow to be sure.

app/components/documents/document-card.js
```diff
@@ -205,6 +207,21 @@ export default class DocumentsDocumentCardComponent extends Component {
...
+    this.loadAccessPillVisibilityData.perform();
+  }
+
+  @task 
+  *loadAccessPillVisibilityData() {
+    const decisionActivity = 
+    this.args.decisionActivity ? this.args.decisionActivity : 
+      yield this.store.queryOne('decision-activity', {
+        'filter[treatment][agendaitems][pieces][:id:]': this.args.piece.id
+      })
```
ValenberghsSven: I wouldn't try to query `decisionActivity` here.
Case and point > a subcase with 20 docs > 20 queries (all different) vs 1 passing of an already loaded `decisionActivity` to `<Documents::DocumentCardList>`

We already have logic in most places to show the alerts, could we do something there?
We could just make a new params for document-card (`showAccessLevel`) and set it to false

translations/nl-be.json
```diff
@@ -1343,7 +1343,7 @@
   "search-case-title": "Zoek op dossiernaam",
   "agendaitem-postponed-no-documents-available": "Dit agendapunt is uitgesteld. Er zijn geen documenten beschikbaar.",
   "agendaitem-retracted-no-documents-available": "Dit agendapunt is ingetrokken. Er zijn geen documenten beschikbaar.",
-  "agendaitem-document-postponed--message": "De documenten van een uitgesteld agendapunt zijn niet zichtbaar voor de gebruikers van een overheidsorganisatie of het Vlaams Parlement. De toegangsrechten 'publiek' of 'intern overheid' zijn op de documenten van een uitgesteld punt niet van toepassing.",
+  "agendaitem-document-postponed--message": "De documenten van uitgestelde agendapunten zijn niet zichtbaar voor de overheid.",
   "agendaitem-document-retracted--message": "De documenten van een ingetrokken agendapunt zijn niet zichtbaar voor de gebruikers van een overheidsorganisatie of het Vlaams Parlement. De toegangsrechten 'publiek' of 'intern overheid' zijn op de documenten van een ingetrokken punt niet van toepassing.",
```
ValenberghsSven: see comments on Jira ticket, also for retracted

ValenberghsSven (changes requested):
- The image Sergio linked would be `<Documents::DocumentBadgeList>` which we might have missed on the original feature
- Retracted also needs to change
- Querying in document card for every single piece is load heavy. In most cases we already know the `decisionActivity`. If we don't have it we need to calculate it prior so we can send a boolean down to documentCard and CardList instead.

There is also no mention of the documents route viewing, does that stay unchanged?
That might the only place where the context would have to queried, .. don't do that as of now.

Check comments, we might have an issue with propagation of `decisionActivity` happens after documents are propagated.
So we need to rethink part of this story

app/components/documents/document-card.js
ValenberghsSven: right now, we don't pass `decisionActivity` for subcase documents, but there can be more than 1.
We should only check the resultCode of the latest one? or check if there is atleast 1 approved/acknowledged one.

afaik `agendaItemTreatment` and `decisionActivity` is not propagated until after release of decisions for other profiles but some documents already are, how are we going to check decision status in that case? Need to look into that.
If we have docs but no `decisionActivity` > also hide the accessLevels? 
I think we haven't thought of this tbh.
**But it will be impossible to check decision status for other profiles prior to releasing decisions.**
@tdn

ValenberghsSven: so profiles will see all documents that are allowed + the access levels right. When we finally release decision, they will be able to see less if the item is postponed, which is weird.

sergiofenoll: This thing is about "uitgestelde agendapunten", specifically ones that are re-used on a later agenda (because users on the original agenda will see an postponed agendaitem with documents that are seemingly readable.

While yes, there's a window where an agendaitem can get postponed and before the agenda is finalized ministers are allowed to see the agendaitem and the documents, it's very unlikely that it will have already been put on a later agenda (and thus, the access level being changed). So I don't think it really matters here.

tdn: As it's implemented now, the accessLevelPill does idd not show on agenda's where the decisions haven't been released yet, which is not what was intended here.

Sergio's point is valid, so I think it's more important to show the pill if we can't be sure than to hide it

tdn (approved):

ValenberghsSven (approved):

PR #2071, opened 2024-03-26
Feature/kas 4590 ratifications

tdn:


tdn:
This still has 2 annoying bugs I couldn't fix:
1. autofill on submission is not done properly
2. sometimes the mandatee selector panel doesn't load the ratifiedBy mandatees right away

To be fixed on ACCEPT.

PR #2049, opened 2024-02-27
Feature/upgrade ember acm idm

sergiofenoll:
Bumps our dependency of https://github.com/lblod/ember-acmidm-login to v2.0.0 and adapts the code slightly to use the new version.

PR #2073, opened 2024-03-28
ember acm-idm upgrade

ValenberghsSven:
This code is from https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2049 but without the package-lock base.
All kudos goes to  @sergiofenoll. I just copy-pasted.

Does not work as-is, something needs to change in dispatcher.
But the popup itsme is now a full page like you would see in other apps.
Works fine until we get redirected to our app

new ticket: 
https://kanselarij.atlassian.net/browse/KAS-4906

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/500


ValenberghsSven:
**got it to work on DEV**, couldn't get it working locally to a localhost address but who cares about that.
Added minor error handling to it

ValenberghsSven:
FYI, still need to test for mobile with this branch. Just checking that all goes smoothly and if any styling needs to happen

tdn (approved):

PR #2074, opened 2024-03-29
feature/KAS 4566 parliament case view

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4566
Related PR's:
- https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/24
- https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/462


app/components/cases/parliament/description-panel.hbs
```diff
@@ -0,0 +1,111 @@
...
+          </div>
+        </Auk::Toolbar::Item>
+      </Auk::Toolbar::Group>
+      <Auk::Toolbar::Group @position="right">
+        <Auk::Toolbar::Item>
+          <AuLinkExternal
+            @icon="external"
+            @iconAlignment="right"
+            href={{concat
+              "https://www.vlaamsparlement.be/nl/parlementaire-documenten/parlementaire-initiatieven/"
```
tdn: We're using this URI on a few places now. Maybe we want to put it in a constant?

app/components/cases/parliament/description-panel.hbs
```diff
@@ -0,0 +1,111 @@
...
+            class="vlc-status-timeline__item vlc-status-timeline__item--neutral"
+          >
+            <div class="vlc-status-timeline__dot"></div>
+            <div class="vlc-status-timeline__text">
+              {{t "accepted-in-plenary-meeting-on"}}
+              <AuLinkExternal
+                @icon="external"
+                @iconAlignment="right"
+                href={{concat
+                  "https://www.vlaamsparlement.be/nl/parlementaire-documenten/parlementaire-initiatieven/"
```
tdn: see comment above. Would put this in a constant

app/components/cases/parliament/description-panel.hbs
```diff
@@ -0,0 +1,111 @@
...
+            <li
+              class="vlc-status-timeline__item vlc-status-timeline__item--neutral"
+            >
+              <div class="vlc-status-timeline__dot"></div>
+              <div class="vlc-status-timeline__text">
+                {{t "ratified-and-announced-on"}}
+                <AuLinkExternal
+                  @icon="external"
+                  @iconAlignment="right"
+                  href="https://www.vlaamsparlement.be/nl/parlementaire-documenten/parlementaire-initiatieven/1671548"
```
tdn: hardcoded URI???

app/components/cases/parliament/description-panel.hbs
```diff
@@ -0,0 +1,111 @@
...
+              <li
+                class="vlc-status-timeline__item vlc-status-timeline__item--neutral"
+              >
+                <div class="vlc-status-timeline__dot"></div>
+                <div class="vlc-status-timeline__text">
+                  {{t "published-in-belgian-official-gazette"}}
+                  <AuLinkExternal
+                    @icon="external"
+                    @iconAlignment="right"
+                    href="https://www.vlaamsparlement.be/nl/parlementaire-documenten/parlementaire-initiatieven/1671548"
```
tdn: also a hardcoded URI, I assume for testing?

app/components/cases/parliament/description-panel.js
```diff
@@ -0,0 +1,59 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { service } from '@ember/service';
+import { task } from 'ember-concurrency';
+
+export default class CasesParliamentDescriptionPanelComponent extends Component {
+  @service store;
+
+  @tracked publciationStatus;
```
tdn: typo: should be `publicationStatus` (apply everywhere in the code of course)

tdn: Also duplicate with the ` @tracked publicationStatus;` a few lines below?

app/components/cases/parliament/newest-documents-panel.js
```diff
@@ -0,0 +1,58 @@
...
+
+    const submittedPieces = await this.store.queryAll('submitted-piece', {
+      'filter[parliament-submission-activity][:id:]': submissionActivities
+        .map((x) => x.id)
+        .join(','),
+      sort: '-parliament-submission-activity.start-date',
+      include: 'piece',
+    });
+
+    const uniquePieces = removeDuplicatePiece(submittedPieces.slice());
```
tdn: I'm wondering what the removeDuplicatePieces will do in case a `file` was changed and re-submitted. 
It's probably fine since the `piece` itself is the same, and the `file` isn't loaded until someone actually clicks on the `piece` to go to the document-viewer, but we should definitely verify this during testing

app/components/cases/subcases/new-subcase-form.hbs
```diff
@@ -1,38 +1,3 @@
...
-              @maxLength={{250}}
-            />
-            -
-            {{t "create-subcase"}}
-          </h4>
-        </div>
-      </Auk::Toolbar::Item>
-    </Auk::Toolbar::Group>
-  </Auk::Toolbar>
-</Auk::Navbar>
```
tdn, line 34: Was this removed on purpose, or the result of a merge with `development`?

app/components/on-this-page-links.hbs
```diff
@@ -0,0 +1,18 @@
+<div>
+  <div class="c-anchor-nav-holder">
+    <p>{{t "on-this-page"}}: </p>
+    <ul class="au-c-anchor-nav">
+      {{#each @items as |item|}}
+        <li>
+          <a
+            href={{concat "#" item.elementId}}
+            class="js-anchor"
+            onclick={{fn this.scrollToFragment item.elementId}}
```
tdn, line 10: Does this take into account @ValenberghsSven 's fix to ensure the scroll works and doesn't add whitespace at the viewport bottom as it did in the past?

app/components/on-this-page-links.js
```diff
@@ -0,0 +1,17 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+
+export default class OnThisPageLinksComponent extends Component {
+  @action
+  scrollToFragment(fragName, e) {
+    // chrome does not work correctly with fragment scrolling, custom override needed
```
tdn, line 7: I know @ValenberghsSven did some work to fix this recently for the other subcase views. Is this double, or something else?

tdn: Aha, I see this functionality was moved from the subcase view to this component. So am I correct to assume it includes the earlier fix?

app/controllers/cases/case/subcases/subcase.js
```diff
@@ -7,6 +7,6 @@ export default class CasesCaseSubcasesSubcaseController extends Controller {
 
   @action
   refreshSubcases() {
-    this.router.refresh('cases.case.subcases');
+    this.router.refresh('cases.case.index');
```
tdn: hmm. Don't we want to keep the subcases route in this case? Otherwise the view might jump from the subcase view to the parliament/publication view after an action has been completed

app/controllers/cases/case/subcases/add-subcase.js
```diff
@@ -7,6 +7,6 @@ export default class CasesCaseSubcasesAddSubcaseController extends Controller {
 
   @action
   onCreateSubcase() {
-    this.router.refresh('cases.case.subcases');
+    this.router.refresh('cases.case.index');
```
tdn: This is called after a `subcase` is created. In theory this would mean it will always route to the latest subcase, which would be fine. But we do need to verify this during testing. If not, this needs to stay on the `subcases` route.

translations/nl-be.json
```diff
@@ -1354,6 +1354,18 @@
...
   "automatic-check-performed-possible-pieces-missing":"Op basis van een automatische controle lijken de volgende documenten te ontbreken. Wil je ze toch indienen? Laad ze dan op en stuur ze vervolgens door.",
+  "case-info": "Dossier info",
+  "design-of": "Ontwerp van",
+  "last-statusses": "Laatste statussen",
+  "to-flemish-parliament": "Naar Vlaams Parlement",
+  "number": "Nummer",
+  "accepted-in-plenary-meeting-on": "Aangenomen in plenaire vergadering op",
+  "ratified-and-announced-on": "Bekrachtigd en afgekondigd op",
+  "published-in-belgian-official-gazette": "Bekendgemaakt in het Belgisch Staatsblad op",
+  "updated-on-date": "Geüpdatet op {date}",
```
tdn: We're using "Bijgewerkt op" throughout the application. I'd use that instead of "Geüpdatet op"

tdn (changes requested):
Some minor things to fix and a few reminders to ensure correct functionality during testing.

Could not run locally due to a build error:

```
SyntaxError: frontend-kaleidos\node_modules\reactiveweb\dist\function.js: Static class blocks are not enabled. Please add `@babel/plugin-transform-class-static-block` to your configuration.
  37 |  * State container that represents the asynchrony of a `trackedFunction`
  38 |  */
> 39 | class State {
     | ^
  40 |   static {
  41 |     g(this.prototype, "data", [tracked], function () {
  42 |       return null;
```

If this is due to a new library, this needs to be fixed before we can continue review.

app/components/cases/subcases/new-subcase-form.hbs
kobemertens: It's from a merge, not sure why this is detected as a change.

app/controllers/cases/case/subcases/add-subcase.js
kobemertens: Will revert to subcase route

app/components/on-this-page-links.js
kobemertens: Probably, might need @ValenberghsSven to confirm

tdn (approved):
Fixed and tested locally during pair review.

@ValenberghsSven when you're back on Monday, please do a final check to make sure none of your earlier fixes are overwritten.

If it's fine, this can be merged to DEV, and then all tickets on DEV may be added to the ACCEPT release.

Don't forget to bump the `vlaams-parlement-sync` service and restart migrations and resource for the `app-kaleidos` changes

app/routes/cases/case/parliament-flow/index.js
```diff
@@ -0,0 +1,32 @@
...
+      subcase: this.store.queryOne('parliament-subcase', {
+        'filter[parliament-flow][:id:]': parliamentFlow.id,
+      }),
+      submissionActivities: this.store.queryAll(
+        'parliament-submission-activity',
+        {
+          'filter[parliament-subcase][parliament-flow][:id:]':
+            parliamentFlow.id,
+        }
+      ),
```
ValenberghsSven: is this used anywhere? 
DocumentVersionsPanel could use this `submissionActivities` (needs sorting), but doesn't.
Same with `subcase` loaded here, there seems to be no direct use of `model.subcase` or `model.submissionActivities`

app/routes/cases/case.js
```diff
@@ -1,12 +1,42 @@
...
+      ),
+      parliamentFlow: this.store.queryOne('parliament-flow', {
+        'filter[case][decisionmaking-flow][:id:]': params.id,
+        include: 'status,parliament-subcase',
+      }),
+      subcases: this.store.queryAll('subcase', {
+        'filter[decisionmaking-flow][:id:]': params.id,
+        sort: 'created',
+        include: 'type',
+      }),
```
ValenberghsSven: If we have to load this much data in the RSVP, we could just make the model async and do some preloading instead?

app/routes/cases/case.js
```diff
@@ -1,12 +1,42 @@
...
+        params.id
+      ),
+      latestParliamentSubmissionActivity: this.store.queryOne(
+        'parliament-submission-activity',
+        {
+          'filter[parliament-subcase][parliament-flow][case][decisionmaking-flow][:id:]':
+            params.id,
+          sort: '-start-date',
+        }
+      ),
```
ValenberghsSven: queries to `'parliament-submission-activity'`:
Here you do a query of one with `decisionmakingFlow` id
Later, you do another query to all submissions with `parliamentFlow` id
Then in document versions, another query to all submissions with `parliamentSubcase` id
So 3 calls for every case, none of them reused.
So i'm guessing we need all submissions.
Query parliament flow first (if that doesn't exists we know we can stop checking?)
Can we query them here (sorted and from parliament flow), grab the latest and put both in the model?

app/components/parliament/status-pill.hbs
```diff
@@ -0,0 +1,24 @@
+{{#let (load @parliamentFlow.status) as |statusPromise|}}
```
ValenberghsSven: how is this different than an await.
`{{#if (await @parliamentFlow.status)}}` > loading, else check the status does the same thing.

app/components/on-this-page-links.js
ValenberghsSven: yes this component version works fine.
The styling was already a bit off, spacing in weird places can now be fixed here rather than in multiple code copies.
This is good :+1:

app/components/cases/parliament/newest-documents-panel.hbs
```diff
@@ -0,0 +1,61 @@
+<Auk::Panel data-test-subcase-description-panel>
```
ValenberghsSven: copied test selector

app/components/cases/parliament/document-versions-panel.js
```diff
@@ -0,0 +1,19 @@
...
+  get submissionActivities() {
+    return new TrackedAsyncData(
+      this.store.queryAll('parliament-submission-activity', {
+        'filter[parliament-subcase][parliament-flow][:id:]':
+          this.args.parliamentFlow.id,
+        sort: '-start-date'
+      })
+    );
+  }
+}
```
ValenberghsSven: why not data down? args model has a similar list?

app/components/cases/parliament/description-panel.js
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    this.publicationStatus = publicationStatus;
+    this.latestMeeting = latestMeeting;
+    this.latestPublicationActivity = latestPublicationActivity;
+  });
+
+  get parliamentCaseLink() {
+    return `https://www.vlaamsparlement.be/nl/parlementaire-documenten/parlementaire-initiatieven/${this.args.parliamentFlow.parliamentId}`;
+  }
```
ValenberghsSven: Hardcoded link.
A shortcut already exists `CONSTANTS.PARLIAMENT_CASE_URL_BASE`

app/components/cases/parliament/newest-documents-panel.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  loadData = task(async () => {
+    const subcase = await this.args.parliamentFlow.parliamentSubcase;
+    const submissionActivities = await this.store.queryAll(
+      'parliament-submission-activity',
+      {
+        'filter[parliament-subcase][:id:]': subcase.id,
+      }
+    );
```
ValenberghsSven: from args?

app/components/cases/parliament/description-panel.hbs
```diff
@@ -0,0 +1,111 @@
...
+      </Auk::Toolbar::Group>
+    </Auk::Toolbar>
+  </Auk::Panel::Header>
+  <Auk::Panel::Body>
+    <AuHeading @level="4" @skin="4">
+      {{#if @latestSubmissionActivity}}
+        {{@latestSubmissionActivity.title}}
+      {{else}}
+        {{@latestRetrievalActivity.title}}
+      {{/if}}
```
ValenberghsSven, line 32: testcase on DEV has no title. Looks a bit weird. Do we need to fallback to something?

app/components/cases/subcases/new-subcase-form.hbs
ValenberghsSven: something got removed here.
Before the add-subcase route was a different route with the same header as case index route
So it looked the same but it wasn't, it was a copy
I had to adapt a test to fix that.
Now we are showing the true header again, so i'll revert that part of the test changes
The add-subcase is now yielded in the template on the right side, next to the sidebar

ValenberghsSven: With the add-subcase now rendered next to the sidebar, it is much easier to navigate away during the creation. Might need to look into that later.
Hide the sidebar on some occasions?

app/components/subcases/side-nav.js
```diff
@@ -0,0 +1,34 @@
+import Component from '@glimmer/component';
+
+export default class SubcasesSideNavComponent extends Component {
+  get items() {
+    let items = [];
+
```
ValenberghsSven: this get is called many times, triggering a visual change because decision activity results show "no result yet" everytime you move between subcases.
Ideally, we only have to load this once and rerender when adding/removing subcases or reloading the page

translations/nl-be.json
```diff
@@ -1354,6 +1354,17 @@
...
+  "design-of": "Ontwerp van",
+  "last-statusses": "Laatste statussen",
+  "to-flemish-parliament": "Naar Vlaams Parlement",
+  "number": "Nummer",
+  "accepted-in-plenary-meeting-on": "Aangenomen in plenaire vergadering op",
+  "ratified-and-announced-on": "Bekrachtigd en afgekondigd op",
+  "published-in-belgian-official-gazette": "Bekendgemaakt in het Belgisch Staatsblad op",
+  "sent-on-date-by-person": "Verzonden op {date} door {person}",
+  "no-statusses-available-yet": "Er zijn nog geen statussen beschikbaar",
+  "no-documents-submitted-to-vp-yet": "Er werden nog geen documenten ingediend bij het Vlaams Parlement",
```
ValenberghsSven, line 1367: "doorgestuurd naar"? Not fully in the loop of naming here, but I know there was a discussion between these 2 terms

ValenberghsSven (approved):
Approved, but I have made changes (this is urgent after all)
small changes:
- No `AuLoader`, Karel made a wrapper around it so we keep using `Auk::Loader` for now.
- No copied duplicate test selectors
- a few fixes where prettier formatting does weird stuff `>{{date-phrase newline ... newline }}</` instead of `> newline {{date-phrase  ....  }} newline </`
- hardcoded link prefix to use of a constant
- replace `(load model.status .. as promise   > promise.isResolved ..` to `(await @parliamentFlow.status)` (loader shows just the same)
- `AbbreviatedText` was complaining of not having text as a result of the moving and data still needed to be loaded. Wrapped with a task await in the template
- changed a test to reflect that the `add-subcase` route now no longer has a copy of the header instead of the actual header.

big changes:
- Reworked cases.case route model from a RSVP hash to an async model: Changed the order of queries, first we check if there is a parlement flow. If yes we use that flow to get the submission-activites > use that to get latest.
Now we we have to option to pass down that list of `submission-acitivities`  instead of querying it multiple times (with different params). Incread load on initial entry but reduced load overal
- passed the `submission-acitivities` down to 2 component who needed that list
- Made the document-versions-panel template-only by doing that
- changed the sidenav getter to be a tracked property instead. So it is only calculated when the model refreshes normally.
- Added a loader page to ensure ember switches to that page when the async model is loading
- changed the model of cases.case.parliament-flow to get the `submission-acitivities` from parent route and pass them down instead of doing a new query + cleaned up query for subcase. This result was not being used.
- removed a duplicate `SubcasesOverviewHeader` as a result of moving that header to a higher route. You would only see this when there were 0 subcases. Looked quite funky and test failed on it.
image

ValenberghsSven:
The added package `"ember-async-data": "^1.0.3",` could be removed I think?
I don't fully get the specific purpose that this offers that we can't already solve with tasks, trackedFunctions, awaiting, yielding and whatnot.
The only import of it has been removed with the passed down data.

translations/nl-be.json
kobemertens: Not sure, there is no design for this, feel free to change.

kobemertens:
@ValenberghsSven Mostly agree with the comments. Only this one I don't see how it would be better:

> Reworked cases.case route model from a RSVP hash to an async model: Changed the order of queries, first we check if there is a parlement flow. If yes we use that flow to get the submission-activites > use that to get latest.

How can this be faster than running all queries in parallel?

ValenberghsSven:
> @ValenberghsSven Mostly agree with the comments. Only this one I don't see how it would be better:
> 
> > Reworked cases.case route model from a RSVP hash to an async model: Changed the order of queries, first we check if there is a parlement flow. If yes we use that flow to get the submission-activites > use that to get latest.
> 
> How can this be faster than running all queries in parallel?

We could run some of those queries in parallel.
I didn't think about that to be honest.

With the reduced load I meant that we don't execute the 2 `Latest... activity` when there is no `parliament-flow` and the passing down of the submissionActivites to avoid having to query them multiple times with a different uncached query (one with decisionMakingFlow id, one with parliament flow id, one with parliament subcase id)

kobemertens:
@ValenberghsSven Yes good catch about the `submission-activity`'s, loading the `submission-activity`'s in the route's model is better. But we should load what we can in parallel. I've updated the code

PR #2075, opened 2024-04-02
KAS-4571: if mandatee is not active, use person to try and find new mandatee

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4571


app/components/cases/subcases/mandatee-selector-panel.js
```diff
@@ -119,6 +119,18 @@ export default class MandateeSelectorPanel extends Component {
         );
         if (mandatee) {
           selectedMandatees.push(mandatee);
+        } else {
+          const oldMandateePerson = oldMandatee.person;
```
sergiofenoll: This needs to be `yield`ed. This code needs to move out of a `forEach` and into a normal `for`-loop.

sergiofenoll: Resolved in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/commit/0ecc0fc8cde6b1784b4cfb9113bdf432f10f32fb

ValenberghsSven:
The ticket was quite empty, the "best effort" idea was prior to the "toon vorige ministers".
Using that newer feature, we can now always show all mandatees (either the person is still active or not).
It improves our "best effort" to "maximum effort", we also use the submitter even if it's old.
So when making a new subcase you have the full context of the mandatees of the old one.
If all else fails, we will still fall back to just selecting the first of the list.

Since we do have a permission to use past mandatees I put some of the automation behind that permission.
 
@tdn Didn't think you would mind this? the ticket mentions "Alternatief: we zouden ook gewoon een melding kunnen tonen met de vorige mandatarissen er in."  which is solved this way.

ValenberghsSven (approved):
Very biased approval :D
Test it with no mandatees, single active mandatees, single old mandatees, combined old & new mandatees.
If we can match the person to an active mandatee, we select an active mandatee.
If we can't > just show the old mandatee (if permission)
The submitter should always be the same person it used to be either active or old.

@StuyckRuben I've added stuff that was not in the ticket, perhaps you want to check if you agree.

PR #2078, opened 2024-04-05
KAS-4598: now updates agendaitem number when changing agendaitemtype

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4598


app/components/subcase/description-panel/edit.js
```diff
@@ -122,13 +122,31 @@ export default class SubcaseDescriptionEdit extends Component {
...
+    const agendaItem = await this.store.queryOne('agendaitem', {
+      'filter[agenda-activity][subcase][:id:]': this.args.subcase.id,
+      'filter[:has-no:next-version]': 't',
+    })
+
+    const latestAgendaitemOfType = await this.store.queryOne('agendaitem', {
+      'filter[agenda][agendaitems][agenda-activity][subcase][:id:]': this.args.subcase.id,
+      'filter[:has-no:next-version]': 't',
+      'filter[agenda-activity][subcase][agenda-item-type][:uri:]': agendaitemType.uri,
+      sort: 'number'
```
sergiofenoll: Needs to be `-number`. Sorting is ascending by default (so 1, 2, 3, ... will be returned). We want the largest number first, so we want to make it descending, using `-`.

See: https://github.com/mu-semtech/mu-cl-resources?tab=readme-ov-file#sorting

app/components/subcase/description-panel/edit.js
```diff
@@ -122,13 +122,31 @@ export default class SubcaseDescriptionEdit extends Component {
...
+    const agendaitemType = await this.args.subcase.agendaItemType;
+    const agendaItem = await this.store.queryOne('agendaitem', {
+      'filter[agenda-activity][subcase][:id:]': this.args.subcase.id,
+      'filter[:has-no:next-version]': 't',
+    })
+
+    const latestAgendaitemOfType = await this.store.queryOne('agendaitem', {
+      'filter[agenda][agendaitems][agenda-activity][subcase][:id:]': this.args.subcase.id,
+      'filter[:has-no:next-version]': 't',
+      'filter[agenda-activity][subcase][agenda-item-type][:uri:]': agendaitemType.uri,
```
sergiofenoll: `'filter[type][:uri:]': agendaitemType.uri` also works and is less heavy on the query.

app/components/subcase/description-panel/edit.js
```diff
@@ -122,13 +122,31 @@ export default class SubcaseDescriptionEdit extends Component {
     if (this.agendaItemType.uri !== oldAgendaItemType.uri) {
       await this.updateNewsletterAfterRemarkChange();
       await this.updateDecisionReports();
+      await this.recalculateAgendaitemNumber();
```
sergiofenoll: Recalculating after the agendaitem has been saved won't always work.

For example: this subcase is a nota with number 7. There are 3 mededelingen on the agenda. When switching the type to mededeling, the query in this function (after fixing it) will find this subcase (as 7 is larger than 3) and set its number to 8.

We need to perform this calculation before saving, or at the very least filter out the current agendaitem when trying to find the number.

sergiofenoll (changes requested):

ValenberghsSven:
When changing type, we change the decision name ('punt'; 'mededeling') but not the number.
Going to try and fix this quick

ValenberghsSven (approved):
New number is now also used to generate the new piece name.
Tested locally.
Can merge if jenkins is green

ValenberghsSven:
only flaky tests are also on dev

PR #2080, opened 2024-04-09
chore: add redirect route for subcase

sergiofenoll:
Going to /procedurestap/:subcase_id will redirect you to the right case/subcase page.


app/routes/subcase.js
```diff
@@ -0,0 +1,16 @@
...
+export default class SubcaseRoute extends Route {
+  @service store;
+  @service router;
+
+  model(params) {
+    return this.store.findRecord('subcase', params.subcase_id);
+  }
+
+  async redirect(model) {
+    const decisionmakingFlow = await model.decisionmakingFlow;
```
ValenberghsSven, line 13: faulty id given > redirect to "dossiers"? will model be undefined/null?
will `model.decisionmakingFlow` error in that case?

ValenberghsSven (commented):
I guess this route can't hurt, but how would users get here?
Or is this more a developer thing where I query the subcase and don't want to query more to get the decisionMakingFlow id?

In any case, i'm all for it if it works

sergiofenoll:
> I guess this route can't hurt, but how would users get here? Or is this more a developer thing where I query the subcase and don't want to query more to get the decisionMakingFlow id?
> 
> In any case, i'm all for it if it works

The latter. It's just like the [`agendaitem` route](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/routes/agendaitem.js)

ValenberghsSven:
> > I guess this route can't hurt, but how would users get here? Or is this more a developer thing where I query the subcase and don't want to query more to get the decisionMakingFlow id?
> > In any case, i'm all for it if it works
> 
> The latter. It's just like the [`agendaitem` route](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/routes/agendaitem.js)

never used that one, didn't know it existed

ValenberghsSven (approved):

PR #2082, opened 2024-04-10
KAS-4619

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4619

A normal "definitieve goedkeuring" is the same as before:
image

A "definitieve goedkeuring" that was created from a VP flow now includes a text "Van het Vlaams Parlement" and the remarks if there were any:
image

### DIFF
--- a/app/components/subcase/description-panel/view.hbs
+++ b/app/components/subcase/description-panel/view.hbs
@@ -58,6 +58,9 @@
       {{#if @subcase.title}}
         <p data-test-subcase-description-title>{{capitalize @subcase.title}}</p>
       {{/if}}
+      {{#if @subcase.parliamentRetrievalActivity}}
+        <AuHelpText @skin="secondary">{{t "from-parliament"}}</AuHelpText>
+      {{/if}}
     </Auk::Panel::Body>
     <Auk::Panel::Body>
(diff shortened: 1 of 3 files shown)


ValenberghsSven (commented):
I guess this makes sense, but I don't fully see the full picture.
Right now we have 2 sets of panels:
- 1 for normal subcases ( now with additions for this ticket)
- 1 for "bekrachtiging" types only

Should we change it so we have a clear difference between user created subcase and parliament subcase?
The distinction between the two is becomes less every week

sergiofenoll:
> I guess this makes sense, but I don't fully see the full picture. Right now we have 2 sets of panels:
> 
>     * 1 for normal subcases ( now with additions for this ticket)
> 
>     * 1 for "bekrachtiging" types only
> 
> 
> Should we change it so we have a clear difference between user created subcase and parliament subcase? The distinction between the two is becomes less every week

I agree it's getting muddy, I think it's something we should take up with design. We might've missed the fact that "definitieve goedkeuring" subcases also come in from the Flemish Parliament during user consultation.

ValenberghsSven:
fyi, not requesting a change, just thinking out loud. will review this later

ValenberghsSven (approved):

PR #2083, opened 2024-04-10
Feature/add authority domain

sergiofenoll:
Flag on https://kanselarij.atlassian.net/browse/KAS-4535

Display authority domain and themes on all subcases coming from VP

Related PRs:

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/464
- [x] https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/27

### DIFF
--- a/app/components/utils/government-areas/government-areas-panel.hbs
+++ b/app/components/utils/government-areas/government-areas-panel.hbs
@@ -25,9 +25,15 @@
     </Auk::Toolbar>
   </Auk::Panel::Header>
   <Auk::Panel::Body @noPadding={{if @governmentAreas.length true false}}>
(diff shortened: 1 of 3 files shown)


ValenberghsSven (approved):
I guess this is fine.
The check on `isBekrachtiging` was added to ensure we don't do try to check the `parliamentRetrievalActivity` on all other subcases we have.

In any case, this is a util where args `@subcase` is only passed in on the subcase views.
but in theory you could show those fields also on agendaitem that way.
Wonder if we should just calculate `parliamentRetrievalActivity` earlier and pass that as args instead.
that way there is no silent await for `@subcase.parliamentRetrievalActivity` but we could do it in the model (or afterModel) of the route

Not right now ofc.
This change is ok for fixing the current flags

ValenberghsSven (approved):

PR #2085, opened 2024-04-15
KAS-4562 document stamping

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4562

Related PR's:
- https://github.com/kanselarij-vlaanderen/document-stamping-service/pull/2
- https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/466
- https://github.com/kanselarij-vlaanderen/file-bundling-service/pull/14


app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -381,6 +382,15 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
     }
   }
 
+  async stampDocuments() {
```
tdn: This would be better located in a service. Fetches in components are not best practice.

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -365,6 +365,7 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
     }
     try {
       await approveAgendaAndCloseMeeting(this.args.currentAgenda);
+      await this.stampDocuments();
```
tdn: I'm a bit worried that we don't inform the user of what we're doing here. I'll modify the translation key of the modal that says the agenda is being approved

app/components/agenda/agenda-header/agenda-version-actions.js
tdn: Also, this is now only when the meeting is closed. We want to trigger this for every agenda version

tdn (changes requested):

tdn:
Works well now, just needs polling to make sure the stamping job has finished and report this to the user

ValenberghsSven:
please see https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2105 for merging

PR #2090, opened 2024-04-19
KAS-4624: add check on signPreparationActivity or signCompletionActivity to disable "nieuwe versie opladen"

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4624

### DIFF
--- a/app/components/documents/document-card.js
+++ b/app/components/documents/document-card.js
@@ -57,6 +57,8 @@ export default class DocumentsDocumentCardComponent extends Component {
 
   @tracked hasSignFlow = false;
   @tracked hasMarkedSignFlow = false;
+  @tracked hasSignPreparationActivity = false;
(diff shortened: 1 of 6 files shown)


app/components/documents/document-card.js
```diff
@@ -138,7 +140,9 @@ export default class DocumentsDocumentCardComponent extends Component {
     return (
       !this.args.hideUpload
         && (!this.hasSignFlow
-            || (this.hasMarkedSignFlow && !!this.args.decisionActivity))
+            || ((this.hasMarkedSignFlow || this.hasSignPreparationActivity || !!this.signCompletionActivity )
```
tdn: This doesn't make sense. A signFlow with a preparationActivity will always have a markingActivity, so this statement will never reach the 2nd OR.

tdn (approved):
works locally

PR #2091, opened 2024-04-19
KAS-4586: regenerate pdf & change name of BF after aproving agenda

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4586


app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -176,6 +178,36 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
...
+        const agendaitem = await this.store.queryOne('agendaitem', {
+          'filter[:has-no:next-version]': true,
+          'filter[treatment][decision-activity][report][:id:]': report.id,
+        });
+        const documentContainer = await report.documentContainer;
+        const pieces = await documentContainer.pieces;
+        report.name = await generateReportName(agendaitem, this.args.meeting, pieces.length);
+        await report.belongsTo('file').reload();
+        await report.save();
+      }));
```
ValenberghsSven: this part is a carbon copy of what happens in the `edit-meeting-modal`.
So it works yes, but I will probably move this to a service call then.
something like 
`this.decisionReportGeneration.regenerateDecisionReportNamesForMeeting.perform(this.args.meeting.id);` (the name might need work)
followed by actual regeneration of the reports similar to `edit-meeting-modal` (need to see)

ValenberghsSven (commented):
Technically fine, but will refactor due to duplicate code now.

ValenberghsSven (approved):
Tested locally after changes.
Some weird scenarios are possible in combination with digital agenda but it seems ok.
The contents of the decision report may be wrong, but reasonably, the report should only be made when all documents are added and approved at least once.

Wait for jenkins at least once

tdn (approved):
Tested locally. Fixes the scenario from the story

PR #2096, opened 2024-04-24
Feature/kas 4607

kobemertens:
Ticket:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?assignee=63c552d46178fcc941d857c9&selectedIssue=KAS-4607

Related PR's:
- https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/468

New repository:
- https://github.com/kanselarij-vlaanderen/document-naming-service


app/components/agenda/agenda-header/agenda-check.js
```diff
@@ -0,0 +1,55 @@
...
+    const mappingsMap = new Map(
+      mappings.map(({ uri, newTitle }) => [uri, newTitle])
+    );
+
+    return mappingsMap;
+  });
+
+  fileNameMappings = trackedTask(this, this.getFileNameMappings);
+
+  get notaGroups() {
```
tdn: This basically duplicates/re-writes `groupAgendaitemsOnGroupName` in the agenda-service: https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/services/agenda-service.js#L180-L213

This would cause us to have 2 places to maintain the same action, so I'd prefer if the existing one would be reused.

app/components/agenda/agenda-header/agenda-version-actions.hbs
```diff
@@ -7,12 +7,12 @@
   class="auk-u-hidden@print"
 >
   {{#if @currentAgenda.status.isDesignAgenda}}
-  <AuButton data-test-agenda-version-actions-approve-agenda @skin="link" {{on "click" this.openConfirmApproveAgenda}}
+  <AuButton data-test-agenda-version-actions-approve-agenda @skin="link" {{on "click" this.openConfirmApproveAgendaSidebar}}
```
tdn: The naming of `openConfirmApproveAgendaSidebar` and `showConfirmForApprovingAgendaSidebar` is a bit confusing here. Why "sidebar"? 

The property controls whether or not the `AgendaCheck` component is shown, so `openAgendaCheck` and `showAgendaCheck` would be more logical, no?

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -346,6 +359,17 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
     this.showConfirmForApprovingAgendaAndClosingMeeting = false;
   }
 
+  @action
+  async openConfirmApproveAgendaAndCloseMeetingSidebar() {
```
tdn: Again with the "sidebar" here... This is really confusing naming.

tdn: Ok, after testing locally I see where the "sidebar" naming is coming from. It was implemented as a sidebar on usertest, even though it behaves more like a modal. Still, I'd prefer the naming for these properties and functions to be a bit more in line with the actual components it shows/hides.

app/routes/agenda.js
```diff
@@ -27,10 +29,36 @@ export default class AgendaRoute extends Route {
...
+      if (type.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA) {
+        notas.push(agendaitem);
+      } else {
+        announcements.push(agendaitem);
+      }
+    }
+
+    // Load agendaitem pieces
+    await Promise.all(
+      agendaitems.map(this.throttledLoadingService.loadPieces.perform)
```
tdn: Any idea what the performance impact of loading this here will be for an agenda with 100+ agendaitems and even more pieces? 

I'm not sure, but I think there's a good reason we didn't load them in the model here at this stage. 

Furthermore, I'm not even sure this needs to be done at all. The story specified clearly that the `document-naming` service should load and provide all the information the `AgendaCheck` needs. Now, we're still loading it in the frontend, which is much slower.

Unless I'm misunderstanding this?

tdn (changes requested):
I think it's best that we pair-review this and fix some issues.

I added some comments about naming and code duplication, which are minor, but I fear there might be major issues too. For example I'm worried about the loading of all agendaitems and pieces in the `agenda` route, and locally, the modal does not show the `piece` names for all agenda's (see screenshots)

agenda-overview-shows-pieces-for-agendaitem-2

agenda-check-not-showing-pieces-for-agendaitem-2

tdn:
The bug I reported in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2096#pullrequestreview-2052174096 was due to a long-outstanding bug in the print view apparently. This commit also solves https://kanselarij.atlassian.net/browse/KAS-4447 https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2096/commits/9c37f21bb988c22619cca26fabc256bee7b889c7

tdn (approved):

tdn:
@ValenberghsSven tests will need to be changed to account for the agendacheck before the approval modal is shown

ValenberghsSven:
please see https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2105 for merging

PR #2098, opened 2024-04-25
KAS-4631: changed title from "Laatste statussen" to "historiek" and from "Nieuwste documenten" to "Recentste documenten+ check proper date for "aangenomen in plenaire vergadering + no longer showing links for "bekrachtigd en afgekondigd" and "gepubliceerd in Belgisch Staatsblad"

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4631

- [ ] merge: https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/33


app/components/cases/parliament/description-panel.js
```diff
@@ -48,7 +48,7 @@ export default class CasesParliamentDescriptionPanelComponent extends Component
     );
 
     const publicationStatus = await this.store.queryOne('publication-status', {
-      'filter[publications][case][decisionmaking-flow][:id:]':
```
ValenberghsSven, line 51: Hmm, I think I was wrong here. `publications` is correct.
This query is now just throwing errors.
In these types of queries, we have to use the relation as defined in the model
in this case for the model of `publication-status`
`@hasMany('publication-flow', { inverse: 'status', async: true }) publications;`

So what is the reason for not showing that timeline during demo?
Well, it might be because we have put this timeline behind the the if check of `meeting.agenda`
We might have forgotten to finalize the agenda on ACC?
Did we have a published publication-flow in that decisionmakingFlow?
Did we have multiple publication-flows in that decisionmakingFlow? this query grabs a random one because we have no sorting.
We could instead query all publications with a filter on `CONSTANTS.PUBLICATION_STATUSES.PUBLISHED`


Another possibility could be a cache issue.
Where the result of this query is kept in cache because the path is so distant and we were getting old status here.
In frontend, we normally get this status by doing `publicationFlow.status`.

tdn (approved):

PR #2099, opened 2024-04-26
KAS-4104 enforce known deprecations throw errors

ValenberghsSven:

added deprecation-workflow to prevent re-introduction of known deprecations


ValenberghsSven:
the template  helper `(map-by)` seems to be an issue when using with an unresolved relation.
cannot use A() with proxyArray.
ex. `{{join ", " (map-by "idName" @publicationFlow.numacNumbers)}}`

Assuming this happens in the helper itself, which we have no control over.
So we either have to make sure to await it (right now I did it in the template) or change the helper.

ValenberghsSven:
solved some more issues where we silently await relation in a template
This is no longer allowed.
We have to make sure the relation is resolved in the template.
Either by using a `@tracked` or a  `{{#let (await thing.proxyRelation) as |resolvedRelation|}}`
Or some other forms of properly awaiting.
Also when creating a record, don't use proxies from other models, make sure they are resolved

ValenberghsSven:
Tests were green, now flaky because flaky on DEV.
Merging

PR #2100, opened 2024-05-02
cypress/KAS-4643-expand-profile-tests

MathiasZabot:


ValenberghsSven (approved):
Setup is in ZIP now, profile tests are green.
Need to cleanup the comments and good to merge

PR #2103, opened 2024-05-15
feature/KAS-4629: updateBetreftContent & updateBeslissingenContent now generate text for bekrachtiging

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4629


sergiofenoll (approved):

app/components/agenda/agendaitem/decision/digital.js
```diff
@@ -312,12 +312,24 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component
...
     const agendaActivity = await this.args.agendaitem.agendaActivity;
     const subcase = await agendaActivity?.subcase;
-    const newBetreftContent = await generateBetreft(shortTitle,
-      title,
-      this.args.agendaitem.isApproval,
-      documents,
-      subcase?.subcaseName
-    );
+    let newBetreftContent;
+    if (subcase?.isBekrachtiging) {
```
ValenberghsSven, line 321: `subcase?.isBekrachtiging` (sync getter) needs `await subcase?.type` to resolve.

app/components/agenda/agendaitem/decision/digital.js
```diff
@@ -361,7 +375,9 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component
         newBeslissingHtmlContent = this.intl.t('retracted-item-decision');
         break;
       default:
-        if (this.args.agendaitem.isApproval) {
+        if (subcase?.isBekrachtiging) {
+          newBeslissingHtmlContent = "De Vlaamse Regering bekrachtigt bovengenoemd decreet en beslist over te gaan tot de afkondiging ervan.";
+        } else if (this.args.agendaitem.isApproval) {
```
ValenberghsSven: look into `generateApprovalText`, we might want something similar for both decisions and minutes?

Also, scopecreep: Could we use `generateApprovalText` here as well. duplicate code atm

app/components/agenda/agendaitem/decision/digital.js
```diff
@@ -312,12 +312,24 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component
...
-      this.args.agendaitem.isApproval,
-      documents,
-      subcase?.subcaseName
-    );
+    let newBetreftContent;
+    if (subcase?.isBekrachtiging) {
+      const ratification = await subcase.ratification;
+      newBetreftContent = await generateBetreft(
+        subcase.shortTitle,
+        null,
```
ValenberghsSven: We would probably still want to `agendaitem.shorttitle` here.
subcase title and agendaitem should be the same regardless

ValenberghsSven (changes requested):
`generateBetreft` is used also in minutes which this does not cover.

Is there a way we could only call `generateBetreft`once in digital version?
Perhaps we could pass in `subcase` instead of just `subcaseName` and implement ratification logic in 1 spot rather then 2.
The ratification document could be collected in the `generateBetreft` and added to the list of documents if present.

app/components/agenda/agendaitem/decision/digital.js
ValenberghsSven: "we might want something similar for both decisions and minutes? Since it is only 1 translation without special cases, opted to just use a translation in both scenarios instead of a util.

app/components/agenda/agendaitem/decision/digital.js
```diff
@@ -315,15 +315,29 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component
...
+    if (subcase?.isBekrachtiging) {
+      const ratification = await subcase.ratification;
+      newBetreftContent = await generateBetreft(
+        shortTitle,
+        title,
+        this.args.agendaitem.isApproval,
+        ratification ? [...documents, ratification] : documents,
+        null, // This seems unused on ratifications
+        agendaitemType,
+      );
```
ValenberghsSven, line 330: wondering If I should move the logic into `generateBetreft`

tdn (approved):
pair reviewed

PR #2107, opened 2024-05-16
KAS-4647: Enable new mu-auth for testing

ValenberghsSven:
Don't merge this until approved


app/components/m-header.js
```diff
@@ -81,7 +81,7 @@ export default class MHeader extends Component {
   @task
   *loadRoles() {
     this.roles = yield this.store.queryAll('role', {
-      'filter[concept-scheme][:uri:]': CONSTANTS.CONCEPT_SCHEMES.USER_ROLES,
```
ValenberghsSven, line 84: is this right? we are doing `filter[concept-scheme][:uri:]': CONSTANTS.CONCEPT_SCHEMES.USER_ROLES,` in a few other places aswell.

tdn (approved):
After vigorous testing, this can be merged

PR #2108, opened 2024-05-21
KAS-4635: sort list of agenda-activities from new to old, re-added links on certain agenda-activities, sidebar should retrieve correct decision-activity, should show most recent agenda at "Geagendeerd op"

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4635

### DIFF
--- a/app/components/subcase/timeline.js
+++ b/app/components/subcase/timeline.js
@@ -15,6 +15,24 @@ export default class SubcaseTimeline extends Component {
     this.loadSubcasePhases.perform();
   }
 
+
+  get processedPhases() {
+    return this.phases.map((phase) => {
+      if (phase.label.includes(this.intl.t('activity-phase-approved-on-agenda'))) {
+        const index = phase.label.indexOf('agenda');
+        const before = phase.label.substring(0, index);
+        const after = phase.label.substring(index, phase.label.length);
+        return {
+          ...phase,
+          beforeText: before,
+          linkText: `${after} ${dateFormat(phase.date, 'dd MMMM yyyy')}`,
+        };
+      } else {
+        return phase;
+      }
+    });
+  }
+
   subcaseTimelineItemText = (timelineItem) => {
     const label = timelineItem.label || '';
     const date = timelineItem.date;
@@ -30,7 +48,7 @@ export default class SubcaseTimeline extends Component {
     const phases = [];
     const sortedAgendaActivities = yield this.store.queryAll('agenda-activity', {
       'filter[subcase][:id:]': this.args.subcase.id,
-      sort: 'start-date',
+      sort: '-start-date',
     });
 
     for (const activity of sortedAgendaActivities.slice()) {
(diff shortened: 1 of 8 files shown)


app/components/subcases/side-nav/subcase.js
```diff
@@ -28,7 +28,7 @@ export default class SubcaseSideNavSubcaseComponent extends Component {
 
   loadData = task(async () => {
     this.decisionActivity = await this.store.queryOne('decision-activity', {
-      'filter[treatment][agendaitems][agenda-activity][subcase][:id:]':
+      'filter[subcase][:id:]':
         this.args.subcase.id,
```
ValenberghsSven: There can be more than 1 (postponed). we need latest here? add sort on `-start-date`

app/components/subcase/timeline.js
```diff
@@ -30,7 +48,7 @@ export default class SubcaseTimeline extends Component {
     const phases = [];
     const sortedAgendaActivities = yield this.store.queryAll('agenda-activity', {
       'filter[subcase][:id:]': this.args.subcase.id,
-      sort: 'start-date',
+      sort: '-start-date',
     });
```
ValenberghsSven: this reverse seems wrong. full reverse happens later.
So we build from oldest to newest and reverse everything.

Or this + we reverse phases per agenda-activity, which we are not doing right now.

app/components/subcase/timeline.js
```diff
@@ -54,6 +72,9 @@ export default class SubcaseTimeline extends Component {
         phases.push({
           label: this.intl.t('activity-phase-approved-on-agenda'),
           date: meeting.plannedStart,
+          meetingId: meeting.id,
+          agendaId: agenda.id,
+          agendaitemId: firstAgendaitemOfActivity.id
         });
```
ValenberghsSven: Side effect of this data: we always get the first agendaitem (on agenda A). but we want the link to point to the most recent agenda instead.

ValenberghsSven (changes requested):
Pair reviewed with @tdn 
Found some things that were not fully correct.
Ended up changing stuff on the fly.
- no longer creating the link in the top of the panel because it was duplicate in a lot of cases.
- changed some rules so that there is always 1 link in the phases per agenda (even if it was not approved yet), changed the label to say "designagenda" in that case)
- cleanup some code which was no longer used due to change above + code that was just no longer used.
- changed tests that check the timeline

app/components/subcases/side-nav/subcase.js
ValenberghsSven: Done this

app/components/subcase/timeline.js
ValenberghsSven: Done the former. build full list from oldest to newest, from submission to approval per block. fully reversed that list.

app/components/subcase/timeline.js
ValenberghsSven: Several scenarios came up.
Editors got links to approved agenda, even if there was a designagenda that came after.
So 2 links right above each other to different agenda versions.
Was confusing.
Changed it so the link always goes to known latest agenda (design or approved)
Viewers only have approved so that is fine.
Editors only have 1 link and it always points to the most recent agenda (design or approved)
Only side effect > we used to only show "ingediend" when not approved yet, now we also show "geagendeerd op" which seems fine because we need the link now (removal of the top link). Only editors will see this.

ValenberghsSven (approved):
After changes
Postpone.cy.js was the only non flaky. tested locally and was ok

PR #2109, opened 2024-05-21
KAS-4652: now shows different message in AgendaGroupHeader if the subcase.type is bekrachtiging

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4652


sergiofenoll (approved):
Looks good and works as expected

app/controllers/agenda/agendaitems.js
```diff
@@ -118,14 +118,20 @@ export default class AgendaAgendaitemsController extends Controller {
     let currentAgendaitemGroup;
     for (const agendaitem of agendaitemsArray) {
       yield animationFrame(); // Computationally heavy task. This keeps the interface alive
+      const subcase = yield this.store.queryOne('subcase', {
+        'filter[agenda-activities][agendaitems][:id:]': agendaitem.id,
+        include: 'type'
+      });
```
ValenberghsSven: I get we need the subcase here.
A query is an option.
In the child route agendaitems/agendaitem we also need the subcase.
to get the subcase there we do
```
    this.agendaActivity = await model.agendaActivity;
    this.subcase = await this.agendaActivity?.subcase;
```
this means we are now loading the similar data in 2 ways which will slow down the general loading.
It will be fine locally but with large agendas it could mean 100 extra uncached data calls.
I think anyway. it makes more sense to reuse the same calls (either query or relation calls)

app/components/agenda/printable-agenda/list-section/item-group.js
```diff
@@ -1,9 +1,26 @@
...
+    super(...arguments);
+    this.loadData.perform();
+  }
+
+  @task
+  *loadData() {
+    const agendaitem = this.args.items.at(0);
+    const agendaActivity = yield agendaitem.agendaActivity;
+    const subcase = yield agendaActivity?.subcase;
+    this.isBekrachtiging = subcase?.isBekrachtiging;
```
ValenberghsSven: `subcase?.isBekrachtiging;` is an sort of sync getter on the modal that uses possible unloaded async data (type)
so we should do `yield subcase?.type;` when we want to use it.

app/services/agenda-service.js
```diff
@@ -214,6 +214,16 @@ export default class AgendaService extends Service {
...
+        const subcase = await this.store.queryOne('subcase', {
+          'filter[agenda-activities][agendaitems][:id:]': agendaitem.id,
+          include: 'type'
+        });
+
+        if (subcase && subcase.isBekrachtiging) {
+          agendaitem.set('groupName', this.intl.t('ratification-title'));
+          currentAgendaitemGroupName = this.intl.t('ratification-title');
+        }
+
```
ValenberghsSven: This has nothing to do with agenda views but with KB views.
Is this needed here? Not mentioned in ticket so unsure what it is supposed to do.

ValenberghsSven (changes requested):
not fully ok it seems. some comments
I have 2 ratifications locally.

There is some logic to include if a ratification subcase does have mandatees but that messes up the print view.
image
image

Should we show Jambon here or ratification?

app/components/agenda/printable-agenda/list-section/item-group.hbs
```diff
@@ -10,23 +10,34 @@
     {{else}}
       <div class="l-printable-agenda__faux-item">
         <div class="auk-u-non-breakable-group@print">
-          {{#let this.sortedMandateesFromFirst as |mandatees|}}
+          {{#if this.isBekrachtiging}}
```
tdn: `if (and (eq this.sortedMandateesFromFirst.length 0) (not this.isBekrachtiging))` here as well?

app/services/agenda-service.js
tdn: It would seem logical to me, but given Kort Bestek's reluctance to change, and this being a request from secretarie, we might want to leave this out for now

app/components/agenda/printable-agenda/list-section/item-group.hbs
tdn: Test this by creating 2 ratifications at the end of the agenda. Give the last one 1 or more mandatees, and the mandatee titles should show in the print view. Now it shows "Bekrachtiging en afkondiging" twice.

image

tdn (changes requested):
Some small comments left to be resolved. 
Does work nicely already

tdn (approved):

PR #2111, opened 2024-05-22
KAS-4628 add ratification also in subcase views

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4628
https://kanselarij.atlassian.net/browse/KAS-4638

Code that would happen pre submitting to meeting now done when received from VP.
Ratification also needed to be available PRE submit in subcase views.
So made a new component (copy) of the current one.
Since it makes more sense that this component is under `<Subcase::` I reused my new component also in agenda views.
Only change was to remove some unused params to determine "draft access levels". 

In ratification subcases, we don't show mandatee panel but show the ratification doc and signing mandatee panel instead.

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/474
- [x] https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/35


ValenberghsSven (commented):
Tested/Pair review with Tom

PR #2113, opened 2024-05-23
add restamping in frontend

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4640


tdn (approved):
pair review. Worked well

PR #2114, opened 2024-05-24
feat: expand supported file types for conversion

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4627

Expands the config in the frontend to allow more files to be sent to the conversion service.

Conversion service (well, the API we use behind the screens) supports converting DOC(X), ODP, XLS(X), and ODT files (and some more that we're not interested in). So we also send these files for conversion.

TBD if we want/need to merge this → draft status


tdn (approved):
Tested and working with all file types. Nice!

PR #2115, opened 2024-05-27
KAS-4649: create publication-table-row component for publication/shortlist + made date decision sortable + add link to agendaitem

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4649


app/components/publications/shortlist/publications-table-row.js
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor() {
+    super(...arguments);
+    this.loadData.perform();
+  }
+
+  @task
+  *loadData() {
+    const piece = yield this.args.row;
+    const agendaitems = yield piece.agendaitems;
```
tdn: Better to query the store here for caching etc.

app/components/publications/shortlist/publications-table-row.hbs
```diff
@@ -16,16 +23,16 @@
...
   {{else}}
     -
-  {{/each}}
+  {{/if}}
 </td>
 <td class="auk-table__col--1">
   {{date this.decisionDate}}
 </td>
-<td class="auk-table__col--1">
+<td class="auk-table__col--1 u-hide-on-print">
```
ValenberghsSven, line 35: column was still showing on print. might need input for @brenner-company for this.
I used a class from `wu-legacy.scss`

app/components/publications/shortlist/publications-table-row.hbs
```diff
@@ -16,16 +23,16 @@
...
 </td>
 <td class="auk-table__col--1">
-  {{#each this.mandateePersonNames as |name|}}
-    {{name}}
+  {{#if this.mandateePersonNames}}
+    {{join ", " this.mandateePersonNames}}
   {{else}}
     -
-  {{/each}}
+  {{/if}}
```
ValenberghsSven, line 30: names were stitched together, now they are comma separated.

app/components/publications/shortlist/publications-table-row.hbs
```diff
@@ -1,13 +1,20 @@
...
+  <AuLink
+    @route="agenda.agendaitems.agendaitem"
+    @models={{array this.meetingId this.agendaId this.agendaitem.id}}
+  >
     {{this.formattedAgendaNumber}}
   </AuLink>
 </td>
-<td class="auk-table__col--6">
-  <AuLink data-test-route-publications-overview-shortlist-row-name @route="document" @model={{@row.id}}>
+<td class="auk-table__col--7">
```
ValenberghsSven, line 9: changed 6 to 7, we had 11 instead of 12 (max 12 I think)

app/components/publications/shortlist/publications-table-row.js
```diff
@@ -19,30 +22,19 @@ export default class ShortlistPublicationsTableRowComponent extends Component {
...
-        this.agendaitem = maybeAgendaitem;
-        break;
-      }
-    }
+    this.agendaitem = yield this.store.queryOne('agendaitem', {
+      'filter[pieces][:id:]': piece.id,
+      'filter[:has-no:next-version]': 't',
+      sort: '-created',
+    });
+
```
ValenberghsSven, line 30: avoid `piece.agendaitems` since we can get 1 to 5 agendaitems meaning more looping and data loading to check what is the latest agendaitem.

app/components/publications/shortlist/publications-table-row.js
```diff
@@ -19,30 +22,19 @@ export default class ShortlistPublicationsTableRowComponent extends Component {
...
-    const agendaitemNumberLength = agendaitemNumber.toString().length;
-    if (agendaitemNumberLength < 4) {
-      for (let index = 0; index < 4 - (agendaitemNumberLength * 1); index++) {
-        this.formattedAgendaNumber += "0"
-      }
-      this.formattedAgendaNumber += agendaitemNumber.toString();
-    } else {
-      this.formattedAgendaNumber = this.agendaitem.number
-    }
+    this.formattedAgendaNumber = addLeadingZeros(agendaitemNumber, 4);
```
ValenberghsSven, line 37: simpler way to pad with zeroes

app/components/publications/shortlist/publications-table-row.js
```diff
@@ -51,16 +43,19 @@ export default class ShortlistPublicationsTableRowComponent extends Component {
   *loadMandateeData() {
     const mandatees = yield this.agendaitem.mandatees;
     const persons = yield Promise.all(
-      mandatees.map((mandatee) => mandatee.person)
+      mandatees
+        .slice()
+        .sort((m1, m2) => m1.priority - m2.priority)
+        .map((mandatee) => mandatee.person)
```
ValenberghsSven, line 49: sort mandatees on priority here

app/controllers/publications/overview/shortlist.js
```diff
@@ -16,14 +16,21 @@ export default class PublicationsOverviewShortlistController extends Controller
...
   agendaitemForPublication;
   decisionActivityForPublication;
   caseForPublication;
   mandateesForPublication;
 
+  @action
+  changeSort() {
+    this.sort = (this.sort === this.SORT_FIELD) ? `-${this.SORT_FIELD}` : this.SORT_FIELD;
+  }
+
```
ValenberghsSven, line 33: the default sorting table has 3 options:
Up, down or "no sorting" applied.
We don't want the "no sorting" option, so make it only toggle between the two

app/routes/publications/overview/shortlist.js
```diff
@@ -22,20 +22,15 @@ export default class PublicationsOverviewShortlistRoute extends Route {
...
-          'agendaitems.agenda.next-version',
-          'agendaitems.agenda.created-for',
-          'agendaitems.mandatees.person',
-          'agendaitems.treatment.decision-activity',
-          'document-container.type',
-        ].join(','),
+      const model = await this.store.query('piece', {
+        include: ['agendaitems.treatment.decision-activity'].join(','),
         sort: params.sort,
         'filter[:id:]': result.data.map((record) => record.id).join(','),
```
ValenberghsSven, line 28: Most if the includes are not needed since we want to load the model fast and have each row load it's own data after.
The only needed included is that for the sorting (which will include `agendaitems` and `agendaitems.treatment` anyway

app/routes/publications/overview/shortlist.js
```diff
@@ -22,20 +22,15 @@ export default class PublicationsOverviewShortlistRoute extends Route {
...
       });
+      // the above works but 5 id's can turn into 7 piece records with duplicates
+      // if agendaitem position was changed between agendas
+      return new Set(model.slice());
```
ValenberghsSven, line 32: Locally I tested a use case where an agendaitem was moved from position 3 to 4 on the second version of the agenda.
The result was that 5 ID's turned into 7 records (because of the sort on `agendaitem.number`).
Unsure how realistic that use case is.

app/templates/publications/overview/shortlist.hbs
```diff
@@ -2,18 +2,18 @@
...
-          class="auk-u-min-width-medium"
+          class="auk-table__col--1"
           @currentSorting={{this.sort}}
           @field="agendaitems.treatment.decision-activity.start-date,agendaitems.number"
-          @onChange={{fn (mut this.sort)}}
+          @onChange={{this.changeSort}}
           @label={{t "decision-date"}}
         />
-        <th></th>
+        <th class="auk-table__col--1 u-hide-on-print"></th>
```
ValenberghsSven, line 16: Added column widths here (is this needed here if it's already on `<td>`?) + also the hide on print was added so hide the column
Sorting is no longer a mutation of the sort tracked but a function call

ValenberghsSven (approved):
I've made some changes, tried to justify them in comments.
The sorting works good, the print looks good (except when there are many mandatees) but that is rather rare.
Links are clickable in the pdf when pdf is created.

tdn (approved):

PR #2118, opened 2024-05-28
KAS-4700 show move subcase button only if there is no parliamentflow

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4700

### DIFF
--- a/app/components/subcases/subcase-header.hbs
+++ b/app/components/subcases/subcase-header.hbs
@@ -27,7 +27,7 @@
                   {{t "delete-subcase"}}
                 </AuButton>
               {{/if}}
-              {{#if @parliamentFlow}}
+              {{#unless @parliamentFlow}}
                 <AuButton
                   data-test-subcase-header-action-move-subcase
                   @skin="link"
@@ -36,7 +36,7 @@
                 >
                   {{t "move-subcase"}}
                 </AuButton>
-              {{/if}}
+              {{/unless}}
               {{#if this.canPropose}}
                 <AuButton
                   data-test-subcase-header-action-show-proposed-agendas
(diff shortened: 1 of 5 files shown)


app/components/subcases/subcase-header.hbs
```diff
@@ -8,44 +8,48 @@
...
+                  data-test-subcase-header-action-delete-subcase
+                  @skin="link"
+                  @alert={{true}}
+                  {{on "click" (fn this.requestDeleteSubcase @subcase)}}
+                  role="menuitem"
+                >
+                  {{t "delete-subcase"}}
+                </AuButton>
+              {{/if}}
+              {{#if @parliamentFlow}}
```
sergiofenoll: Might be misunderstanding ticket, shouldn't this be `{{#if (not @parliamentFlow)}}` → only show "move subcase" action if (sub)case is not linked to a parliament flow?

kobemertens: Whoops that was an `unless` I forgot to put back

kobemertens: fixed in [d033020](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2118/commits/d033020fcd42e882bc29df4b9fe8f989e2595cff)

app/components/subcases/subcase-header.js
```diff
@@ -84,6 +84,10 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
     this.subcaseToDelete = subcase;
   }
 
+  get hasActions() {
+    return this.canDelete || this.canPropose || this.args.parliamentFlow;
```
sergiofenoll: @SchrodingersCat00 here as well, right? Should be `!this.args.parliamentFlow` I think.

kobemertens: You're right, not sure how I missed that

ValenberghsSven (approved):

PR #2122, opened 2024-05-31
KAS-4598: now recalculates other agendaitem numbers if the type changes

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4598?focusedCommentId=18823


tdn (approved):
pair review

PR #2123, opened 2024-06-03
KAS-4674: hotfix show documentcontainer position if it exists

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4674


ValenberghsSven (approved):
This looks ok.
There are some weird side effects on data that has no position and you add a document.
But that can be solved in bulk edit
Wait for jenkins though

ValenberghsSven (approved):
small changes, batch row needed extra table column

PR #2124, opened 2024-06-03
KAS-4691: spatie voor document version suffix

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4691


app/utils/vr-document-name.js
```diff
@@ -89,7 +89,7 @@ export default class VRDocumentName {
   }
 
   withOtherVersionSuffix(pieceNr) {
-    return `${this.withoutVersionSuffix}${CONFIG.latinAdverbialNumberals[pieceNr].toUpperCase()}`;
+    return `${this.withoutVersionSuffix} ${CONFIG.latinAdverbialNumberals[pieceNr].toUpperCase()}`;
```
tdn: What if there is no version suffix? Will all piece names get a space behind the name?

ValenberghsSven: @tdn we only use this method when adding a new piece to an existing piece. There should always be a next version suffix.

ValenberghsSven (changes requested):
Looked at this with Tom on monday.
We have multiple pieces of code where we construct a VR number with a version suffix.
I think we need to do more changes in more places.
The ticket only mentions having the space in there for new versions.
In that case this looks fine.
but looking further it is not.

What we should do is trim the first part (if present)
Adding a TER to a BIS that already has a space results in double space (then triple, quadruple) etc.
image

Also, document-naming service does not work properly with it, we have double suffix now (which might predate this change tbh, unsure what is going on there)
image

There may be other places we need to change things. We need to test it more and change where needed

ValenberghsSven:
> "Also, document-naming service does not work properly with it, we have double suffix now (which might predate this change tbh, unsure what is going on there)"

That was fixed in another branch

tdn (approved):

PR #2125, opened 2024-06-03
KAS-4675: add document bulk edit in subcase view

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4675

### DIFF
--- a/app/components/subcase/detail/regular.hbs
+++ b/app/components/subcase/detail/regular.hbs
@@ -45,6 +45,22 @@
                 </div>
               </Auk::Toolbar::Item>
             </Auk::Toolbar::Group>
+            <Auk::Toolbar::Group @position="right">
+              {{#if (and @documentsAreVisible @pieces.length)}}
+                <Auk::Toolbar::Item>
+                  {{#if (user-may "manage-documents")}}
+                    <AuButton
+                      data-test-route-agenda-documents-batch-edit
+                      @skin="secondary"
+                      @icon="pencil"
+                      {{on "click" this.openBatchDetails}}
+                    >
+                      {{t "batch-edit"}}
+                    </AuButton>
+                  {{/if}}
+                </Auk::Toolbar::Item>
+              {{/if}}
+            </Auk::Toolbar::Group>
           </Auk::Toolbar>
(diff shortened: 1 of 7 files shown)


app/components/subcase/detail/regular.hbs
```diff
@@ -45,6 +45,22 @@
                 </div>
               </Auk::Toolbar::Item>
             </Auk::Toolbar::Group>
+            <Auk::Toolbar::Group @position="right">
+              {{#if (and @documentsAreVisible @pieces.length)}}
+                <Auk::Toolbar::Item>
+                  {{#if (user-may "manage-documents")}}
+                    <AuButton
+                      data-test-route-agenda-documents-batch-edit
```
ValenberghsSven: no copied test selectors

app/components/subcase/detail/regular.hbs
```diff
@@ -45,6 +45,22 @@
                 </div>
               </Auk::Toolbar::Item>
             </Auk::Toolbar::Group>
+            <Auk::Toolbar::Group @position="right">
+              {{#if (and @documentsAreVisible @pieces.length)}}
```
ValenberghsSven: `documentsAreVisible` is not really applicable here since we check `(user-may "manage-documents")` below anyway.
just checking if `@pieces.length` should be fine here

ValenberghsSven (changes requested):
Small changes only.

Side note:
I do have 1 concern and that is the delete option (just thinking about that now)
These documents are a collection of all documents of all submission-activities of this subcase.
with postponing / reproposing it is possible to have documents on subcase only + documents on approved agendaitems.
When these documents are on an approved agenda, we might want to prevent deletes from this screen.

We would need complex custom logic to ensure agendaitem versions are updated (which we had but was removed with the rework of this screen.
But i'm not in favor of bringing it back. deleting from this bulk edit screen should just not be possible maybe.

If they want to remove a document, they can do it via documentcard or document-viewer.

@tdn should we prevent delete here in certain scenarios or trust users will behave?

ValenberghsSven (approved):
Added some side stuff unrelated to PR

PR #2126, opened 2024-06-03
check if any agenda was approved

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?assignee=63c552d46178fcc941d857c9&selectedIssue=KAS-4696

### DIFF
--- a/app/components/agenda/agenda-header/agenda-version-actions.js
+++ b/app/components/agenda/agenda-header/agenda-version-actions.js
@@ -37,6 +37,7 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
   @service router;
   @service intl;
   @service toaster;
+  @service jobMonitor;
 
   @tracked showLoadingOverlay = false;
   @tracked loadingMessage = false;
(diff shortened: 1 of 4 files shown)


app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -333,6 +333,16 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
       this.showNotAllowedToast();
       return;
     }
+    const hasJobAfterOpenModal = !!(await this.store.queryOne('job', {
+      'filter[status]': CONSTANTS.JOB_STATUSSES.SUCCESS,
+      'filter[:gte:time-ended]': this.openAgendaCheckTimestamp.toISOString(),
+    }));
```
ValenberghsSven: Job is a very widely used type. any job from any service will be queried here.
We need to specify `DocumentNamingJob`

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -607,4 +617,12 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
...
     );
   }
+
+  @action
+  showAgendaHasChangedToast() {
+    this.toaster.error(
+      this.intl.t('an-agenda-was-approved-since-modal-was-opened'),
+      this.intl.t('warning-title')
+    )
+  }
```
ValenberghsSven: agenda approve / agenda approve and close already have a try catch with an error toast. We could use those

ValenberghsSven (changes requested):
Also for close and approve agenda.
Perhaps put the method in the `documentService` just to keep the document naming job logic more or less in 1 place.
If you include the method calling inside the try catch, you could throw a custom error from it and we can reuse the toaster with a custom message here.

ValenberghsSven (approved):
Reviewed and changed with @tdn 
needed way more then I first thought

PR #2128, opened 2024-06-06
fix document regex to not strip when no type was found, support both ways

ValenberghsSven:
no ticket yet.
Document name stripping on upload is too aggressive

changed the query to be less aggressive and support both "subject-type-index" and "subject-index-type" regardless of how many dashes subject has


ValenberghsSven:
Tested with Tom

PR #2137, opened 2024-06-17
KAS-4608: Reorder on change type

sergiofenoll:
Fix for flag on: https://kanselarij.atlassian.net/browse/KAS-4608

Related PRs:

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/479
- [x] https://github.com/kanselarij-vlaanderen/agenda-submission-service/pull/6

### DIFF
--- a/app/components/subcase/description-panel/edit.js
+++ b/app/components/subcase/description-panel/edit.js
@@ -6,6 +6,7 @@ import { task } from 'ember-concurrency';
 import { action } from '@ember/object';
 import { trimText } from 'frontend-kaleidos/utils/trim-util';
 import addLeadingZeros from 'frontend-kaleidos/utils/add-leading-zeros';
+import { reorderAgendaitemsOnAgenda } from 'frontend-kaleidos/utils/agendaitem-utils';
 
 export default class SubcaseDescriptionEdit extends Component {
   /**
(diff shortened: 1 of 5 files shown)


app/components/subcase/description-panel/edit.js
```diff
@@ -151,32 +151,20 @@ export default class SubcaseDescriptionEdit extends Component {
...
       await this.updateNewsletterAfterRemarkChange();
       await this.updateDecisionReport(propertiesToSetOnAgendaitem.number);
-      await this.recalculateAllAgendaitemNumbersOnAgenda()
+      if (this.agendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA) {
+        const agenda = await this.store.queryOne('agenda', {
+          'filter[agendaitems][agenda-activity][subcase][:id:]': this.args.subcase.id,
+          'filter[:has-no:next-version]': 't',
+        });
+        await this.agendaService.reorderAgenda(agenda);
+      }
```
ValenberghsSven: `reorderAgendaitemsOnAgenda` also checks if any decision reports need to be regenerated, only the agendaitems that were reordered need it. That is missing with this change.

ValenberghsSven (changes requested):
If nota > announcement:
- Reorder current notas (there may be a gap)
- Make agendaitem the last announcement
- Rename and Recreate reports where needed (notas only really, announcement should stay the same)

if announcement > nota:
- Make nota as last
- Call service to reorder according to mandatee logic
- Reorder current announcements (there may be a gap)
- Rename and Recreate reports where needed (notas and announcements)

`reorderAgendaitemsOnAgenda` was doing most of the above. (all except reordering notas to mandatee logic)
With the current code, only part of the above will happen

ValenberghsSven (approved):
used old logic with change to service call instead of manual numbering in 1 scenario out of 4

Added a blocking loader so users can't click away

PR #2139, opened 2024-06-17
KAS-4674: hotfix show documentContainer position in agenda overview

StuyckRuben:
https://kanselarij.atlassian.net/browse/KAS-4674

### DIFF
--- a/app/components/documents/document-badge.hbs
+++ b/app/components/documents/document-badge.hbs
@@ -3,9 +3,11 @@
     href={{@doc.viewDocumentURL}} rel="noreferrer noopener"
     data-test-document-badge-link
   >
-    {{#if @doc.documentContainer}}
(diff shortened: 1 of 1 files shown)


app/components/documents/document-badge.hbs
```diff
@@ -3,6 +3,9 @@
     href={{@doc.viewDocumentURL}} rel="noreferrer noopener"
     data-test-document-badge-link
   >
+    {{#if @doc.documentContainer}}
+      {{await @doc.documentContainer.position}}.&nbsp;
+    {{/if}}
```
ValenberghsSven: containers without position will still show the dot.
image

ValenberghsSven (commented):
was not ok for older agendaitems

Quick fix is to await it using a `#let`, will do so we can merge

ValenberghsSven (approved):
ok now

PR #2140, opened 2024-06-17
check if piece can be saved before making any changes

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?assignee=63c552d46178fcc941d857c9&selectedIssue=KAS-4667

### DIFF
--- a/app/components/documents/document-details-panel.js
+++ b/app/components/documents/document-details-panel.js
@@ -97,6 +97,14 @@ export default class DocumentsDocumentDetailsPanel extends Component {
 
   @task
   *saveDetails() {
+    // This is to avoid concurrency issues where the piece
+    // might have changed already
+    try {
+      const oldPiece = yield this.args.piece;
+      yield oldPiece.save();
+    } catch (e) {
+      return;
+    }
     const signMarkingActivity = yield this.args.piece.belongsTo('signMarkingActivity').reload();
     if (signMarkingActivity) {
       const signSubcase = yield signMarkingActivity?.signSubcase;
--- a/app/models/piece.js
+++ b/app/models/piece.js
@@ -1,5 +1,6 @@
-import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
+import { attr, belongsTo, hasMany } from '@ember-data/model';
 import sanitize from 'sanitize-filename';
+import ModelWithModifier from './model-with-modifier';
 
 /**
  * !when adding a new inverse relation (piece belongsTo/hasMany someModel), make sure to add the polymorphic type {as: 'piece'}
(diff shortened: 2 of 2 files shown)


app/models/piece.js
```diff
@@ -1,13 +1,14 @@
...
 import sanitize from 'sanitize-filename';
+import ModelWithModifier from './model-with-modifier';
 
 /**
  * !when adding a new inverse relation (piece belongsTo/hasMany someModel), make sure to add the polymorphic type {as: 'piece'}
  * !When setting the relation here or in the other model (someModel belongsTo/hasMany piece), make sure to add {polymorphic: true}
  * Referencing a relation between piece and piece requires both
  */
 
-export default class Piece extends Model {
```
ValenberghsSven, line 10: small change, large impact.

app/components/documents/document-details-panel.js
```diff
@@ -97,6 +97,14 @@ export default class DocumentsDocumentDetailsPanel extends Component {
...
   @task
   *saveDetails() {
+    // This is to avoid concurrency issues where the piece
+    // might have changed already
+    try {
+      const oldPiece = yield this.args.piece;
+      yield oldPiece.save();
+    } catch (e) {
+      return;
+    }
```
ValenberghsSven, line 107: why save here? I'm guessing to test concurrency?
if you are going to use `ModelWithModifier` you can just use `preEditOrSaveCheck()` to check if anyone else made changes to piece before you start the process below.

ValenberghsSven (changes requested):
`ModelWithModifier` has it's own save override. so does piece model. How does that work?
`piece` does not have a `modified-by` property in resource but is needed for `ModelWithModifier`. Is there a PR missing for app-kaleidos?

Does this work locally? many failing tests because POST calls of /pieces fails.
 
Implementing `ModelWithModifier` on any model will require more changes and consideration.
For piece, I am reluctant to just merge it because so many processes (like services) are changing pieces but not updating the `modified` or `modified-by` so this will not work unless we also update our services to do that.

I need to look at the actual issue that is going on here.
Ember's aggressive save is saving old "file" relations when backend already changed it.
We only want to save file relation on creation, or on manual replace.
In all other cases, we shouldn't try to save the file.

For now, we can try to fix the issues so the `ModelWithModifier` works, but I might look into alternatives.

For older concurrency issues on agendaitem and subcase I had to resort to reloading possible stale relations before the save. That might be a better solution unless I can detect when the relation was actually updated in the frontend and only save it then.

ValenberghsSven:
Wont merge this, solution has too many side effects on all pieces

PR #2152, opened 2024-07-03
feature/KAS 4722 confidence level at upload

kobemertens:
- **do not allow to clear on access level pill**
- **add access level to uploaded-document component**


tdn (approved):
Looks good, works well.
I also added the feature for cabinet submissions, as it will be very important to have there.

PR #2155, opened 2024-07-05
set default subcase type

kobemertens:
https://kanselarij.atlassian.net/jira/software/c/projects/KAS/boards/1?selectedIssue=KAS-4692


ValenberghsSven:
@tdn should this also apply to new submissions? 
New case > "principiele"
existing case > "definitieve"
maybe?
or always "definitieve" ?

ValenberghsSven:
> @tdn should this also apply to new submissions? New case > "principiele" existing case > "definitieve" maybe? or always "definitieve" ?

We already do the second part on submissions, on existing case we select "definitive".
Guessing on new submissions we force them to choose (non optional, none selected by default) 
so you can ignore my question

ValenberghsSven (approved):

PR #2159, opened 2024-07-12
KAS-4752 change usage of subcaseName for decision tekst, only for notas

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4752


ValenberghsSven:
cherry picked (to be sure not to get any DEV code)

PR #2162, opened 2024-07-18
feature/KAS 4690 cabinet notifications

kobemertens:
- **KAS-4755 fix pagination and empty toolbar item**
- **add notification email on cabinet submission**
- **add cabinet emails to mail settings screen**
- **make sure case title is filled in**
- **send mails on resubmission**


app/components/cases/new-submission.js
```diff
@@ -201,6 +235,21 @@ export default class CasesNewSubmissionComponent extends Component {
...
   });
 
+  createNotificationMailResources() {
+    if (!this.emailSettings) {
+      this.toaster.warning(
+        this.intl.t('notification-mails-could-not-be-sent'),
+        this.intl.t('warning-title')
+      );
+      return;
+    }
```
tdn, line 285: Nice to have: maybe we want to check this on page load and disable submissions if null

app/controllers/cases/case/subcases/subcase/new-submission.js
```diff
@@ -209,7 +214,7 @@ export default class CasesCaseSubcasesSubcaseNewSubmissionController extends Con
       }
     );
     await submissionStatusChange.save();
-
+    this.cabinetMail.sendUpdateSubmissionMails(this.submission);
```
tdn: No comments here?

app/services/cabinet-mail.js
```diff
@@ -0,0 +1,248 @@
...
+      this.store.findRecordByUri('mail-folder', PUBLICATION_EMAIL.OUTBOX),
+    ]);
+
+    if (!(mailSettings && outbox)) {
+      this.toaster.warning(
+        this.intl.t('notification-mails-could-not-be-sent'),
+        this.intl.t('warning-title')
+      );
+      return;
+    }
```
tdn: Put this check in a separate `loadSettings` function?

app/services/cabinet-mail.js
```diff
@@ -0,0 +1,248 @@
...
+
+    const hostUrlPrefix = `${window.location.protocol}//${window.location.host}`;
+    const submissionUrl = this.router.urlFor(
+      'cases.submissions.submission',
+      submission
+    );
```
tdn, line 41: clever! But I would put it in a reusable function/util

app/services/cabinet-mail.js
```diff
@@ -0,0 +1,248 @@
...
+
+    const notificationEmail = caseResubmittedEmail(params);
+    const submitterEmail = caseResubmittedSubmitterEmail(params);
+
+    const notificationEmailResources = [
+      ...submission.approvedBy,
+      ...submission.notified,
+    ].map((address) =>
+      this.store.createRecord('email', {
+        to: address,
```
tdn: We're sending a separate mail to each address here, which is not what is intended. Possibly it should be 1 email with all addresses in 'to:'. That way they can "reply all" with comments.

app/services/cabinet-mail.js
```diff
@@ -0,0 +1,248 @@
...
+    const submissionUrl = this.router.urlFor(
+      'cases.submissions.submission',
+      submission
+    );
+
+    const params = {
+      submissionUrl: `${hostUrlPrefix}${submissionUrl}`,
+      submissionTitle: submission.shortTitle,
+      caseName: submission.title,
+      comment,
```
tdn: The comment to the secretarie cannot be in the email, just the comment on the appropriate field. Will check if this is the case

tdn (changes requested):
A few small remarks, but looks good overall. Will fix the comments myself

app/services/cabinet-mail.js
tdn: TBD later during testing

tdn (approved):

PR #2174, opened 2024-08-14
Feature/kas 4764 submission status change user

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4764

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/485

- added user to `status-change-activity`
- removed `creator` and `beingTreatedBy` from submission model
- fixed inverse between submission and `status-change-activity` so updates happen automatically
- made `submission.statusChangeActivities` read-only 
- Added a save on the `submission` to always set `modifed`, `modified-by` and `created`(once)
- implemented a service for reusable action on submissions
- replaced duplicate getting of statusChangeActivities with service call
- replaced submission model action with service call
- replaced duplicate creation of statusChangeActivities with service call
- replaced  `creator` and `beingTreatedBy` with service call
- added a cleanup of all statusChangeActivities when removing submission

note: some of the service calls could be just a util (no extra services needed) but I put them in the service for now.

### DIFF
--- a/app/services/cabinet-mail.js
+++ b/app/services/cabinet-mail.js
@@ -17,6 +17,7 @@ export default class CabinetMailService extends Service {
   @service router;
   @service toaster;
   @service intl;
+  @service draftSubmissionService;
 
   async loadSettings() {
     const [mailSettings, outbox] = await Promise.all([
@@ -55,7 +56,7 @@ export default class CabinetMailService extends Service {
       };
 
       const mail = caseSendBackEmail(params);
-      const creator = await submission.creator;
+      const creator = await this.draftSubmissionService.getCreator(submission);
       const mailResource = this.store.createRecord('email', {
         to: creator.email,
         from: mailSettings.defaultFromEmail,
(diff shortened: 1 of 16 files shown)


app/services/cabinet-mail.js
```diff
@@ -55,7 +56,7 @@ export default class CabinetMailService extends Service {
       };
 
       const mail = caseSendBackEmail(params);
-      const creator = await submission.creator;
+      const creator = await this.draftSubmissionService.getCreator(submission);
       const mailResource = this.store.createRecord('email', {
```
ValenberghsSven, line 60: I noticed that these mail methods are being called without being awaited. If the user quits too soon the mails might not have been created
Did not change it here

app/templates/cases/submissions/submission.hbs
```diff
@@ -6,6 +6,7 @@
 
 <Submissions::OverviewHeader
   @submission={{@model}}
+  @beingTreatedBy={{this.beingTreatedBy}}
```
ValenberghsSven, line 9: only when currently in treatment, not when accepted or sent back

ValenberghsSven: will do later

PR #2179, opened 2024-08-21
KAS-4712: Better DOCX conversion error handling

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4712

When getting the error message from the DOCX conversion service, we need to broaden the check for if the response is JSON (because sometimes, for some reason, the service adds `; charset=utf-8` to the Content-Type header).

Related PR:
- [ ] https://github.com/kanselarij-vlaanderen/docx-conversion-service/pull/6


ValenberghsSven (approved):
this seems ok.
Forced the service to throw an error.
The message does not say for which document it failed. should we add more info? (out of scope?)
In a large batch, maybe only 1 will fail because of filesize or timeouts.

PR #2180, opened 2024-08-21
KAS-4661: Make subcase name editable from agendaitem route

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4661

Copies over the logic that already existed in the subcase route to make the subcase name editable.

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
+++ b/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
@@ -34,59 +34,61 @@
           {{on "input" (pick "target.value" (set @agendaitem "title"))}}
         />
       </AuFormRow>
-      <AuFormRow>
-        <AuLabel>{{t "subcase-name"}}</AuLabel>
-        <div
-          class="auk-o-flex auk-o-flex--vertical-center auk-o-flex-gap--small au-u-maximize-width"
-        >
-          {{#if (not this.isEditingSubcaseName)}}
-            <Utils::ModelSelector
-              class="auk-u-maximize-width"
-              @modelName="shortcut"
-              @searchField="label"
-              @sortField="label"
-              @filter={{this.filter}}
-              @selected={{this.selectedShortcut}}
-              @onChange={{this.selectSubcaseName}}
-              as |model|
-            >
-              {{capitalize model.label}}
-            </Utils::ModelSelector>
+       {{#if @subcase}}
+        <AuFormRow>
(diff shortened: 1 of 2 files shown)


app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
```diff
@@ -16,9 +17,21 @@ export default class AgendaitemCasePanelEdit extends Component {
...
+  @tracked isEditingSubcaseName = false;
+  @tracked selectedShortcut;
+  @tracked subcaseName;
   confidentialChanged = false;
   propertiesToSet = Object.freeze(['title', 'shortTitle', 'comment']);
 
+  constructor() {
+    super(...arguments);
+    this.subcaseName = this.args.subcase.subcaseName;
+    this.isEditingSubcaseName = this.subcaseName?.length;
```
ValenberghsSven: `this.args.subcase?`. optional for approval items or some legacy agendaitems

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
```diff
@@ -34,6 +34,59 @@
           {{on "input" (pick "target.value" (set @agendaitem "title"))}}
         />
       </AuFormRow>
+      <AuFormRow>
```
ValenberghsSven: entire FormRow `{{#if @subcase}}`

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
```diff
@@ -16,9 +17,21 @@ export default class AgendaitemCasePanelEdit extends Component {
   @service pieceAccessLevelService;
   @service agendaitemAndSubcasePropertiesSync;
 
+  @tracked filter = Object.freeze({
+    type: 'subcase-name',
+  });
+  @tracked isEditingSubcaseName = false;
+  @tracked selectedShortcut;
+  @tracked subcaseName;
```
ValenberghsSven, line 25: out of scope, wondering if we could pour this into a separate component at some stage due to the repetitiveness and the amount of lines duplicated.

ValenberghsSven (approved):
Subcase doesn't always exist so needs to be optional.
Fixed what I found

ValenberghsSven:
Assumed a cypress test would have failed because of it. (locally edit window wouldn't open)
Apparently we don't edit titles of approval items in our tests. I need add that somewhere...

PR #2184, opened 2024-08-23
Feature/kas 4690 new submission notifications

tdn:
See https://kanselarij.atlassian.net/browse/KAS-4690?focusedCommentId=19012 for the requested template for the emails.

One known issue with this PR that would need to be solved, perhaps in a separate ticket:  see `loadAgenda` in `SubmissionHeaderComponent`:  https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2184/files#diff-ac71f19bd3e036fec04c6915614a649504a625e1494d029f24613b41553eb423R47-R75

We should somehow be able to get the intended meeting here more easily and accurately.
However, we don't store the meeting kind anywhere, so this could be any of the meetings on the submission.plannedStart date.
Most likely this will mean we have to store the meeting kind in the submission data model, or even better, the actual meeting uuid.

Tested locally for new submission, send back, resubmit, and BIS update submission. The true test will be on acceptance.

### DIFF
--- a/app/components/submission/header.js
+++ b/app/components/submission/header.js
@@ -37,43 +37,35 @@ export default class SubmissionHeaderComponent extends Component {
   }
 
   loadAgenda = task(async () => {
-    const meeting = await this.args.submission.meeting;
-    if (meeting?.id) {
-      this.selectedMeeting = meeting;
-      this.selectedAgenda = await this.store.queryOne('agenda', {
-        'filter[created-for][:id:]': meeting.id,
-        'filter[:has-no:next-version]': true,
-      });
-    } else {
-      const proposableAgendas = (await this.agendaService.getOpenMeetings()).data.map(
-        (meeting) => ({
-          id: meeting.attributes.agendaId,
-          uri: meeting.attributes.agenda,
-          serialnumber: meeting.attributes.serialnumber,
-          createdFor: {
-            id: meeting.id,
-            uri: meeting.attributes.uri,
-            plannedStart: new Date(meeting.attributes.plannedStart),
-            kind: {
-              label: meeting.attributes.type,
-            }
-          },
-        })
-      );
-      if (proposableAgendas?.length) {
-        // TODO: we should somehow be able to get the intended meeting here more easily and accurately.
-        // However, we don't store the meeting kind anywhere,
-        // so this could be any of the meetings on the submission.plannedStart date.
-        // most likely this will mean we have to store the meeting kind in the submission data model,
-        // or even better, the actual meeting uuid.
-        for (const proposableAgenda of proposableAgendas) {
-          if (proposableAgenda.createdFor.plannedStart.getTime() === this.args.submission.plannedStart?.getTime()) {
-            this.selectedAgenda = proposableAgenda;
-            break;
-          }
-        }
+    if (this.args.submission) {
+      const meeting = await this.args.submission.meeting;
+      if (meeting?.id) {
+        this.selectedMeeting = meeting;
+        this.selectedAgenda = await this.store.queryOne('agenda', {
+          'filter[created-for][:id:]': meeting.id,
+          'filter[:has-no:next-version]': true,
+        });
+      } else {
+        // get meeting when not propagated yet
+        const meetingData = await this.agendaService.getMeetingForSubmission(this.args.submission);
+        const agenda = {
+            id: meetingData.data.attributes.agendaId,
+            uri: meetingData.data.attributes.agenda,
+            serialnumber: meetingData.data.attributes.serialnumber,
+            createdFor: {
+              id: meetingData.data.id,
+              uri: meetingData.data.attributes.uri,
+              plannedStart: new Date(meetingData.data.attributes.plannedStart),
+              kind: {
+                label: meetingData.data.attributes.type,
+              }
+            },
+          };
+
+        this.selectedAgenda = agenda;
+        this.selectedMeeting = agenda.createdFor;
       }
-    }
+  }
   });
 
   get items() {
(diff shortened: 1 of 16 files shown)


ValenberghsSven (commented):
A mail is sent to the creator, even if if the creator has no email set
Going to make a few changes, to make the createRecord of an email generic and reusable.
Then we can check on 1 spot if the "to" mail has any content before making an email with no recipients.

app/components/submission/header.js
```diff
@@ -38,10 +39,40 @@ export default class SubmissionHeaderComponent extends Component {
...
+      );
+      if (proposableAgendas?.length) {
+        // TODO: we should somehow be able to get the intended meeting here more easily and accurately.
+        // However, we don't store the meeting kind anywhere,
+        // so this could be any of the meetings on the submission.plannedStart date.
+        // most likely this will mean we have to store the meeting kind in the submission data model,
+        // or even better, the actual meeting uuid.
+        for (const proposableAgenda of proposableAgendas) {
+          if (proposableAgenda.createdFor.plannedStart.getTime() === this.args.submission.plannedStart?.getTime()) {
+            this.selectedAgenda = proposableAgenda;
```
ValenberghsSven: if there are 2 agendas (MR and PVV) on the same day they will also have the same time (10:00).
We might get the agenda of the wrong meeting here?
Also, we can set `this.selectedMeeting` here?

app/components/submission/header.js
```diff
@@ -142,7 +173,7 @@ export default class SubmissionHeaderComponent extends Component {
       CONSTANTS.SUBMISSION_STATUSES.OPNIEUW_INGEDIEND,
       this.comment
     );
-    await this.cabinetMail.sendResubmissionMails(this.args.submission, this.comment);
+    await this.cabinetMail.sendResubmissionMails(this.args.submission, this.comment, this.selectedAgenda?.createdFor);
```
ValenberghsSven: why not just pass `this.selectedMeeting`, you will have to set it in your loop

app/components/submission/header.js
ValenberghsSven: tested locally, added PVV and immediately got the wrong agenda in the header

ValenberghsSven:
- [x] https://github.com/kanselarij-vlaanderen/agenda-submission-service/pull/9
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/488

app/services/cabinet-mail.js
```diff
@@ -47,15 +47,17 @@ export default class CabinetMailService extends Service {
     if (mailSettings && outbox) {
       const hostUrlPrefix = `${window.location.protocol}//${window.location.host}`;
       const submissionUrl = this.getSubmissionUrl(submission);
+      const meeting = await submission.meeting;
```
ValenberghsSven: cache busting and possible no propagated meeting

ValenberghsSven (changes requested):
In absence of Tom I have changed things where needed based on my review.

Pair reviewed the changes with Sergio
Found that `meeting.kind.uri` was used but wasn't always present. Added the `kind.uri` to the response of my new endpoint.
Found that any indent made in the mails with backticks was also shown in the email (the entire body seems shifted by 1 space)
Kept that as-is but made sure also the "aanvullende informatie" was 1 space, not 2.

app/components/submission/header.js
ValenberghsSven: used custom endpoint and always get the correct meeting for submission (unless component is used for a new submission)

app/components/submission/header.js
ValenberghsSven: done

app/components/submission/header.js
```diff
@@ -29,20 +29,43 @@ export default class SubmissionHeaderComponent extends Component {
...
-    }
+    if (this.args.submission) {
+      const meeting = await this.args.submission.meeting;
+      if (meeting?.id) {
+        this.selectedMeeting = meeting;
+        this.selectedAgenda = await this.store.queryOne('agenda', {
+          'filter[created-for][:id:]': meeting.id,
+          'filter[:has-no:next-version]': true,
+        });
+      } else {
```
ValenberghsSven: realized this after PR review.
SelectedAgenda will usually be wrong for cabinet worker?
They should submit for the latest agenda, but the latest agenda (designagenda) is not propagated yet.
So we should get from ember store for kanselarij editors and from custom endpoint for cabinet workers.

With this, if we have a BIS update where meeting is propagated and agenda A. We are going to fill in `defaultAgenda=A` in the `proposableAgendasModal`, while this should be agenda B.

ValenberghsSven: hmm, this is a bit confusing.
We use this header component in 2 places.

1. new-submission from case view (either new submission or BIS submission)
    here we never have a submission yet, so `selectedAgenda is null`, even for BIS update we don't check what the current meeting of the previous submission was, that happens on creation in a different component (mails are sent there aswell)
    -we only pass it as `defaultAgenda` when it is a new submission (not BIS). In that case, we never need a default because the user has yet to choose an agenda. (wshich will always be null because there is no `args.submission` so we never set `selectedAgenda`
    -in case of a BIS, we open a different modal and don't even use the `selectedAgenda`
2. when editing a submission
   -editing an existing submission does need the agenda, because we want to fill in the agenda (latest version) for secretarie to match the choice of meeting that happened on initial creation of the submission. We also need the meeting for the email sent (we create subcase in the header)
   
   So my initial comment is correct, we load the incorrect agenda for cabinet worker, but the agenda itself is never used so it doesn't really matter.

ValenberghsSven: tested it to confirm.
in case 2 (editing existing) 
Secretarie profile sees agenda B as `selectedAgenda` (for use in Proposableagendamodal)
Cabinet profile sees agenda A as `selectedAgenda` (but not used)

ValenberghsSven: Added `this.currentSession.may('create-subcases-from-submissions')` just to be safe.
So only editors will try the store, others will use the custom endpoint.
Good enough for now. Might want to pull apart into 2 components later (maybe with a shared mini component for the scrolling to documents etc. that is what they have in common.

app/services/cabinet-mail.js
ValenberghsSven: Only used in 2 spots where meeting was already known. passed it as args instead. no cache busting needed because only editors can access this and the cache issue was on the cabinet side

app/components/submission/header.js
ValenberghsSven: fixed

PR #2185, opened 2024-08-27
KAS-4785: Rework case selection for creating/editing submissions

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4785

Replaces some button labels and brings over the case selector for editing submissions.

### DIFF
--- a/app/components/cases/submissions/decisionmaking-flow-selector.hbs
+++ b/app/components/cases/submissions/decisionmaking-flow-selector.hbs
@@ -1,47 +1,67 @@
 <AuFormRow ...attributes>
   <AuLabel @required={{true}} @inline={{true}}>{{capitalize (t "case")}}</AuLabel>
-  <PowerSelect
-    @disabled={{@disableEdit}}
-    @search={{perform this.searchDecisionmakingFlows}}
-    @searchMessage={{t "search-on-case-name"}}
-    @searchEnabled={{true}}
-    @noMatchesMessage={{t "no-results-found"}}
-    @loadingMessage={{t "please-be-patient"}}
-    @selected={{this.decisionmakingFlowOrTitle}}
-    @onChange={{this.onSelectDecisionmakingFlow}}
-    class="au-u-1-1"
-    as |decisionmakingFlow|
-  >
-    {{or decisionmakingFlow.case.shortTitle decisionmakingFlow}}
-  </PowerSelect>
-  {{#unless @disableEdit}}
-    <AuButton
-      @skin="link"
-      @icon="add"
-      class="au-u-margin-top-small"
-      {{on "click" (queue (toggle "isAddingNewDecisionmakingFlow" this) this.setFocus)}}
-    >
-      {{t "add-new-case"}}
-    </AuButton>
-  {{/unless}}
+  <div class="au-c-inline-input-button">
+    {{#if @selectedDecisionmakingFlow}}
+      <AuInput
+        value={{@selectedDecisionmakingFlow.case.shortTitle}}
+        @disabled={{true}}
+      />
+      {{#unless @disableEdit}}
+        <AuButton
+          @skin="secondary"
+          @icon="pencil"
+          {{on "click" (queue this.onClearSelectedDecisionmakingFlow this.setFocus)}}
+        >
+          {{t "new-case"}}
+        </AuButton>
+        <AuButton
+          @skin="secondary"
+          @icon="add"
+          {{on "click" (queue (toggle "isUsingExistingDecisionmakingFlow" this) this.setFocus)}}
+        >
+          {{t "use-an-existing-case"}}
+        </AuButton>
+      {{/unless}}
+    {{else}}
+      <AuInput
+        id="newCaseTitleId"
+        value={{@selectedDecisionmakingFlowTitle}}
+        {{on "input" (pick "target.value" this.onChangeDecisionmakingFlowTitle)}}
+      />
+      <AuButton
+        @skin="secondary"
+        @icon="add"
+        {{on "click" (queue (toggle "isUsingExistingDecisionmakingFlow" this) this.setFocus)}}
+      >
+        {{t "use-an-existing-case"}}
+      </AuButton>
+    {{/if}}
+  </div>
 </AuFormRow>
 
 <ConfirmationModal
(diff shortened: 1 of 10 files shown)


app/components/cases/submissions/decisionmaking-flow-selector.js
```diff
@@ -34,38 +34,41 @@ export default class CasesSubmissionsDecisionmakingFlowSelectorComponent extends
...
   onSelectDecisionmakingFlow = (decisionmakingFlow) => {
     this.decisionmakingFlowTitle = null;
     this.selectedDecisionmakingFlow = decisionmakingFlow;
-    this.args.onChangeTitle?.(null);
-    this.args.onChangeSelected?.(decisionmakingFlow);
   }
 
   onChangeDecisionmakingFlowTitle = (title) => {
-    this.decisionmakingFlowTitle = title;
+    this.decisionmakingFlowTitle = title.trim();
```
ValenberghsSven: this trims are now too aggressive, when typing you can't actually add spaces at the end.
f.e. I want to type "nieuwe dossier titel" but it comes out as "nieuwedossiertitel".

app/components/cases/submissions/decisionmaking-flow-selector.hbs
```diff
@@ -1,47 +1,67 @@
...
+      <PowerSelect
+        @search={{perform this.searchDecisionmakingFlows}}
+        @searchMessage={{t "search-on-case-name"}}
+        @searchEnabled={{true}}
+        @noMatchesMessage={{t "no-results-found"}}
+        @loadingMessage={{t "please-be-patient"}}
+        @selected={{this.selectedDecisionmakingFlow}}
+        @onChange={{this.onSelectDecisionmakingFlow}}
+        class="au-u-1-1"
+        as |decisionmakingFlow|
```
ValenberghsSven, line 72: it's possible to confirm without selecting a case.
`@disabled={{not this.selectedDecisionmakingFlow}}` on the ConfirmationModel`

app/components/cases/submissions/decisionmaking-flow-selector.hbs
```diff
@@ -1,47 +1,67 @@
...
-    <AuButton
-      @skin="link"
-      @icon="add"
-      class="au-u-margin-top-small"
-      {{on "click" (queue (toggle "isAddingNewDecisionmakingFlow" this) this.setFocus)}}
-    >
-      {{t "add-new-case"}}
-    </AuButton>
-  {{/unless}}
+  <div class="au-c-inline-input-button">
```
ValenberghsSven: what is this class? can't seem to find it locally or in the kaleidos-css repo

ValenberghsSven: Will be changing this, the buttons are a bit off vertical center. Also the width of the input field is small.

app/components/cases/submissions/decisionmaking-flow-selector.hbs
```diff
@@ -1,47 +1,67 @@
...
+          @skin="secondary"
+          @icon="pencil"
+          {{on "click" (queue this.onClearSelectedDecisionmakingFlow this.setFocus)}}
+        >
+          {{t "new-case"}}
+        </AuButton>
+        <AuButton
+          @skin="secondary"
+          @icon="add"
+          {{on "click" (queue (toggle "isUsingExistingDecisionmakingFlow" this) this.setFocus)}}
```
ValenberghsSven: `this.setFocus` does nothing with the current changes (when opening the confirmationModal).
It used to point to the input field inside the confirmationModal, but now it points to the new case input field while you are opening a modal with a powerselect. So setting focus to a field behind a modal is not needed.

I added the focus to reduce clicks. I don't think we can open the trigger of the powerselect so you can start typing right away. So only focusing when you press "nieuw dossier" seems ok. you can start typing right away.

app/components/cases/submissions/decisionmaking-flow-selector.hbs
```diff
@@ -1,47 +1,67 @@
...
+    {{else}}
+      <AuInput
+        id="newCaseTitleId"
+        value={{@selectedDecisionmakingFlowTitle}}
+        {{on "input" (pick "target.value" this.onChangeDecisionmakingFlowTitle)}}
+      />
+      <AuButton
+        @skin="secondary"
+        @icon="add"
+        {{on "click" (queue (toggle "isUsingExistingDecisionmakingFlow" this) this.setFocus)}}
```
ValenberghsSven: same, `this.setFocus` does nothing useful here.

app/components/cases/submissions/decisionmaking-flow-selector.hbs
```diff
@@ -1,47 +1,67 @@
 <AuFormRow ...attributes>
   <AuLabel @required={{true}} @inline={{true}}>{{capitalize (t "case")}}</AuLabel>
-  <PowerSelect
-    @disabled={{@disableEdit}}
-    @search={{perform this.searchDecisionmakingFlows}}
-    @searchMessage={{t "search-on-case-name"}}
-    @searchEnabled={{true}}
-    @noMatchesMessage={{t "no-results-found"}}
-    @loadingMessage={{t "please-be-patient"}}
-    @selected={{this.decisionmakingFlowOrTitle}}
```
ValenberghsSven, line 10: getter `decisionmakingFlowOrTitle` no longer used

app/components/cases/submissions/decisionmaking-flow-selector.js
ValenberghsSven: We will have to trim these when saving (like we do for short and longtitle)

ValenberghsSven (changes requested):
input field for case is a bit small.
Maybe same size as the short titel and 
image

Dossierbeheerder is able to edit cases on existing submissions, I think we don't want to allow that. No `@disableEdit` is passed as args in `submission/description-panel/edit.hbs`

it's also possible to change a BIS submission to a different case. This shouldn't be possible I think. Makes for a weird flow. you still have the "aanvaard update en agendeer" option, event though you have selected a new case. (this was before updating with ACC branch. edit panel should no longer be editable when it is an update)


Note on this: I've changed most of what I have found in review to speed up the process.

ValenberghsSven (approved):
pair reviewed with Sergio

PR #2199, opened 2024-09-09
KAS-4795: Various fixes for Kabinet indieningen

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4795

Fixes for multiple bugs/remarks by users about the kabinet indieningen.

- Kabinetdossierbeheerder can't add documents to a submission unless submission "head" minister is part of their cabinet
- Show correct pop-up when deleting a BIS upload in a submission and allow for users to add new versions to existing submission (without having to create a new submission)
- Only show "Dien nieuwe documenten in" if the previous submission has been propagated
- Correctly set the confidentiality of documents on submission

### DIFF
--- a/app/components/subcases/subcase-header.js
+++ b/app/components/subcases/subcase-header.js
@@ -30,6 +30,7 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
   @tracked subcaseToDelete = null;
   @tracked canPropose = false;
   @tracked canDelete = false;
+  @tracked canSubmitNewDocuments = false;
   @tracked meetingIsClosed = false;
   @tracked currentSubmission;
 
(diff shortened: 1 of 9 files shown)


app/components/subcases/subcase-header.js
```diff
@@ -75,6 +77,18 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
...
+      for (const submission of submissions) {
+        const pieces = yield submission.pieces;
+        for (const piece of pieces) {
+          const actualPiece = yield piece.acceptedPiece;
+          if (!actualPiece) {
+            draftPieceWithoutAccepted = true;
+          }
+        }
+      }
+      this.canSubmitNewDocuments = !draftPieceWithoutAccepted;
```
tdn, line 92: One small remark here: since the `currentSubmission` is retrieved using `getOngoingSubmissionForSubcase`, in the case where `this.canSubmitNewDocuments` is `false` because the documents have not propagated yet, the button `Huidige indiening` will not be shown, as it has already been treated. 

So there is a small window of time where the kabinetsdossierbeheerder cannot see the status of their submitted files.

tdn: Fixed this in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2199/commits/448e8851a4dd6c4bbe22d8af752643168c490e6d

tdn (approved):
Tested locally

ValenberghsSven (approved):

PR #2200, opened 2024-09-09
Cypress/kas 4775 cabinet submissions tests

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4775

New tests, need to get them working on jenkins
This will be ticket 1 of many. too much to test in 5 story points.

Run in isolation:
image


app/components/cases/submissions/notifications.js
```diff
@@ -13,16 +14,27 @@ export default class CasesSubmissionsNotificationsComponent extends Component {
   @tracked showNotificationAddressModal = false;
   @tracked emailSettings;
   @tracked isEditing = false;
-
+  @tracked approvalAddresses = new TrackedArray([...this.args.approvalAddresses]);
+  @tracked approvalComment;
+  @tracked notificationAddresses = new TrackedArray([...this.args.notificationAddresses]);
+  @tracked notificationComment;
```
ValenberghsSven: these could/should be empty `TrackedArray` or just have no value on construction. I copied these to the `init` method later.

app/components/cases/submissions/notifications.js
```diff
@@ -37,12 +49,13 @@ export default class CasesSubmissionsNotificationsComponent extends Component {
     }
   });
 
-  updateDefaultAddresses = task(async () => {
-    await this.addApprovalAddress(this.defaultApprovalAddress, true);
-    await this.removeNotificationAddress(this.unusedDefaultNotificationAddress);
+  updateDefaultAddresses = task({ maxConcurrency: 1, enqueue: true }, async () => {
```
ValenberghsSven, line 52: I attempted to prevent `did-update` in template from running a second `updateDefaultAddresses` during construction when confidential is pressed (with slow network)
However that did not seem to work. A second task was always created regardless of these values.

app/components/cases/submissions/notifications.js
```diff
@@ -90,68 +104,76 @@ export default class CasesSubmissionsNotificationsComponent extends Component {
...
+    if (!this.approvalAddresses.includes(address)) {
       if (insertAsFirst === true) {
-        newNotificationData.approvalAddresses.splice(0, 0, address)
+        this.approvalAddresses.splice(0, 0, address)
       } else {
-        newNotificationData.approvalAddresses.push(address);
+        this.approvalAddresses.push(address);
+      }
+      if (!this.updateDefaultAddresses.isRunning) {
+        this.args.onNotificationDataChanged?.(this.notificationData);
```
ValenberghsSven: explanation regarding this:
When calculating the default we were calling the parent multiple times to set our new data.
But the getter `notificationData` was dependent on the data from args.
So we were updating the `notificationData` we are using to calculate the `notificationData`.
on Init, only call the parent once after everything is calculated, then we can update individual parts.

app/controllers/cases/submissions/submission.js
```diff
@@ -65,13 +66,30 @@ export default class CasesSubmissionsSubmissionController extends Controller {
...
     this.notificationComment = newNotificationData.notificationComment;
   };
 
+  onCancelNotificationData = async () => {
+    this.approvalAddresses = this.model.approvalAddresses;
+    this.approvalComment = this.model.approvalComment;
+    this.notificationAddresses = this.model.notificationAddresses;
+    this.notificationComment = this.model.notificationComment;
+  }
+
```
ValenberghsSven: model was not changed, but we should rollback the tracked values or we won't see our old data on cancel.
Need to verify if this is ok in all situations. Should only be applicable when using the panel view/edit toggle mode.

app/controllers/cases/submissions/submission.js
```diff
@@ -181,7 +199,7 @@ export default class CasesSubmissionsSubmissionController extends Controller {
     );
     const defaultAccessLevel = await this.store.findRecordByUri(
       'concept',
-      (this.confidential || parsed.confidential)
+      (this.model.confidential || parsed.confidential)
```
ValenberghsSven: This may have been fixed in other PR but wanted to rule out any side effects

tdn (approved):
pair reviewed and demo during sprint planning

ValenberghsSven:
make a ticket for the notification changes

ValenberghsSven:
Pair reviewed with Tom

PR #2202, opened 2024-09-12
add stub to SH call in login to prevent errors in DB

ValenberghsSven:
Digital-signing service on jenkins does not have connection to the actual API so errors are thrown when trying to setup a session.
Mainly because we pass a fake url. This results in errors in DB

image

Checking if this stubbing is enough.


ValenberghsSven:
DB logs no longer show errors of `digital-signing` unable to execute queries.
Success. Merging

PR #2203, opened 2024-09-12
Feature/kas 4799 submission changes

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4799


tdn (approved):
Pair reviewed, works well.

Note for future reference: we might want a quick-access link in the subcase-header to the history panel. However, we will confer with design about this.

PR #2204, opened 2024-09-13
KAS-4809 added link to submit new documents on agendaitem after user tests

tdn:


app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -41,30 +42,43 @@ export default class AgendaitemControls extends Component {
...
+          this.canSendToVP = false;
+        }
       } else {
         this.canSendToVP = false;
       }
-    } else if (this.currentSession.may('send-cases-to-vp')) {
-      this.canSendToVP = await this.parliamentService.isReadyForVp(this.args.agendaitem);
+      if (isEnabledCabinetSubmissions() && this.currentSession.may('create-submissions')) {
+        if (currentUserOrganizationMandateesUris.includes(submitter?.uri)) {
+          this.canSubmitNewDocuments = await this.parliamentService.isReadyForVp(this.args.agendaitem);
```
ValenberghsSven: should `this.canSubmitNewDocuments = await this.parliamentService.isReadyForVp(this.args.agendaitem);` be `this.canSubmitNewDocuments = true` ?
not sure what VP has to do with it.

app/components/agenda/agendaitem/agendaitem-controls.js
```diff
@@ -41,30 +42,43 @@ export default class AgendaitemControls extends Component {
...
+        } else {
+          this.canSubmitNewDocuments = false;
+        }
+      } else {
+        this.canSubmitNewDocuments = false;
+      }
```
ValenberghsSven: You could start off by setting `this.canSubmitNewDocuments = false` before everything and only setting it to true in that one case.
Saves you making the `else` so explicit and duplicate.

ValenberghsSven: same with `canSendToVP`

ValenberghsSven: will come back to this after cleanup

ValenberghsSven (approved):
pair reviewed

PR #2210, opened 2024-09-19
Cypress/fix tests on dev

ValenberghsSven:
No ticket. Too many flaky tests on DEV/ACC make it difficult to see what is actually failing.
Improving on the tests.


ValenberghsSven:
the failing newsitem spec is so annoying (just sorting really)
We have a list of 3 agendaitems (1 hidden approval) and both notas should show in the zebra view.
For some reason, sorting is causing issues when sorting halfway the spec where 1 item goes missing.
Can't seem to reproduce it locally but it's a constant on jenkins. Usually an issue with 'someone else messed up the data or cache in a previous test' but that shouldn't happen. This view is only used in profile specs and most of those run after this one.
It seems we are getting 3 results on our query in the model (query to all agendaitems of agenda) which is what is expected. but The row is not showing. need to find a good way to log this better. Maybe this is another instance of the `sort` messing up with integers and properties and returning a meta of X but results of Y
I could try a post query sort, to see if that fixes the test.

Might have to test this in isolation on jenkins or move the entire spec to the all-flaky-tests folder to see if running the spec as one of the first helps.

ValenberghsSven:
next are the profile specs. It makes no sense why certain data would not be viewable. I was thinking a session issue where we fail to get the session after logout login. but hard to pinpoint. The failing spec is also not very consistent. different routes, different users. Wonder if we are hitting some sort of choke because we are switching between a lot of things in a short amount of time.
To eliminate that I would have to try adding an arbitrary timeout to each test to slow them down.
But since it is flaky that would mean running the entire suite multiple time to ensure it does anything.

ValenberghsSven:
move subcases is also flaky. In the screenshots you can see the correct call is made to search, but the view is not updating with the results. locally not reproducable.
I thought since it was search the case may not have been indexed yet, but then you should see "no results yet" which we don't.
Maybe an issue with the fact that cypress doesn't slowly type like a user. it is instant (can be slowed down in settings) which may sometimes (but almost consistently fail here). have not tried to test this locally with cypress (I tried manually and typing always updates the view) Wonder if this could be an electron issue (cypress default browser) (can't forget about those..)

ValenberghsSven:
systemalert's dropdown failure have not reproduced yet (dropdown not clickable yet).
Is something loading maybe? did we try to click a disabled option.
Could add a small delay (200 ms) before clicking or checking `should(not.be.disabled).click()`
note: systemalert.cy has commented code. see if we can re-enable (datepicker was broken in this view once, used and agenda to test the datepicker instead. we might want to keep both (since datepicker is implemented differently in some places.)

ValenberghsSven:
update for move subcase: it is possible to start searching before the first search call completed.
It seems that if you typed text during that time we don't search again after so all cases are shown, not your filtered list.

ValenberghsSven:
Most of the tests are fixed. I'm going to merge this branch. Most of the flaky tests are more stable now.
The newsitem spec might be fixed by using 1 login per spec or a change in the way we get data.
Or perhaps the new mu-auth will have fixed this specific issue where sometimes we don't get the resources we should get.
Going to stop spending time on this for now.

PR #2213, opened 2024-09-23
KAS-4822 added read-only support for certain browsers

tdn:
Uses the existing `ember-useragent` service to allow Edge, but then checks a list of read-only permissions in case the browser is Edge or mobile.

At least locally this seems to work at first glance.


app/config/permissions.js
```diff
@@ -122,6 +122,23 @@ const groups = [
       'edit-in-treatment-submissions',
       'create-subcases-from-submissions',
       'delete-submissions',
+    ],
+    readOnlyPermissions: [
+      'view-all-ongoing-signatures',
+      'search-publication-flows',
```
ValenberghsSven: Not a full fan of the duplications.
Could we:
- create 2 lists
- 1 with all reading permissions
- 1 with all writing permissions
- if not mobile or edge > combine the lists

Reason: we could forget to add/remove permissions on both lists.

app/services/current-session.js
```diff
@@ -44,6 +45,11 @@ export default class CurrentSessionService extends Service {
   }
 
   may(permission, checkImpersonator = false) {
+    if (this.userAgent.browser.isMobile || this.userAgent.browser.isEdge) {
```
ValenberghsSven: checked with smartphone.
`this.userAgent.browser.isMobile` does not seem to be a thing on samsung.
Is this phone specific?

My `browser` when using samsung internet
```
{"Info":    {"name":"Samsung Internet","version":"26.0","major":"26"},
   "isChrome":false,"isChromeHeadless":false,"isEdge":false,"isFirefox":false,"isIE":false,"isSafari":false}
```
and edge
```
{"Info":    {"name":"Edge","version":"128.0.0.0","major":"128"},
   "isChrome":false,"isChromeHeadless":false,"isEdge":true,"isFirefox":false,"isIE":false,"isSafari":false}
```

ValenberghsSven: should be `this.userAgent.device.isMobile` you also have `isTablet`  (`isDesktop`) for pc

ValenberghsSven:
At the time of review, jenkins run is going with all tests in edge browser. So far it seems ok (50% done)
So maybe we don't have to restrict desktop edge users tbh.
Unless Edge is doing weird stuff visually, cypress wouldn't pick that up, I don't see why we shouldn't just allow edge at this stage.
Wait for full test results, maybe a manual check of some things (like rdfa-editor)

ValenberghsSven:
Cypress with edge desktop showed nothing special. I see no reason not to "support" edge on desktop; even the rdfa-editor tests are fine.

app/config/permissions.js
```diff
@@ -160,137 +152,183 @@ const groups = [
...
+        'view-decisions-before-release',
+        'view-all-confidential-documents',
+        'view-ratification-before-release',
+        'search-confidential-cases',
+        'search-confidential-documents',
+      ],
+      full: [
+        'manage-news-items',
+        'manage-themis-publications',
+        'view-documents-postponed-and-retracted',
```
ValenberghsSven: deze bij readOnly?

ValenberghsSven: done

PR #2214, opened 2024-09-24
KAS-4614: Better support for moving subcases

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4614

Relinks the parliament flow to the new decisionmaking flow (when applicable) as well as submissions and sign flows.

Some restrictions exist as to when a subcase can be moved:

- It must have no parliament flow
- Or if it has a parliament flow it must be the only subcase
- Or if it has a flow and it's not the only subcase, it must have a parliament retrieval activity (denoting that it's a Bekrachtiging)

- [x] https://github.com/kanselarij-vlaanderen/vlaams-parlement-sync-service/pull/41


app/components/subcases/subcase-header.js
```diff
@@ -57,6 +59,12 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
     if (!this.canSubmitNewDocuments && this.currentSession.may('view-submissions')) {
       this.currentSubmission = yield this.draftSubmissionService.getOngoingSubmissionForSubcase(this.args.subcase);
     }
+    const decisionmakingFlow = yield this.args.subcase.decisionmakingFlow;
+    const _case = yield decisionmakingFlow.case;
+    const parliamentFlow = yield _case.parliamentFlow;
+    const parliamentRetrievalActivity = yield this.args.subcase.parliamentRetrievalActivity;
+    const subcases = yield decisionmakingFlow.hasMany('subcases').reload();
+    this.canMove = !parliamentFlow || (parliamentRetrievalActivity && subcases.length === 1);
```
ValenberghsSven: the `parliamentFlow` is passed as args already `@parliamentFlow={{@model.parliamentFlow}}`.
Calculated first in the cases/case route.
For the length of subcases, the model of the cases/case/subcases/subcase route already has that list. The `SubcaseHeader` component is being used in that template.
So we just need to set it on the controller and use that instead of a relation reload.

 ```
this.canMove = !this.args.parliamentFlow || (parliamentRetrievalActivity && this.args.subcases.length === 1);
```

Edit: `parliamentFlow` is being passed to `<Subcases::SubcaseHeader>` but the model doesn't actually have that property. Must have been a refactor mishap... We need to pass it down properly from the parent routes...

Comment (ValenberghsSven) [reply app/components/subcases/subcase-header.js]: Also, do we want to put this in a getter like `maySubmitNewDocuments`?
including `this.loadData.isIdle` and a permissions for when you are allowed to move subcases? 
Right now, all the actions are behind a `(user-may "manage-agendaitems")` in the template which is fine for now?
TLDR: getter would be nice? permission optional since it is already behind one.

File: app/models/parliament-retrieval-activity.js [hunk 2]
```diff
@@ -1,8 +1,8 @@
 import Model, { attr, hasMany, belongsTo } from '@ember-data/model';
 
 export default class ParliamentRetrievalActivity extends Model {
-  @attr('date') startDate;
-  @attr('date') endDate;
+  @attr('datetime') startDate;
+  @attr('datetime') endDate;
```
ValenberghsSven, line 5: this seems ok compared to our model.
Any further actions needed because we created models with incorrect type?

app/components/subcases/subcase-header.js
```diff
@@ -193,13 +201,63 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
...
+      submission.decisionmakingFlow = this.newDecisionmakingFlow;
+      await submission.save();
+    }
+
+    if (parliamentRetrievalActivity) {
+      const newParliamentFlow = await newCase.parliamentFlow;
+      if (newParliamentFlow) {
+        const oldParliamentSubcase = await parliamentRetrievalActivity.parliamentSubcase;
+        const oldParliamentFlow = await oldParliamentSubcase.parliamentFlow;
+
```
ValenberghsSven, line 225: we only come in here when there is a retrieval.
Then we check if the new case has a parliament flow already.
What about submission activities?
Do we need to update VP for that? (when there is a parliament flow but no retrieval yet)

ValenberghsSven (changes requested):
reuse data from routes where possible. Rest is ok.
Not manually tested yet (was hoping to combine multiple VP tickets for testing)

app/components/subcases/subcase-header.js
ValenberghsSven: nvm, those submissions are not attached to subcases directly

app/components/subcases/subcase-header.js
ValenberghsSven: Doing this locally, will commit later

ValenberghsSven (approved):

PR #2218, opened 2024-09-27
KAS-4713: Betreft automatisch updaten in service

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4713

When approving an agenda, we now always regenerate the decisions of agendaitems that have new document names AND tell the backend to regenerate the "betreft" part so we ensure that the (potentially) new piece names are used.

Note that this does not include BIS documents. For those, the users still need to manually update the decision.

There frontend functionality to generate the betreft part isn't removed, it's now duplicated across the FE and the microservice. In a future we might want to always use the service.

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/decision-report-generation-service/pull/24
- [x] https://github.com/kanselarij-vlaanderen/document-naming-service/pull/6


app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -438,9 +437,10 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
       await this.reloadAgenda(this.args.currentAgenda);
       await this.reloadAgendaitemsOfAgenda(this.args.currentAgenda);
       await this.reloadMeeting();
-      if (agendaitemsNotOk?.length) {
-        await this.decisionReportGeneration.regenerateDecisionReportsForMeeting.perform(this.args.meeting, true);
-      }
+      // Regenerate ALL beslissingsfiches, so we ensure that the concerns part
+      const newNames = true;
+      const shouldRegenerateConcerns = true;
+      await this.decisionReportGeneration.regenerateDecisionReportsForMeeting.perform(this.args.meeting, newNames, shouldRegenerateConcerns);
```
tdn: This seems to work fine, but I'm concerned we'll be regenerating too much for large agendas with a few documents added in the last version. This would not only lead to an unnecessary wait/use of server resources, but also would mean the kanselarij needs to re-check all decisions on every approval, even the ones they didn't touch.

I do get why this was the easier route to take, and we discussed it in the standup and I said it would be fine iirc.

Obviously we can't simply look at the agendaitems  with documents without an `originalName`, since we just named them all.

I'm wondering if we can somehow leverage the DocumentNamingJob to filter it down to only the necessary decisions. I'll look into this for a bit.

tdn: I think I almost have it figured out, will continue tomorrow

tdn: Made a small adjustment to `document-naming-service` for this, and at least locally, it seems to work now

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -342,22 +342,25 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
...
-        this.agendaCheckMapping,
+        agendaCheckMappingsMap,
         this.openAgendaCheckTimestamp,
       );
       await this.documentService.stampDocumentsOfAgenda(this.args.currentAgenda.id);
       // Data reloading
       await this.reloadAgenda(this.args.currentAgenda);
       await this.reloadAgendaitemsOfAgenda(this.args.currentAgenda);
       await this.reloadMeeting();
-      if (agendaitemsNotOk?.length) {
```
ValenberghsSven, line 358: `if (agendaitemsNotOk?.length) {`
this was used to fix gaps in numbering if a new agendaitem was removed during approval.
After first refactor > always regenerate everything, this removal was fine.
After second refactor (only regen those with new documents be name (not BIS)), this is no longer ok

app/components/agenda/agenda-header/agenda-version-actions.js
```diff
@@ -342,22 +342,25 @@ export default class AgendaAgendaHeaderAgendaVersionActions extends Component {
...
-        await this.decisionReportGeneration.regenerateDecisionReportsForMeeting.perform(this.args.meeting, true);
-      }
+      // Regenerate ALL beslissingsfiches, so we ensure that the concerns part
+      const newNames = true;
+      const agendaitemsToRegenerateConcernFor = [...new Set(this.agendaCheckMapping.map(({ agendaitem }) => agendaitem.id))];
+      await this.decisionReportGeneration.regenerateDecisionReportsForMeeting.perform(this.args.meeting, newNames, agendaitemsToRegenerateConcernFor);
```
ValenberghsSven: this.decisionReportGeneration.regenerateDecisionReportsForMeeting only when there are mappings. or we will regen everything when there are no changes

ValenberghsSven (approved):
heavily pair reviewed.

PR #2220, opened 2024-09-30
KAS-4831 fix tests to work with new government and secretaries

ValenberghsSven:
Fix tests on acceptance (could be cherry picked to production when done)


ValenberghsSven:
The only tests that are still failing have been fixed (or in the process of getting fixed) in a different branch.
That one came first, but this was more urgent

tdn (approved):

PR #2223, opened 2024-10-02
KAS-4842 move submissions page

ValenberghsSven:
Also change model to show not on closed agenda, pill label and color change

https://kanselarij.atlassian.net/browse/KAS-4842

Tests will fail because of the mandatees (fixing that in another branch)


app/components/cases/submissions/status-pill.js
```diff
@@ -0,0 +1,46 @@
...
+    super(...arguments);
+    this.loadStatusText.perform();
+  }
+
+  loadStatusText = task(async () => {
+    // await promises
+    const status = await this.args.status;
+    const meeting = await this.args.meeting;
+    // show status already
+    this.label = status.label || status.altLabel;
```
tdn: Probably due to faulty data, but it appears `status` can be `null`. I'll add a check here.

image

app/routes/submissions.js
```diff
@@ -0,0 +1,105 @@
...
+    if (!this.currentSession.may('view-submissions')) {
+      this.router.transitionTo('cases.index');
+    }
+  }
+
+  async model(params) {
+    // *note: the cache busting delays the loading a bit, even locally with only 2 submissions it take half a second
+    const options = {
+      'filter[:has:created]': `date-added-for-cache-busting-${new Date().toISOString()}`,
+      'filter[pieces][accepted-piece][agendaitems][agenda][created-for][:has-no:agenda]': 't', // filter out submissions on closed meetings
```
tdn, line 56: Edge case: if the last piece was removed after being accepted, an "empty" submission might perpetually remain in the list.

Adding a `:has:pieces` filter to cover this.

tdn (approved):

PR #2224, opened 2024-10-02
KAS-4838 added new document-preview for mobile devices

tdn:
This seems to work fine on my laptop while mocking a mobile device, but when I try to open the inlineViewLink on my actual phone, it gives me a 404. This might be a bigger issue than pure frontend...

Update: @ValenberghsSven and I tried this with PDFjs viewer, and turns out that one just works perfectly on mobile.
So we're going with that instead of the redesign hack


app/components/documents/document-view.hbs
```diff
@@ -1,12 +1,16 @@
...
   {{#if this.isPdfDocument}}
-    <object
-      data-test-document-view-file
-      type="application/pdf"
-      data={{@inlineViewLink}}
-      class="au-u-maximize-width au-u-maximize-height auk-document-view"
-    >
-    </object>
+    {{#if @isMobile}}
+      <PdfView @filePath={{@file.namedDownloadLink}} />
```
ValenberghsSven, line 4: this should be the namedDownloadLink of piece. (actually the getter `downloadLink` of the parent)
the "download" button in the viewer itself is now downloading with the file name (no VR number) instead of the piece name (VR number)
Will fix on ACC

PR #2225, opened 2024-10-03
KAS-4814 add internal review model and prefer it over agendaitem prop…

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4814

- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/496


app/components/agenda/agenda-overview/agenda-overview-item.hbs
```diff
@@ -205,7 +205,16 @@
         <div class="vlc-agenda-items__remarks">
           {{!-- Empty container below so that, when it is empty, the adjacent element is still pushed to the right  --}}
           <div>
-            {{#if @agendaitem.privateComment}}
```
ValenberghsSven, line 208: I might add a permission check here for good measure.
But if I use "manage-agendaitems" then OVRB and KB will no longer be able to see this comment.
They can see it now on prod

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-view.hbs
```diff
@@ -153,7 +153,22 @@
       </div>
     </Auk::Panel::Body>
   {{/if}}
-  {{#if @agendaitem.privateComment}}
+  {{#if (and (user-may "manage-agendaitems") @subcase.internalReview)}}
```
ValenberghsSven, line 156: my thought process here (and other places- was that the permission check would prevent calling the relation for other profiles when not needed. Technically we could just call the relation and it will be empty.

app/components/agenda/agenda-overview/agenda-overview-item.hbs
tdn: If there's no check on PROD now, I wouldn't add it. Could lead to an unintended side-effect

app/components/cases/subcases/new-subcase-form.js
```diff
@@ -223,6 +223,8 @@ export default class NewSubcaseForm extends Component {
           this.intl.t('warning-title')
         );
       }
+    } else {
+      yield this.agendaService.createInternalReview(this.subcase, null, privateComment);
```
ValenberghsSven, line 227: was wondering about this. subcase without agenda-activity.
It makes sense to not have the current "agendaitem.privateComment" yet.
But the process of their checklist can already happen with this new model so I already added an internalReview at this point.

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-view.hbs
tdn: Since this is a "new" place for this remark, I'm fine with the permission check here. If there are users that don't have that permission, they're used to check it in the agenda, not here.

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-view.hbs
```diff
@@ -153,7 +153,22 @@
...
+          <span>{{t "private-comment-title"}}:</span>
+        </div>
+        <pre
+          class="au-u-flex-item--fill auk-u-text-pre-line auk-u-text-muted"
+          data-test-agendaitem-titles-private-comment
+        >
+          {{@subcase.internalReview.privateComment}}
+        </pre>
+      </div>
+    </Auk::Panel::Body>
```
ValenberghsSven, line 170: note that I kept the styling for agendaitem to be the same.
The views in subcase and submissions are different ('interne opmerking" is above, not on the side.)
Didn't implement this change here.
Not sure what looks better.

app/components/subcase/description-panel/edit.hbs
```diff
@@ -134,6 +134,21 @@
...
+          </AuLabel>
+          <AuTextarea
+            id="private-subcase-comment"
+            rows="7"
+            value={{this.internalReview.privateComment}}
+            @width="block"
+            {{on "input" (pick "target.value" (set this.internalReview "privateComment"))}}
+          />
+        </AuFormRow>
+      {{/if}}
```
ValenberghsSven, line 151: I forgot to add similar code to the "bekrachtiging description panel"

app/components/submission/header.js
```diff
@@ -373,24 +382,58 @@ export default class SubmissionHeaderComponent extends Component {
...
-      this.comment
-    );
-    await this.cabinetMail.sendBackToSubmitterMail(
-      this.args.submission,
-      this.comment,
-      this.selectedMeeting,
-    );
-    if (isPresent(this.args.onStatusUpdated)) {
-      this.args.onStatusUpdated();
+  getOrCreateInternalReview = async () => {
```
ValenberghsSven: was not a fan of this name. "createOrUpdateInternalReview" might be better

app/components/submission/header.js
```diff
@@ -373,24 +382,58 @@ export default class SubmissionHeaderComponent extends Component {
...
+    //   this.comment
+    // );
+    // await this.cabinetMail.sendBackToSubmitterMail(
+    //   this.args.submission,
+    //   this.comment,
+    //   this.selectedMeeting,
+    // );
+    // if (isPresent(this.args.onStatusUpdated)) {
+    //   this.args.onStatusUpdated();
+    // }
```
ValenberghsSven: TODO revert this
This was an easy way to test the above code...

app/components/submission/header.js
```diff
@@ -373,24 +382,58 @@ export default class SubmissionHeaderComponent extends Component {
...
+  
+    if (!internalReviewOfSubmission?.id) {
+      await this.agendaService.createInternalReview(this.args.subcase, [this.args.submission], CONSTANTS.PRIVATE_COMMENT_TEMPLATE);
+      // const internalReview = await this.store.createRecord('submission-internal-review', {
+      //   created: new Date(),
+      //   privateComment: CONSTANTS.PRIVATE_COMMENT_TEMPLATE,
+      //   submissions: [this.args.submission],
+      //   subcase: this.args.subcase,
+      // });
+      // await internalReview.save();
```
ValenberghsSven: remove, replaced by service call

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
```diff
@@ -105,14 +105,25 @@
...
         </AuLabel>
-        <AuTextarea
-          data-test-agendaitem-titles-edit-private-comment={{true}}
-          id="private-comment-title"
-          rows="7"
-          value={{@agendaitem.privateComment}}
-          @width="block"
-          {{on "input" (pick "target.value" (set @agendaitem "privateComment"))}}
-        />
+        {{#if this.internalReview}}
```
tdn, line 108: My original thought was to check the CABINET_SUBMISSIONS feature flag here as well. 
With the current way 2 systems can potentially be used at the same time...

Of course then the internalReview would need to be created if it doesn't exist... 

This might be a non-issue, as with your case we'll capture all `agendaitems` that were created as `subcases` or `submission` from the moment we deploy this, which is fine.

app/serializers/submission-internal-review.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class SubmissionInternalReviewSerializer extends ApplicationSerializer {
+  shouldSerializeHasMany(snapshot, key, relationshipType) {
+    if (FORCE_SERIALIZED.includes(key)) {
+      return true;
+    } else {
+      return super.shouldSerializeHasMany(snapshot, key, relationshipType);
+    }
+  }
+}
```
ValenberghsSven, line 13: the HasMany relation would not save without this.
I think Sergio did say something about it awhile ago.
Default one-to-many in ember does not save the many side unless forced.
Here I wanted to force it so we could only update `SubmissionInternalReview.submissions` to avoid concurrency/data loss by profiles who do not have the models in data.

app/models/submission-internal-review.js
```diff
@@ -0,0 +1,12 @@
+import { belongsTo, hasMany, attr } from '@ember-data/model';
+import ModelWithModifier from 'frontend-kaleidos/models/model-with-modifier';
+
+export default class SubmissionInternalReviewModel extends ModelWithModifier {
```
ValenberghsSven, line 4: I need to test the concurrency still. that throws an error which will be uncaught and break logic.
Do we really need the `ModelWithModifier`?
We could try without. Worst case they lose a few lines of text? 
Or we need a way where we don't error but just warn?
I'm inclined to take it out because all the points of error need to be surrounded with try catch

ValenberghsSven: we could also reload `internalReview.submissions` before save since losing connection to a BIS submission is likely to happen when editing a design agenda when BIS update is ongoing?

app/routes/cases/submissions/submission.js
```diff
@@ -130,6 +130,9 @@ export default class CasesSubmissionsSubmissionRoute extends Route {
   async afterModel(model) {
     const decisionmakingFlow = await model.belongsTo('decisionmakingFlow').reload();
     await decisionmakingFlow?.case;
+    if (this.currentSession.may('treat-and-accept-submissions')) {
+      await model.internalReview;
+    }
```
ValenberghsSven, line 135: I haven't done this in the agendaitem routes. Might be cleaner for views than to have a silent await in the template

app/services/agenda-service.js
```diff
@@ -136,6 +147,10 @@ export default class AgendaService extends Service {
     formallyStatusUri = CONSTANTS.FORMALLY_OK_STATUSES.NOT_YET_FORMALLY_OK,
     privateComment = null
   ) {  
+    const internalReview = await subcase.internalReview;
+    if (!internalReview?.id) {
+      await this.createInternalReview(subcase, null, privateComment);
+    }
```
ValenberghsSven, line 153: final check before going to submit subcase to a meeting. We want internalReviews for every single subcase yes?
I was thinking of falling back to the CONSTANT here, but if the editors intentionally empty the private comment we should not default to it again.

tdn (changes requested):
Tested:
- new submission
- try to edit remark before taking in treatment as Admin -> OK
- take in treatment, then edit remark -> OK
- Make subcase and put on agenda ->
One remaining Issue:
The Cases::Subcases::ProposableAgendasModal still uses the old property it seems: 

image

- Edit in subcase -> OK
- Edit in agenda -> OK
- Send back to submitter -> OK
- approve agenda -> OK
- Submit update (BIS) -> small remark: the top panel is not editable for updates, so the remark also cannot be edited. 
image

Since it is the same remark, I'd expect them to be able to edit it on the agenda instead, but when I do that, the remark disappears from the update submission:
image

- Accept update and put on agenda -> OK

tdn (approved):
saw the changes locally

app/components/agenda/agenda-overview/agenda-overview-item.hbs
ValenberghsSven: I've checked other profiles. A non-editor profile on this view does not have the model for internalReviews (as expected, yggdrasil needs to be told explicitly what to collect.
OVRB and KB still see the comment as they did before, but only in the agenda views, not subcase view

app/components/submission/header.js
ValenberghsSven: changed it

app/components/submission/header.js
ValenberghsSven: fixed it

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
ValenberghsSven: I will run a test to see what this does with the submissions flag off.
Since I tested adding this model to subcases that have no submissions I expect it to work just fine.

app/models/submission-internal-review.js
ValenberghsSven: I think we are keeping the concurrency check for now.
Can be finetuned later if we were to extract this bit from the description panels and save it separately.
we did have to do a `internalReview.hasMany('submissions').reload();` for concurrency issues

PR #2226, opened 2024-10-09
Feature/kas 4807 submission fixes

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4807

Semi done + flags from ACC


app/components/cases/submissions/status-pill.js
```diff
@@ -23,14 +23,13 @@ export default class CasesSubmissionsStatusPillComponent extends Component {
   loadStatusText = task(async () => {
     // await promises
     const status = await this.args.status;
-    const meeting = await this.args.meeting;
     // show status already
     this.label = status?.label || status?.altLabel;
-
     // check if submission is on a visible agenda
-    if (meeting?.id){
+    if (this.args.submission){
```
ValenberghsSven: `?.id`

app/components/documents/document-details-panel.js
```diff
@@ -188,6 +214,12 @@ export default class DocumentsDocumentDetailsPanel extends Component {
     } else {
       yield this.documentService.checkAndRestamp([this.args.piece]);
     }
+    
+    if (this.internalReview?.hasDirtyAttributes) {
+      yield this.internalReview.belongsTo('subcase').reload();
```
ValenberghsSven, line 219: Reloading subcase may not be needed since draft-pieces cannot be edited in this viewer (yet)
If they could be edited here in the future (I expect the request) it is possible that the submission is accepted while this view is open or being edited.
In that case we have to reload the subcase or we lose the connection to the concurrently created subcase.

app/components/submission/header.js
```diff
@@ -381,6 +391,7 @@ export default class SubmissionHeaderComponent extends Component {
   );
 
   takeInTreatment = async () => {
+    // TODO update submission data? It could have been changed on subcase
```
ValenberghsSven, line 394: out of scope (some of this code came from ACC flags.)
When BIS submission is ongoing, any updates to the subcase do not cascade to the submission.
If we were to update the submission, this might be the place to do it (before editors start treating)
They could then auto update and send back if needed 'something changed on subcase, change ... on your submission'. I don't know

app/components/submission/header.js
```diff
@@ -400,7 +411,7 @@ export default class SubmissionHeaderComponent extends Component {
       const submissions = await internalReviewOfSubcase.hasMany('submissions').reload();
       addObject(submissions, this.args.submission);
       internalReviewOfSubcase.submissions = submissions;
-      return await this.args.submission.save();
+      return await internalReviewOfSubcase.save();
```
ValenberghsSven, line 414: this was wrong, BIS submission internal review was not saved because of the wrong model.
Don't think anyone noticed yet.

ValenberghsSven (commented):
going to merge, looked at most with Tom.

PR #2227, opened 2024-10-10
Made subcase type editable on agenda

tdn:


ValenberghsSven:
Added a minor change. We make it possible to edit subcase-type in 5 places now.
2 of those those places had a change where we don't update `subcaseName` when changing the `type` of `subcase`.
That was a request from the users (because we were silently setting the `subcaseName` and that was unwanted).
Mainly discovered when creating new subcases. Changing types on existing subcase is rare.
Also applying that change here to the other 3 places this still happens.

ValenberghsSven (approved):

PR #2228, opened 2024-10-11
attempt to investigate flaky spec

ValenberghsSven:
Newsitem.cy sorting test has gone from flaky to consistently broke.
It doesn't seem to make sense.
Test succeeds locally, in full spec or .only, with UI or headless.
The call to agendaitems has a meta of 3, but not all 3 are found in the data attributes.
Wondering if this has anything to do with the sorting.
We are sorting on a datetime via `treatment.news-item.modified` so from agenda > treatment > news-item > modified.

In this scenario
2 of the agendaitems have no news-item, all have a treatment.
In the call for data with meta 3
On 1 should be the agendaitem without newsitem   <--- this one is missing from the data received, but only on jenkins runs.
On 2 should be the agendaitem that is an approval (will be filtered from the list)
On 3 should be the agendaitem with a newsitem that was recently modified.

I thought it might be fixed with the new mu-auth. it's not
So something is going on there.
I thought maybe cross spec interference, but it really doesn't seem possible.
The tests fail halfway through the sorting
Checking the "in newsletter" box works, sorting on "in newsletter" directly works, sorting on number after that works. Sorting in modified > breaks, 1 is missing. 
This is a pain to debug because I can't get it to fail locally. 

Now I am attempting to sort on modified before checking "in newsletter" to identify if the sorting breaks because we check the box (maybe hit a cache invalidation somewhere)
And I have added a small wait right after checking the box and saving the news-item.

The fact that it seems impossible to break locally is somewhat good news, it means there is no immediate issue that would occur with users. But having a near green jenkins run is bothersome.
If all else fails, I might move that sorting test in a separate spec and run in before everything else (all-flaky-tests folder) to see if that helps

sidenote: I noticed pdf-signature-remover could not access files and was throwing errors on jenkins, gave it permissions to write and access to share folder.
I should add a spec that tests that pdf-stripping feature so we will need this eventually

### DIFF
--- a/Jenkinsfile.ci
+++ b/Jenkinsfile.ci
@@ -69,6 +69,8 @@ pipeline {
         sh "cd ${WORKSPACE}/kaleidos-project && docker-compose -p ${projectName} exec -T document-stamping chmod 777 -R /share"
         sh "cd ${WORKSPACE}/kaleidos-project && docker-compose -p ${projectName} exec -T draft-file chmod 777 -R /share"
         sh "cd ${WORKSPACE}/kaleidos-project && docker-compose -p ${projectName} exec -T draft-file-mover chmod 777 -R /share"
+        sh "cd ${WORKSPACE}/kaleidos-project && docker-compose -p ${projectName} exec -T pdf-signature-remover chmod 777 -R /share"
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
Probably going to close this, but I do want to keep the changes to jenkins files

ValenberghsSven:
closing for now, it seems less flaky now. Will try again later. Grabbed the jenkins settings and added on DEV

PR #2229, opened 2024-10-11
KAS-4814 improve internal review by making a component with editing f…

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4814

see comments on 
https://kanselarij.atlassian.net/browse/KAS-4807

Document viewer comment is read-only on submissions docs. this is unwanted by users.
But we don't want to allow edits on draft pieces in that view.
Solution? A reusable component for that `interalReview`
Save and cancel is handled in the component. reduces duplication in the templates/javascript code.

### DIFF
--- a/app/components/documents/document-details-panel.hbs
+++ b/app/components/documents/document-details-panel.hbs
@@ -41,17 +41,6 @@
           data-test-details-tab-edit-access-level
         />
       </Auk::KeyValue>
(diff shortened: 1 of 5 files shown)


ValenberghsSven:
only changing document-viewer here. changing description panels will lead us astray.

PR #2230, opened 2024-10-12
KAS-4834: Add persagenda view

sergiofenoll:


app/components/agenda/press/list-section/item-group.hbs
sergiofenoll, line 1: This is a copy-paste of `Agenda::PrintableAgenda::ListSection::ItemGroup` with the bits we didn't need removed and replacing the CSS class `l-printable-agenda__subcase-name-indented` with `l-printable-agenda__subcase-name`.

app/components/agenda/press/list-section/item-group/item/content.hbs
sergiofenoll, line 1: This is a copy-paste of `Agenda::PrintableAgenda::ListSection::ItemGroup::Item::Content` with the bits we didn't need removed and replacing the CSS class so the text wasn't indented.

app/routes/agenda/press.js
```diff
@@ -0,0 +1,87 @@
...
+      include: 'mandatees,agenda-activity.subcase.type',
+    });
+
+    return {
+      meeting,
+      agendaitems: await this.sortAgendaitems(agendaitems),
+    };
+  }
+
+  async sortAgendaitems(agendaitems) {
```
sergiofenoll, line 42: Re-implementation of [the code from the `agenda-submission` service](https://github.com/kanselarij-vlaanderen/agenda-submission-service/blob/main/lib/agendaitem-order.js#L91-L130).

ValenberghsSven (changes requested):
the ticket mentions:

- een trigger om agenda aan te maken _(we now navigate to a view page)_
- een automatisch gegenereerde persagenda (als pdf) op basis van de agenda _(not done, how are they expecting this agenda? generate a pdf for every press of the action?)_
- de mogelijkheid om deze te downloaden  _(guessing you can print the page, but I think they wanted to download the pdf from above)_
- Deze functie is enkel beschikbaar voor admin & secretarie _(not done, everyone can see this, also allow kanselarij to view this)_

sergiofenoll:
> the ticket mentions:
> 
>     * een trigger om agenda aan te maken _(we now navigate to a view page)_
> 
>     * een automatisch gegenereerde persagenda (als pdf) op basis van de agenda _(not done, how are they expecting this agenda? generate a pdf for every press of the action?)_

Ticket also links to a Usertest view where none of this is implemented and instead a view like the "printable agenda view" is implemented. Given that the ticket content is just a copy-paste of the Feature Passport from Design, I assumed that the FP was out-of-date with the actual expected implementation from Usertest.
 
>     * de mogelijkheid om deze te downloaden  _(guessing you can print the page, but I think they wanted to download the pdf from above)_

Yes they can just print the page.

>     * Deze functie is enkel beschikbaar voor admin & secretarie _(not done, everyone can see this, also allow kanselarij to view this)_

I forgot about this.

ValenberghsSven:
@tdn 
Creating a pdf with just the content or printing directly from your browser will be very different.
Browser print contains a lot of data in the header and footer. Has this been checked/verified with the users if this is ok?
For me this works, but then we should update the ticket since Kenny will not be looking at usertest links.

ValenberghsSven (approved):
fixed small bugs, added permission

tdn (approved):

PR #2232, opened 2024-10-16
fix wrong container position when adding new documents when BIS exist…

ValenberghsSven:
No ticket, noticed this in other PR's.
This shouldn't wait.

fix wrong container position when adding new documents when BIS exists
fix stamping BIS when original was never stamped

### DIFF
--- a/app/components/subcases/subcase-header.js
+++ b/app/components/subcases/subcase-header.js
@@ -52,10 +52,10 @@ export default class SubcasesSubcaseHeaderComponent extends Component {
 
   get canMove() {
     return (
-      (this.loadData.isIdle &&
+      this.loadData.isIdle &&
         this.currentSession.may('manage-agendaitems') &&
-        !this.args.parliamentFlow) ||
-      (this.parliamentRetrievalActivity && this.args.subcases.length === 1)
+        (!this.args.parliamentFlow ||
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
for the BIS stamping bit.
When you add new documents to an agendaitem > no stamps (no VR number yet)
When you then added a BIS to those document we were going to stamp the BIS (no VR number yet)
Since we stamped already, we don't restamp after the VR number was assigned on agenda approval.
I fixed that so we don't stamp on adding BIS if no stamp exists on previous versions.

What still exists: if you add documents on an approved agenda/agendaitem we will always stamp documents (even if we have no VR number) so the name of the piece must be correct when adding the documents (adding docs to approved agenda shouldn't happen, but it does).
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/controllers/agenda/agendaitems/agendaitem/documents.js#L185

Should we re-evaluate when stamping should occur?
Should we stamp documents we have not numbered first ? (no `piece.originalName`)

side note ->
I'm all for a less aggressive stamping in edge cases and a manual "stamp this document" button.
Just to avoid:
- we stamp documents that already have the VR number in it (like legacy edits).
- we stamp documents that might not have a VR number yet. (like what was happening here)
- we stamp documents we already had stamped before, but the user replaced the file with a pdf that already contains a manually stamped VR number (for reasons unknown)

PR #2237, opened 2024-10-18
Allow minister profiles to perform all kdb actions

tdn:
Only change that limits this profile: `'manage-only-specific-signatures'` was added to `MINISTER`.
Discussed this with Kenny, and should be OK.

### DIFF
--- a/app/config/permissions.js
+++ b/app/config/permissions.js
@@ -287,6 +287,10 @@ const groups = [
       ],
       full: [
         'manage-signatures',
+        'manage-only-specific-signatures',
+        'send-only-specific-cases-to-vp',
+        'create-submissions',
+        'edit-sent-back-submissions',
       ]
     }
   },
--- a/cypress/e2e/all-flaky-tests/profile-minister.cy.js
+++ b/cypress/e2e/all-flaky-tests/profile-minister.cy.js
@@ -639,6 +639,7 @@ context('Testing the application as Minister user', {
     it('check signatures/start route', () => {
       cy.visit('ondertekenen/opstarten');
       cy.get(appuniversum.loader).should('not.exist');
+      cy.get(route.signatures.dataTable);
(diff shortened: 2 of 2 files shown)


app/config/permissions.js
```diff
@@ -287,6 +287,10 @@ const groups = [
       ],
       full: [
         'manage-signatures',
+        'manage-only-specific-signatures',
+        'send-only-specific-cases-to-vp',
+        'create-submissions',
```
ValenberghsSven, line 292: needs graph access to submissions

app/config/permissions.js
```diff
@@ -287,6 +287,10 @@ const groups = [
       ],
       full: [
         'manage-signatures',
+        'manage-only-specific-signatures',
```
ValenberghsSven, line 290: do these 2 play well together? the "manage-signatures"  > you can manage all of them.
"manage-only-specific-signatures" > you can only manage yours

tdn: `KABINET_DOSSIERBEHEERDER` profile already had both, so I assume they do work together. `manage-only-specific-signatures` basically removes the filter options and pre-fills them in the Signatures tab. It also makes sure you can't mark other cabinets documents for signature in the document preview sidebar

app/config/permissions.js
tdn: https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/499

Will also update the new mu-auth PR

tdn: Apparently @sergiofenoll is clairvoyant, because it's already in the new mu-auth ;-) https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/469/files#diff-c0623dcb5422872d71dd79b855fe9a11e8c09c6b585c53799c5a0f9ef54639dfR522

PR #2241, opened 2024-10-29
KAS-4860: Prevent unload

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4860

Prevents users from leaving/unloading/reloading the page when they are submitting a new submission. You can either finish the submission, click on cancel, or be pestered by pop-ups!


tdn:
Not a complete review yet, but noticed a few things already:

- the prevent unload modifier seems to work on a new submission when trying to navigate using the browser, but not when clicking on any of the links at the top of the page

- I could still navigate away on an update (BIS) submission, thereby losing my submission. 

- documents on update submissions open in a new tab, but their previous versions don't
image

Could be a local issue on my machine. Perhaps a pair review would be a good idea

sergiofenoll:
@tdn Maybe we can check together in a call tomorrow after the standup? The last point I totally missed and is an oversight, the first two didn't seem to be broken locally :thinking:

tdn:
@sergiofenoll perfect, let's see tomorrow morning

tdn (approved):

PR #2242, opened 2024-10-29
KAS-4848: Update submission with latest info when taking into treatment

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4848

When we take in a submission for treatment, we first update it with the info coming from the subcase (since cabinet members cannot update this information, we're never overwriting their changes). This ensures that if the secretary has made changes to a subcase that will get a new submission, their changes aren't overwritten by stale data from the cabinets' submission.

### DIFF
--- a/app/components/submission/header.js
+++ b/app/components/submission/header.js
@@ -393,7 +393,15 @@ export default class SubmissionHeaderComponent extends Component {
   );
 
   takeInTreatment = async () => {
-    // TODO update submission data? It could have been changed on subcase
+    const subcase = await this.args.submission.subcase;
+    if (subcase) {
+      const subcaseType = await subcase.type;
+      this.args.submission.shortTitle = subcase.shortTitle;
+      this.args.submission.title = subcase.title;
+      this.args.submission.type = subcaseType;
+      this.args.submission.subcaseName = subcase.subcaseName;
+      this.args.submission.confidential = subcase.confidential;
+    }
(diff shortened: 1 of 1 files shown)


app/components/submission/header.js
```diff
@@ -393,7 +393,15 @@ export default class SubmissionHeaderComponent extends Component {
   );
 
   takeInTreatment = async () => {
-    // TODO update submission data? It could have been changed on subcase
+    const subcase = await this.args.submission.subcase;
+    if (subcase) {
```
tdn: checking `subcase?.id` here is a little bit better, since a Promise would also be truthy.

app/components/submission/header.js
```diff
@@ -393,7 +393,15 @@ export default class SubmissionHeaderComponent extends Component {
...
+      const subcaseType = await subcase.type;
+      this.args.submission.shortTitle = subcase.shortTitle;
+      this.args.submission.title = subcase.title;
+      this.args.submission.type = subcaseType;
+      this.args.submission.subcaseName = subcase.subcaseName;
+      this.args.submission.confidential = subcase.confidential;
+    }
     await this._updateSubmission(CONSTANTS.SUBMISSION_STATUSES.IN_BEHANDELING);
```
tdn, line 405: Just as a note: this will save the submission, and thus also save the properties we just set.

tdn: The whole "modifying args" thing is not really DDAU, but this is certainly not the only place we do this. Fine for now

tdn (approved):
Works well locally.

PR #2243, opened 2024-11-06
KAS-4678 use sections everywhere for custom rdfa editor parts

ValenberghsSven:
Using id's no longer seems to work easily.
Even if we fix this now, it may end up breaking again later.
Our custom html element we have set up seems to keep working so it seemed optimal to just use that everywhere we need some custom edits.

### DIFF
--- a/app/controllers/agenda/minutes.js
+++ b/app/controllers/agenda/minutes.js
@@ -242,11 +242,6 @@ ${renderNextMeeting(meeting, intl)}`;
       newContentElement.content.querySelector('[data-section="next-meeting"]').innerHTML = nextMeetingText;
     }
 
-    // const nextMeetingText = contentElement.content.querySelector('#next-meeting')?.innerHTML;
-    // if (nextMeetingText) {
-    //   newContentElement.content.querySelector('#next-meeting').innerHTML = nextMeetingText;
-    // }
-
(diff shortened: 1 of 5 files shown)


app/controllers/agenda/minutes.js
```diff
@@ -16,27 +16,29 @@ function renderAttendees(attendees) {
...
+          <tr>
+            <td>De viceminister-presidenten</td>
+            <td>${viceMinisters.join('<br/>')}</td>
+          </tr>
+          <tr>
+            <td>De Vlaamse ministers</td>
+            <td>${ministers.join('<br/>')}</td>
+          </tr>
+          <tr>
+            <td><section data-section="secretary-title"><p>De ${secretaryTitle}</section></p></td>
```
ValenberghsSven: using `<td><section ..>De <span> ${secretaryTitle}` created newlines after "De" so opted to include it in the p element. Also added the "De" to the calculation of `${secretaryTitle}` when replacing since we will be replacing the entire `<p>` element now.
This solution looks ok.

ValenberghsSven: I already fixed the incorrect closing of html seen in this snippet (open section, open p , close section, close p)

brenner-company:
Looks good, thanks for the help!

PR #2246, opened 2024-11-11
KAS-4865: Auk-additions die uit prototype overgebracht zijn overlopen en refactoren/verplaatsen waar nodig

brenner-company:
**Tickets binnen Jira:** 
- https://kanselarij.atlassian.net/browse/KAS-4865

⚠️ Moet op zich niet echt gereviewed worden, maar alle tests moeten wel volledig ok zijn om te weten dat er geen functionele impact is.


app/components/search/result-cards/agendaitem.hbs
```diff
@@ -26,12 +26,12 @@
     <AuHeading @level="3" @skin="6" class="au-u-margin-top-tiny">
       <AuLink
         data-test-route-search-result-cards-agendaitem-shorttitle
-        class="au-u-padding-none card-link"
+        class="au-u-padding-none l-result-card__link"
```
tdn, line 29: We rely on this class to avoid double clicks. Will need to be fixed before this can be merged

tdn (changes requested):

app/components/search/result-cards/agendaitem.hbs
```diff
@@ -26,12 +26,12 @@
     <AuHeading @level="3" @skin="6" class="au-u-margin-top-tiny">
       <AuLink
         data-test-route-search-result-cards-agendaitem-shorttitle
-        class="au-u-padding-none card-link"
```
ValenberghsSven, line 29: `clickEvent?.target.parentElement.className.indexOf('card-link') > -1`

brenner-company:
Good catch @tdn & @ValenberghsSven. Will fix this!

brenner-company:
> Good catch @tdn & @ValenberghsSven. Will fix this!

Should be fixed now, but I'll wait for the tests to be completely certain.

brenner-company:
@ValenberghsSven Can you take a look at the test when you find the time? Keep getting a `ERROR: Error cloning remote repo 'origin'` within Jenkins before the actual tests run.

app/controllers/search/agendaitems.js
```diff
@@ -105,7 +105,7 @@ export default class AgendaitemsSearchController extends Controller {
   navigateToAgendaitem(searchEntry, clickEvent) {
     if (searchEntry.meetingId) {
       // Check if we clicked an emphasis inside a linkTo or a linkTo
-      if (clickEvent?.target.parentElement.className.indexOf('card-link') > -1 || clickEvent?.target.className.indexOf('card-link') > -1) {
+      if (clickEvent?.target.parentElement.className.indexOf('l-result-card__link') > -1 || clickEvent?.target.className.indexOf('l-result-card__link') > -1) {
         // do nothing, this was a clicked link in the card and the router will transition later
         return;
```
ValenberghsSven, line 110: I testes this locally, could it be that this behaviour has changed? (maybe part due to ember upgrade)
What it should do:
When clicking the card you enter the else part, no link was clicked and we `transitionTo` here.
When clicking the link inside the card you would transition twice, once with the LinkTo, once with our `router.transitionTo` so you would get double entries in history.
Now however, when debugging both the if and else, I never hit the return in the if statement when clicking the link. But I also don't hit transitionTo. I only hit that when not clicking a link.

Also a possible issue.
When clicking the link in the card we also stopped hitting our plausible tracking event.
Plausible click are only registered on card clicks, not link clicks
The behaviour on ` @onClickRow={{this.resultClicked}}` seems to have changed.
I think this is besides that styling changing here and should be looked at independently

ValenberghsSven (approved):
I've looked at the style changes, looks ok to me.
The result card thing is not an issue for this PR. Behaviours have changed and that feature needs to be looked at (it actually works better now, but we lose our plausible tracking it seems )

PR #2250, opened 2024-11-19
dispatcher is sending it to /data/torii/redirect.html for some reason

ValenberghsSven:
bit stuck, frontend is logging this:
[error] 101#101: *89 open() "/data/torii/redirect.html" failed (2: No such file or directory), client: 172.20.0.37, server: , request: "GET /torii/redirect.html?code=ETz_AocFOm0 HTTP/1.1", host: "kaleidos-dev.vlaanderen.be"


ValenberghsSven:
so the reason that failed is that we still use the dispatcher to forward `/authorization/callback` to frontend (torii/redirect.html route).
But the newer way to use oauth doesn't seem need forwarding. the dispatcher call is not sending the call to the right place.
Just need to have a router.js change in frontend so that `/authorization/callback` is a valid page (refactored from torii/redirect)

ValenberghsSven:
Don't fully understand it, but managed to login with itsme with this backend branch + the frontend branch
Sad flows (errors etc) are a little harder to check

PR #2252, opened 2024-11-20
KAS-4908 upgraded mock-login package to most recent

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4908

Upgraded minor version gradually.
Only real breaking change for us as v0.8 to v0.9 where some components were removed.
changing version to v0.7 to v0.8 warned about the deprecation and fixed it then.

### DIFF
--- a/app/templates/mock-login.hbs
+++ b/app/templates/mock-login.hbs
@@ -30,12 +30,15 @@
                 {{/if}}
                 {{#if this.model.length}}
                   <AuList
-                    class="auk-u-mb"
+                    class="auk-u-mb auk-table--clickable-rows"
                     data-test-mock-login-list
                     as |Item|
                   >
-                    <login.each-account @accounts={{this.model}} as |account|>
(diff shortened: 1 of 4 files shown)


ValenberghsSven:
There is an issue with cypress tests here, we seem to be losing our session
I rubber-ducked myself in rocket chat https://chat.semte.ch/channel/mu-semtech/thread/vik9m2BD54ibkNZvi
I think I have found the issue.
When we login, we usually instantly visit pages.
Most pages have `simpleAuthSession.requireAuthentication` in the beforeModel of the parent route.
It seems that visiting multiple of those routes in quick succession (cypress quickness) is causing the issue.
fe: system.alert.spec. The last test about datepicker.
We login in the before (visit agendas overzicht) > visit systeemmeldingen > visit an agenda link
this is flaky
We login in the before (visit agendas overzicht) > wait 1 sec > visit systeemmeldingen > wait 1 sec > visit an agenda link seems very robust

ValenberghsSven:
Ended up using v0.9.0 package, v0.10.0 removed ember-fetch which seems needed on some level.
The developers are aware

PR #2254, opened 2024-11-21
KAS-4891: Mark piece for signing when submitting BIS

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4891


ValenberghsSven (approved):
works

PR #2256, opened 2024-11-26
Feature/ember upgrade attempt v5.3

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4922

from v5.0 to v5.3

Fixed a few deprecations now that would be an issue in v6:
- Using ember-inflector like we were doing is deprecated, change to ember internal package
- We had an incorrect type in model `hasMany('themes')` instead of `hasMany('theme')` . Model itself is called `Theme`
- Got a deprecation warning that required us to manually create a transform for number, string and boolean (with import as default). So I did.
image


ValenberghsSven:
No major breaking tests, only known flakys

PR #2258, opened 2024-11-26
Feature/ember upgrade attempt v5.12

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4924

v5.8 to v5.12

- fixed unnecessary braces in templates (linter error)
- added 3 rules to linter:
- we should stop using `@ember/runloop`
- we should stop using `{{did-insert}}` and `{{did-update}}`, need alternative or custom implementation
- we should look at places where we use args to update tracked properties.

Other than those, upgrade went smoothly


ValenberghsSven:
No major breaking tests, only known flakys

PR #2261, opened 2024-11-28
Feature/kas 4877

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4877

The default here was a bit weird.
In other views (like signatures) we select a default and don't allow you to change it (we hide the filter)
In this case, if we just recalculate the default when there are no params.submitters, it gave side effects when toggling off all checkboxes. We were filtering in data but not visually, checkbox was not checked.
I found a way to fix it to work seemingly fine.
When entering the route, we calculate the default mandatees (1 or more)
if you have no params selected, get those defaults and pass them on to the controller (shown checked in filter)
Any edits to the checkboxes then will persist, defaults will not be recalculated.
When you leave the route and come back, it stays persisted (ember does this with the params) 
When you clear all checkboxes and leave, we no longer have params, and will select the defaults again.

I had an issue where having your mandatee checked, checking a second and unchecking that second resulted in a wrong list because we didn't hit the beforeModel/model route when doing that. Managed to fix that by always setting `this.submitters` in the route


tdn (approved):

tdn:
Only one failed test, which is a coincidence because it's a new year in 4 weeks: https://jenkins.kanselarij.org/job/kaleidos/job/FeatureBranches/job/PR-2261/2/artifact/reports/cypress_screenshots/screenshots/unit/agenda.cy.js/Agenda%20tests%20--%20Should%20create%20agenda,%20switch%20to%20PVV,%20then%20switch%20back%20to%20MR%20(failed).png

Merging this

PR #2264, opened 2024-12-02
KAS-4888 implement highlighted pieces when accepted but not propagated

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4888

Added query in route, set result on controller as tracked, pass to component that accepts `@highlightedPieces`

### DIFF
--- a/app/routes/cases/case/subcases/subcase/index.js
+++ b/app/routes/cases/case/subcases/subcase/index.js
@@ -105,6 +105,14 @@ export default class CasesCaseSubcasesSubcaseIndexRoute extends Route {
       const documentPublicationStatus = await documentPublicationActivity?.status;
       this.documentsAreVisible = documentPublicationStatus?.uri === CONSTANTS.RELEASE_STATUSES.RELEASED;
     }
+
+    // show accepted but not propagated pieces as "new" for certain profiles
+    if (this.currentSession.may('create-submissions')) {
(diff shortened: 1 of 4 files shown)


app/routes/cases/case/subcases/subcase/index.js
```diff
@@ -105,6 +105,14 @@ export default class CasesCaseSubcasesSubcaseIndexRoute extends Route {
       const documentPublicationStatus = await documentPublicationActivity?.status;
       this.documentsAreVisible = documentPublicationStatus?.uri === CONSTANTS.RELEASE_STATUSES.RELEASED;
     }
+
+    // show accepted but not propagated pieces as "new" for certain profiles
+    if (this.currentSession.may('create-submissions')) {
```
tdn: I would not perform this check, and rather show the `piecesNotOnAgenda` for all profiles. 
I think it could be useful to avoid mistakes such as when a kanselarij employee would upload a document on a subcase, but not on the agenda. That way it is instantly clear that this piece is not visible on any agenda.

TBD tomorrow

tdn: Scenario: adding a document on the subcase after the agenda has been closed.
Could technically happen, and would entail a very hard to spot user error.

Without the "if" it's crystal clear:
image

I removed the check

tdn (approved):

PR #2265, opened 2024-12-04
Feature/kas 4895 cases filter definitive

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4895

Thought the negative query filter was not possible out of the box, but there is an option we have not used anywhere namely `:not:`.
It has some limitations it seems. using it in combination with ID or URI does not work (406 from resource)


tdn (approved):
Works well!

PR #2266, opened 2024-12-05
KAS-4805 implement emptying interal review private comment feature

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4805

Noticed that `agendaitem.privateComment` is still being filled in on PROD.
So in this code I added it's removal. That does mean saving agendaitems so I copied some the "approveAllAgendaitems" to show a loader and refreshing the route.
This is only needed untill we remove that property from agendaitem model in follow up ticket.

1 side tracking: I discovered that we used `agenda.serial**N**umber` instead of the correct `agenda.serialnumber` in some places.
Need to check if there are any consequences of that.


app/controllers/newsletters/index.js
```diff
@@ -38,7 +38,7 @@ export default class NewslettersIndexController extends Controller {
     const agendas = await meeting.agendas;
     return agendas
       ?.slice()
-      .sort((a1, a2) => a1.serialNumber.localeCompare(a2.serialNumber))
+      .sort((a1, a2) => a1.serialnumber.localeCompare(a2.serialnumber))
```
ValenberghsSven, line 41: this is sidetracking, should not be capital N. No errors ever came of this? `latestAgenda` could have been incorrect in this view

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -352,6 +353,43 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+    const isDesignAgendaA = this.args.currentAgenda.status.get('isDesignAgenda') && this.args.currentAgenda.serialnumber === 'A';
+    return this.currentSession.may('manage-agendaitems') && !isDesignAgendaA;
+  }
+
+  emptyInteralReviews = async() => {
+    this.showConfirmEmptyInternalReviews = false;
+    this.args.onStartLoading(this.intl.t('empty-internal-review'));
+    // getting all valid agendaitems first gets better results than trying submission-internal-review directly via subcase
+    const approvedAgendaitems = await this.store.queryAll('agendaitem', {
+      'filter[:has:previous-version]': true,
```
tdn: This won't work with a single, approved agenda

app/components/agenda/agenda-header/agenda-actions.js
```diff
@@ -352,6 +353,43 @@ export default class AgendaAgendaHeaderAgendaActions extends Component {
...
+    })
+    if (internalReview?.id && !isEmpty(internalReview.privateComment)) {
+      internalReview.privateComment = '';
+      return await internalReview.save();
+    }
+    // This property is still filled in as of now, we should empty it to avoid showing this briefly in agenda overview
+    // TODO KAS-4886 remove when property is removed from model
+    if (!isEmpty(agendaitem.privateComment)) {
+      agendaitem.privateComment = '';
+      return await agendaitem.save();
```
tdn: Could have some unintended effects on approved agendas that have a next version. Will need to implement an extra check here, or wait until https://kanselarij.atlassian.net/browse/KAS-4886 is done.

tdn (changes requested):
One small remark left. Works well otherwise in happy flow.

app/components/agenda/agenda-header/agenda-actions.js
ValenberghsSven: Tom fixed this during initial pair review

app/components/agenda/agenda-header/agenda-actions.js
ValenberghsSven: fixed this by passing `reverseSortedAgendas` from args. We can use that to determine the latest agenda without awaiting. Improved check to also only allow the action on the latestAgenda.

ValenberghsSven:
In our call to the backend we still passed the privateComment so both internal review and agendaitem model was set.
I put that on `null` in this ticket so we stop doing that for agendaitem, which would mean less or next to no `agendaitem.save` on new data.
Cleanup in follow up ticket to remove the code, also in `agenda-submission`

app/controllers/newsletters/index.js
ValenberghsSven: I accidentally found where this has an effect.
right here (screenshot on PROD)
image

If there is more then 1 agenda version > no modified.
Only modified when there is only agenda A.
This should be fixed after merging this.

ValenberghsSven:
Pair reviewed with Tom

PR #2267, opened 2024-12-06
fix void 0 is not a function

ValenberghsSven:


ValenberghsSven (commented):
this works (hosted container locally with prod build)

FYI, the dockerfile image upgrade didn't help, but we should grab it regardless?

PR #2273, opened 2024-12-11
KAS-4938: Disable remove piece for non-admin when VP flow exists

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4938

### DIFF
--- a/app/components/documents/draft-document-card.js
+++ b/app/components/documents/draft-document-card.js
@@ -39,6 +39,7 @@ export default class DocumentsDraftDocumentCardComponent extends Component {
 
   @tracked piece;
   @tracked documentContainer;
+  @tracked retrievedPieces;
 
   @tracked uploadedFile;
   @tracked newPiece;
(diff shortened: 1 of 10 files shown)


app/components/documents/draft-document-card.js
```diff
@@ -82,6 +83,12 @@ export default class DocumentsDraftDocumentCardComponent extends Component {
     return isPresent(this.args.bordered) ? this.args.bordered : true;
   }
 
+  get mayDelete() {
+    return (
+      this.retrievedPieces?.length === 0 || this.currentSession.may('remove-piece-from-parliament')
```
tdn: Draft pieces can never have retrievedPieces, right?

So this will always be falsy for secretarie and minister/KDB I think.

Also, currentSession is not defined here...

Was this tested?

tdn (changes requested):

app/components/documents/draft-document-card.js
ValenberghsSven: removed the code from draft-document-card.
The `model.pieces` here should only be draft-pieces which don't have the `retrieved-piece` relation.
Also, it was now impossible to remove any draft piece as non admin even if it was your own submission.

ValenberghsSven (approved):
reviewed this with Tom.
Waiting for cypress run to check nothing has broken

ValenberghsSven:
only known flaky tests failed that have been fixed on DEV

PR #2274, opened 2024-12-11
KAS-4890 PR review WIP

ValenberghsSven:
- There was no button to "take in treatment".
- There was no button to "request sendback"
- History showed no label
- Wrong meeting with BIS updates of a resubmitted submission (selected the meetin of the original submission) ( :warning:  TODO this is still unsolved because using previousSubmission instead was undefined (it shouldn't)
- BIS update was creating without meeting because it received a pointerEvent instead of a meeting
- decisionResultCode is optional if not propagated, threw errors in console
- It was possible to edit all the information of a new submission from postponed subcase. (this is technically also "isUpdate")
- It was possible to create a BIS update before the agendaitem on new agenda was propagated (the accepted-piece was already propagated but was not visible because the submission-activity was not propagated
- changed a dutch named variable because reasons (apparently "lidwoord" > "article". :eyes: who came up with that)


ValenberghsSven:
A test fails because there is no link to agenda when it was expected.
Will be related to this.

ValenberghsSven:
Found it, in the agenda-submission we are expecting `?meeting`  to have `adms:identifier` predicate.
In our test data, not all meetings have that.
checked PROD data.
Not all meetings have that predicate. so it has to be optional in the query in the service

ValenberghsSven:
Merging this one in the main branch.
Tested this locally with many cases, seems ok now

PR #2277, opened 2024-12-20
KAS-4895 add filter for ratification subcases

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4895

Ratification subcases were showing (nobody asked to not have this)
Fixed it


ValenberghsSven:
1 thing here, this wont work with older subcases where there is no "ratification" document (pre VP integration)

tdn (approved):

tdn:
Will work better than the current solution. However, the best way to do this in the future (if the users ask), is via the search function. We could potentially index an `decisionmakingFlow.isFinal` property using a `script` in the index configuration based on the subcase types

PR #2278, opened 2024-12-20
KAS-4897: the great controllers reset

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4897

In deze ticket ben ik door alle controllers in de app gegaan en waar mogelijk kleine issues met het resetten van de controllers rechtgezet. Het ging voornamelijk over kleine dingen zoals modals die open bleven staan en `thing = store.createRecord()` die niet verwijderd worden.

Dit laatste doen we ook in componenten en daar zouden we dat ook moeten opkuisen, maar dat is misschien voor een andere ticket.

De eigenlijke issue die in de ticket beschreven werd is in f1209186273159b35c4a79362bd29065b5f7cfee gefixt geweest met (zoals Sven al aangaf in de comments van de ticket) een `reload()`. Eigenlijk is dat ook het enige dat écht gemerged "moet" worden.

### DIFF
--- a/app/routes/agenda/documents.js
+++ b/app/routes/agenda/documents.js
@@ -31,13 +31,18 @@ export default class AgendaDocumentsRoute extends Route {
     controller.defaultAccessLevel = this.defaultAccessLevel;
   }
 
-  resetController(controller, isExiting) {
+  resetController(controller, isExiting) { 
     if (isExiting) {
       controller.isOpenBatchDetailsModal = false;
       controller.isOpenPieceUploadModal = false;
-      
+
+      // cleanup any unsaved pieces
       Promise.all(
-        controller.newPieces.map(async (piece) => await deletePiece(piece))
+        controller.newPieces.map(async (piece) => {
+          if (!piece?.id) {
+            await deletePiece(piece);
+          }
+        }),
       ).then(() => (controller.newPieces = new TrackedArray([])));
     }
   }
--- a/app/routes/agenda/documents.js
+++ b/app/routes/agenda/documents.js
@@ -1,7 +1,9 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { TrackedArray } from 'tracked-built-ins';
 import CONSTANTS from 'frontend-kaleidos/config/constants';
 import { sortPieces } from 'frontend-kaleidos/utils/documents';
-import { inject as service } from '@ember/service';
+import { deletePiece } from 'frontend-kaleidos/utils/document-delete-helpers';
 
 export default class AgendaDocumentsRoute extends Route {
   @service store;
@@ -27,5 +29,16 @@ export default class AgendaDocumentsRoute extends Route {
     const agenda = this.modelFor('agenda').agenda;
     controller.agenda = agenda;
     controller.defaultAccessLevel = this.defaultAccessLevel;
+  }
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.isOpenBatchDetailsModal = false;
+      controller.isOpenPieceUploadModal = false;
+      
+      Promise.all(
+        controller.newPieces.map(async (piece) => await deletePiece(piece))
+      ).then(() => (controller.newPieces = new TrackedArray([])));
+    }
   }
 }
--- a/app/routes/agendas.js
+++ b/app/routes/agendas.js
@@ -75,14 +75,18 @@ export default class AgendasRoute extends Route {
 
   resetController(controller, isExiting) {
     if (isExiting) {
-      if (controller.newMeeting) {
-        controller.newMeeting.deleteRecord();
+      // delete any non saved models
+      if (!controller.newMeeting?.id) {
+        controller.newMeeting?.deleteRecord();
+      }
+      if (controller.publicationActivities.length) {
+        controller.publicationActivities.forEach((activity) => {
+          if (!activity?.id) {
+            activity.deleteRecord();
+          }
+        });
       }
       controller.newMeeting = null;
-
-      if (controller.publicationActivities.length) {
-        controller.publicationActivities.forEach((activity) => activity.deleteRecord());
-      }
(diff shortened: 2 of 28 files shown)


ValenberghsSven:
Avoided this one a bit. 
Tests are failing in this one.
I think some resets are to harsh, where documents that should be saved are actually deleted because of the reset.
Maybe because of a fast user (like cypress).
Perhaps I split this up into "safe" actions like reloads and "dangerous" actions like deleting pieces if we think they are to be removed.
That way, the "safe" actions can start moving while we finetune the "dangerous" ones.

ValenberghsSven (changes requested):
Avoided this one for a while.
Tests are failing in this one.
I think some resets are to harsh, where documents that should be saved are actually deleted because of the reset.
Maybe because of a fast user (like cypress).
Perhaps I split this up into "safe" actions like reloads and "dangerous" actions like deleting pieces if we think they are to be removed.
That way, the "safe" actions can start moving while we finetune the "dangerous" ones.

Just to get this PR moving a bit

app/routes/agendas.js
```diff
@@ -70,6 +70,21 @@ export default class AgendasRoute extends Route {
   setupController(controller) {
     super.setupController(...arguments);
     controller.defaultPublicationActivityStatus = this.defaultPublicationActivityStatus;
+    controller.isCreatingNewSession = false;
+  }
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      if (controller.newMeeting) {
+        controller.newMeeting.deleteRecord();
```
ValenberghsSven: tested this one:
You create a meeting
You open the meeting > we flag the just created meeting for deletion.
If you were to then change anything on the meeting and try to save you get errors
`Attempted to set 'extraInfo' on the deleted record Meeting:` 
So too aggressive.

Tried finding other scenarios where we do have a meeting "in flux" but can't find any.
You have to cancel to get out of it and canceling cleans up the meeting.
So what do we want to do with this? guessing just no delete here, just clean the controller maybe.

going to remove the deletes for this one.
We created unsaved meeting and publication activities.
So worst case the meeting is still in the controller.

app/routes/agenda/documents.js
```diff
@@ -28,4 +30,15 @@ export default class AgendaDocumentsRoute extends Route {
...
+
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.isOpenBatchDetailsModal = false;
+      controller.isOpenPieceUploadModal = false;
+      
+      Promise.all(
+        controller.newPieces.map(async (piece) => await deletePiece(piece))
+      ).then(() => (controller.newPieces = new TrackedArray([])));
+    }
```
ValenberghsSven: This one never seems to get called.
I can force it by commenting `newPieces = new TrackedArray([])` in the controller methods.
Then we do hit this, but again, we delete the pieces that should remain.
Going to change this so we first verify: "Is there an id? > Don't delete".
That should make it so we only target pieces that are safe to remove and not remove actual pieces.

app/routes/signatures/decisions.js
```diff
@@ -86,6 +86,12 @@ export default class SignaturesDecisionsRoute extends Route {
     }
   }
 
+  resetController(controller, isExiting) {
+    if (isExiting) {
+      controller.closeSidebar();
+    }
+  }
+
```
ValenberghsSven: duplicate `resetController`, going to keep the second one

app/routes/cases/case/subcases/subcase/new-submission.js
```diff
@@ -160,8 +161,21 @@ export default class CasesCaseSubcasesSubcaseNewSubmissionRoute extends Route {
...
+      controller.hasConfidentialPieces = false;
+      controller.comment = null;
+
+      Promise.all(
+        controller.newPieces.map(async (piece) => await deletePiece(piece))
+      ).then(() => controller.newPieces = new TrackedArray([]));
+      Promise.all(
+        controller.newDraftPieces.map(async (piece) => await deletePiece(piece))
+      ).then(() => controller.newDraftPieces = new TrackedArray([]));
+    }
```
ValenberghsSven: same, making it less aggressive.
No ID > `deleteRecord`

ValenberghsSven (commented):
Made some changes rather than pulling apart the PR.
Any deletes are softened, just trying to cleanup any model without any id.
And in some cases, don't delete the draft-file should that exist.
Because an unsaved piece of a draft-piece should not delete the draft-piece.

ValenberghsSven:
1 test is failing after changes.
Test does:
- Create meeting MR via modal in agendas
- Create meeting PVV via modal in agendas
But what happens is:
- after the first meeting is created the "agendas" view is refreshed so setupController will happen
- Cypress is faster so we open the modal again before the setupController happened (the button to create is available).
- the modal closes because of the setupController as it sets the `isCreatingNewSession` to false (this is new)

Can be done locally by using slow network.

I guess that behaviour is ok, it just happens slower then you would think.
Since only 1 spec is affected I will just add a check so we wait for the refresh of the page before continuing.

Just thinking if this use case of "fast users" could have more serious impact in other scenarios.
Since the setup and reset would happen _after_ the user possibly did something

app/routes/cases/case/subcases/subcase/new-submission.js
ValenberghsSven: changed

app/routes/signatures/decisions.js
ValenberghsSven: changed

app/routes/agenda/documents.js
ValenberghsSven: changed

app/routes/agendas.js
ValenberghsSven: changed

ValenberghsSven:
Wachten op jenkins groen

tdn (approved):

PR #2284, opened 2025-01-10
KAS-4951: Don't toggle changed every time you move an agendaitem

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4951

The previous code would toggle the "have the agendaitems been moved" flag every single time, which means that if you move two agendaitems (or a multiple thereof), the flag is set to false and no changes are persisted.


ValenberghsSven (approved):
looks good.
Can't believe we missed that.
Dragging and dropping is not easy in cypress so we don't have a test for it

PR #2285, opened 2025-01-22
Feature/fix acc flags kas 4896 kas 4890

ValenberghsSven:
Some fixes were needed for retracted/postponed subcase flows.
Difficult to get right in every single scenario


ValenberghsSven:
https://jenkins.kanselarij.org/job/kaleidos/job/FeatureBranches/view/change-requests/job/PR-2285/

Merging for now. Tested this a lot locally, going to test on ACC and users will also test

PR #2286, opened 2025-01-29
KAS-4890 fix mails for postponed subcase submissions

ValenberghsSven:


ValenberghsSven:
Tested this a lot, many different scenarios are available depending on when you make a BIS/TER update submission
Merging

PR #2287, opened 2025-02-05
add config to stop extending deprecated prototypes to environment file

ValenberghsSven:
No ticket, 5 minute thing to see if any tests fail when following the deprecation information

https://deprecations.emberjs.com/id/deprecate-array-prototype-extensions/

update:
ticket now for follow-up
https://kanselarij.atlassian.net/browse/KAS-5002


ValenberghsSven:
I peeked into the ember code for this warning.
I thought the deprecation was being thrown **because** we were actively using array prototype extension.
(and we were, I fixed them)
But the code that throws this warning is basically:
check the ENV properties > did you explicitly disable the array-prototype extentions? > no ? > warning.

Does that mean that if we disable it that there could packages who may use and can fail now?
I updated a lot of them with the ember upgrade to v5.
The upgrade to v6 will have to be more diligent in checking if those packages also stopped using the array prototype extensions.
I think anyway.

ValenberghsSven:
Going to merge this into the ember v6 upgrade branch

ValenberghsSven:
already included in ember upgrade branch

PR #2290, opened 2025-02-12
own implementation of attemptedTransition storage to support redirect after leaving the SPA

tdn:
The problem was that the session service tries to store the `attemptedTransition`, but fails because ACM/IDM login refreshes the whole page.

See https://ember-simple-auth.com/api/services_session.ts.html and https://github.com/mainmatter/ember-simple-auth/blob/master/packages/ember-simple-auth/src/-internals/routing.js#L33-L50

Fixed this by storing the route with all params manually on requireAuthentication, and reading it back in handleAuthentication.

There might be better ways to do this, but for now this should work.


ValenberghsSven:
Profile tests are failing for ovrb and KB.
the transition on login don't happen.
we don't hit the defaultRoute. 
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-4978-fix-redirect-after-acm-idm-login/app/routes/index.js#L9

Doesn't happen locally. Will check tomorrow

ValenberghsSven (approved):
Besides some jenkins issues this branch is ok. Code is good.

PR #2292, opened 2025-02-14
KAS-4989: Regenerate decision reports when submitting new agendaitem

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4989

When a new agendaitem is submitted, due to our automatic sorting it might be inserted in the middle of existing agendaitems. If those already had decision reports generated, they need to be regenerated: the decision reports mention the agendaitem's position, which might have changed.

The backend service now informs the frontend if the agendaitems were reordered. If true, the frontend will regenerate the decision reports for the meeting.

Related PRs:

- [ ] https://github.com/kanselarij-vlaanderen/agenda-submission-service/pull/16


ValenberghsSven (commented):
Tested this (2 existing agendaitems with decisions, 3rd to be injected in between them)
It works, pdf and title was changed, popup was shown.
The only thing I might comment:
We regenerated all decisions of the meeting (in my case: 2) while only 1 needed to be regenerated (item 2 became 3)

Guessing that that is ok, We can't know in the frontend which agendaitems have been reordered.

note: doesn't trigger when not needed (added a 4th that was added at the bottom, no popup, no regeneration)

So ok for me if it's ok for @tdn that we regenerate for the whole meeting.

app/services/agenda-service.js
```diff
@@ -182,6 +183,12 @@ export default class AgendaService extends Service {
     const agendaitem = await this.store.findRecord('agendaitem', json.data.id);
     await subcase.hasMany('agendaActivities').reload();
     await subcase.hasMany('submissionActivities').reload();
+    if (json.data.didReorder) {
+      const meeting = await this.store.queryOne('meeting', {
+        'filter[agendas][agendaitems][:id:]': json.data.id,
+      });
+      await this.decisionReportGeneration.regenerateDecisionReportsForMeeting.perform(meeting, true);
```
tdn: I'm a bit concerned that this will lead to issues with the large meetings. 
Wouldn't it be possible to do a query to all reports that need to be regenerated (basically all of those under the one we just added), and then pass those to `generateReplacementReports` instead?
(we also just need the notas)

So something like:
```
          const reports = await store.queryAll('report', {
            'filter[:has-no:next-piece]': true,
            'filter[:has:piece-parts]': true,
            'filter[decision-activity][treatment][agendaitems][agenda][created-for][:id:]': meeting.id,
            'filter[decision-activity][treatment][agendaitems][type]': NOTA,
            'filter[decision-activity][treatment][agendaitems][:gt:number]': agendaitem.number,
          });
          await this.decisionReportGeneration.generateReplacementReports.perform(reports, false);
```

tdn (changes requested):
Works, but could be a little inefficient due to regenerating the whole meeting. Added a suggestion for improvement in the comments.

app/services/agenda-service.js
ValenberghsSven: `generateReplacementReports` does not make new file names
so content is replaced but old file names are kept. checking how to make this better

app/services/agenda-service.js
```diff
@@ -182,6 +184,34 @@ export default class AgendaService extends Service {
...
+          await Promise.all(reports.map(async (report) => {
+            const agendaitem = await this.store.queryOne('agendaitem', {
+              'filter[:has-no:next-version]': true,
+              'filter[treatment][decision-activity][report][:id:]': report.id,
+            });
+            const documentContainer = await report.documentContainer;
+            const pieces = await documentContainer.pieces;
+            report.name = await generateReportName(agendaitem, meeting, pieces.length);
+            await report.belongsTo('file').reload();
+            await report.save();
```
ValenberghsSven, line 209: this is a duplication (I did this quickly), not sure if we should turn that into a util or service method too.

ValenberghsSven (approved):

PR #2293, opened 2025-02-18
KAS-4897 add concurrency checks on submissions

ValenberghsSven:
## :warning:  started from ACCEPTANCE 

https://kanselarij.atlassian.net/browse/KAS-4987

ticket stated concurrency check on request send back :white_check_mark:   
Also implemented concurrency checks on the other actions.
In short, when submission status is updated > other tabs that try to do an action will pick this up and stop with a warning (1 exception)
Right now I don't have any follow up. 
The refresh of only status does mean you already get to see the status, but we don't enforce users to refresh the page so you don't see all the changes yet.
It's possible to enforce something like that but felt even more out of scope.

### DIFF
--- a/app/components/submission/header.hbs
+++ b/app/components/submission/header.hbs
@@ -5,81 +5,83 @@
         <OnThisPageLinks @items={{this.items}} />
       </Auk::Toolbar::Item>
     </Auk::Toolbar::Group>
-    {{#if this.hasActions}}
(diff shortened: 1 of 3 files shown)


tdn (approved):
Works like a charm. Changed the notification text a little bit, and increased the timeout for the toast when taking in treatment.

PR #2294, opened 2025-02-19
KAS-4983: Disable clearing document type in selector

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4983

All documents in Kaleidos should have a type anyway, so by disabling this we prevent users from accidentally removing a document's type.


ValenberghsSven (approved):
Simple yet effective.
Just wondering if disallowing the removal of document types across the entire app is ok
(should only admin be able to do it maybe?)
But for the resolution of the issue at hand this is fine.
Can finetune later based on user requests

PR #2296, opened 2025-03-03
KAS-4984: Update decision code after changing agenda item type

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-4984

When changing the type of a subcase (and its agendaitem, but changing the type only happens from the subcase page), also reset the decision result code when applicable (i.e. when unset or when it's on "goedgekeurd"/"akte genomen").

Agendaitems which have been "ingetrokken" or "uitgesteld" will not get their decision result codes changed.

### DIFF
--- a/app/components/subcase/description-panel/edit.js
+++ b/app/components/subcase/description-panel/edit.js
@@ -153,7 +153,6 @@ export default class SubcaseDescriptionEdit extends Component {
       propertiesToSetOnAgendaitem,
       propertiesToSetOnSubCase,
       resetFormallyOk,
-      agendaitemTypeChanged,
     );
 
     if (this.confidentialChanged && this.args.subcase.confidential) {
--- a/app/components/subcase/description-panel/edit.js
+++ b/app/components/subcase/description-panel/edit.js
@@ -152,7 +152,8 @@ export default class SubcaseDescriptionEdit extends Component {
       this.args.subcase,
       propertiesToSetOnAgendaitem,
       propertiesToSetOnSubCase,
-      resetFormallyOk
+      resetFormallyOk,
+      agendaitemTypeChanged,
     );
 
     if (this.confidentialChanged && this.args.subcase.confidential) {
--- a/app/components/subcase/description-panel/edit.js
+++ b/app/components/subcase/description-panel/edit.js
@@ -312,8 +312,8 @@ export default class SubcaseDescriptionEdit extends Component {
       && (!oldDecisionResultCode
         || oldDecisionResultCode.uri === CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD)
     ) {
-      const kennisname = await this.store.findRecordByUri('concept', CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME);
-      decisionActivity.decisionResultCode = kennisname;
+      const acknowledgedResult = await this.store.findRecordByUri('concept', CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME);
+      decisionActivity.decisionResultCode = acknowledgedResult;
     }
     await decisionActivity.save();
   }
--- a/app/components/subcase/description-panel/edit.js
+++ b/app/components/subcase/description-panel/edit.js
@@ -241,6 +241,7 @@ export default class SubcaseDescriptionEdit extends Component {
         'filter[decision-activity][treatment][agendaitems][:id:]': agendaitem.id,
       });
       const pieceParts = await report?.pieceParts;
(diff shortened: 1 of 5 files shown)


app/services/agendaitem-and-subcase-properties-sync.js
```diff
@@ -84,9 +87,43 @@ export default class AgendaitemAndSubcasePropertiesSyncService extends Service {
...
+    const newType = await agendaitem.type;
+    const treatment = await agendaitem.treatment;
+    const decisionActivity = await treatment.decisionActivity;
+    const oldDecisionResultCode = await decisionActivity.decisionResultCode;
+
+    if (
+      newType.uri === CONSTANTS.AGENDA_ITEM_TYPES.NOTA
+      && oldDecisionResultCode?.uri === CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME
+    ) {
+      decisionActivity.decisionResultCode = null;
```
ValenberghsSven: not always null, depends if there is a report or not.

ValenberghsSven: So the scenarios would be:
If postponed or retracted > always keep decision result code regardless of report

Is there a report?
no:
- becomes nota: was "acknowledged" > should become null (the default. realistically atleast, notas only have a result when there is a report but can be changed manually)
- becomes announcement: is most likely null > should become "acknowledged" (the default)

yes: 
- **becomes nota >  should become "approved"**
- becomes announcement >  should become "acknowledged"


The one in bold is not covered here. 
Unsure how frequently this would happen that an announcement was made (or accepted via submission), a decision report made, then the realization that this should be a nota instead.
I hope never.
So I could accept always setting null with the slight risk the above occurs and they now have a nota without a decision result instead of an announcement without a decision result :sweat_smile: 
It might be safer not to set a result but users have now adapted to "the decision result is always set to approved when we create a report" so they might not notice a missing result

ValenberghsSven: Small note: there are multiple announcements missing result codes on PROD right now. will fix in a later ticket

app/components/subcase/description-panel/edit.js
```diff
@@ -152,7 +152,8 @@ export default class SubcaseDescriptionEdit extends Component {
       this.args.subcase,
       propertiesToSetOnAgendaitem,
       propertiesToSetOnSubCase,
-      resetFormallyOk
+      resetFormallyOk,
+      agendaitemTypeChanged,
```
ValenberghsSven: since the `subcase-description-panel` is the only place where the agendaitem-type could change it doesn't really make sense to pass it to the `agendaitem-and-subcase-properties-sync`
That service should only do what it says on the tin. Change those 2 models their properties (relations included), not the decision result of a decision activity linked to a subcase and agendaitem (directly or via treatment).

So I will probably move the code to the edit panel.

ValenberghsSven: side note, we duplicated the agendaitem type changes also to bekrachtiging panel.
Should we really allow changing type to announcement there? Out of scope, but we will need to copy this new code to that panel until we decide what to do with it (we could put the entire process in a util or service instead of duplicating it)

I would prefer: If it's a ratification subcase > don't allow change of type (doesn't make sense) > change subcase type to definitive first > refresh > you get the normal subcase panel > allow change of agendaitem type here

ValenberghsSven: Discussed with Tom, ratification should only be nota, cleaned up some code even if it is out of scope. (and made it so it's no longer possible to change the type on ratification)

app/components/subcase/description-panel/edit.js
```diff
@@ -289,4 +290,31 @@ export default class SubcaseDescriptionEdit extends Component {
...
+    ) {
+      const approvedResult = await this.store.findRecordByUri('concept', CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD);
+      decisionActivity.decisionResultCode = hasReport? approvedResult : null;
+    } else if (
+      this.agendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT
+      && (!oldDecisionResultCode
+        || oldDecisionResultCode.uri === CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD)
+    ) {
+      const kennisname = await this.store.findRecordByUri('concept', CONSTANTS.DECISION_RESULT_CODE_URIS.KENNISNAME);
+      decisionActivity.decisionResultCode = kennisname;
```
ValenberghsSven: in hindsight, might want to rename this dutch term before merging

ValenberghsSven (approved):

PR #2306, opened 2025-04-07
KAS-5023: Session polling

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-5023

Poll for the current session every minute. If the session is invalid (the backend services should return a 401 in that case), we invalidate the frontend session and show a warning toast. If the user then clicks away to a page, we will only show data that is visible to an unauthenticated user (i.e. pretty much nothing), but we don't necessarily automatically redirect them to the login page (this depends on what route the user visits).

Related PRs:
- [x] https://github.com/kanselarij-vlaanderen/mock-login-service/pull/6

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -2,6 +2,7 @@ import Service, { inject as service } from '@ember/service';
 import { get } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { isPresent } from '@ember/utils';
+import { later } from '@ember/runloop';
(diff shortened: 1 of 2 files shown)


ValenberghsSven:
added backend branch with mock-login bump. All tests are failing currently

- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/517

tdn (approved):
testing with ACM-IDM on dev

PR #2307, opened 2025-04-16
KAS-5031: Correct the sorting of ongoing signatures and filter out ratifications

sergiofenoll:
https://kanselarij.atlassian.net/browse/KAS-5031

The sorting on sign flows shouldn't use meeting.planned-start, this relation seems to be empty in a lot of cases and thus won't sort correctly.

As opposed to when fetching all the sign flows for ratifications, when filtering the ratifications we cannot base ourserlves on the document type: resources does not provide an "inverse filter". Instead, we choose to use the absence of the ratification-subcase relation as a way to find those sign flows that are not for ratifications.


ValenberghsSven (approved):
works

PR #2313, opened 2025-04-29
Add check to handle a logged out user who is still semi authenticated

ValenberghsSven:
Flag on ACC

After a forced logout, you still have a "valid" session since it is never completely invalidated.

Not a fan of using a "loggedIn" next to a "isAuthenticated" since it should mean the same thing.
Maybe renames are in order to better reflect what is happening.


tdn (approved):
Tested and works as expected. The 2 failing test cases do work when tested manually, so unrelated to this PR. Merging this to include in the PROD release

PR #2316, opened 2025-05-09
KAS-4810 Add util to clean pasted data in AuTextarea components

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4810

- Added a util
   - that can take a `ClipboardEvent` and prevent it from happening.
   - Using the AuTextareas cursor position, determine where the user pasted (and if it's a range or a single point)
   - Get the text from the ClipboardEvent and replace enters / reduce whitespace inside / trim whitespace on the edges
   - insert that text in the correct position of the original text
   - return the new "original" text to be set on model.property
- implemented that util in `subcase/description-panel/edit` and `submission/description-panel/edit`

note: any other AuTextarea where this needs to be implemented? Do we strip unicode?

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
+++ b/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.hbs
@@ -13,25 +13,27 @@
         </AuFormRow>
       {{/if}}
       <AuFormRow>
-        <AuLabel for="name-subcase">{{t "name-subcase"}}</AuLabel>
+        <AuLabel for="short-title-agendaitem">{{t "name-subcase"}}</AuLabel>
         <AuTextarea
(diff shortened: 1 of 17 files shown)


tdn (changes requested):
Works well, but would also be useful to have in the following places:
- Nieuwe indiening: app\templates\cases\new-submission.hbs
- Nieuwe indiening in Dossier: app\templates\cases\case\subcases\new-submission.hbs
- (checken, wss niet nodig) Nieuwe update-indiening app\templates\cases\case\subcases\subcase\new-submission.hbs
- Nieuwe procedurestap form: app\templates\cases\case\subcases\add-subcase.hbs
- Nieuw dossier
- Agendapunt titels
... (eventueel nog andere textareas afchecken)

tdn (approved):
Works really nicely

PR #2317, opened 2025-05-14
KAS-5049 add check for legacy meetings to just show documents

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5049

- added preKaleidos check in 2 places (both agenda views overview and details)
I think that should do it, need to test still. (subcase doesn't check decisionResultCode)


ValenberghsSven:
made a duplicate by mistake, removing this PR

PR #2318, opened 2025-05-14
KAS-5049 add check for legacy meetings to just show documents

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5049

- added preKaleidos check in 2 places (both agenda views overview and details)
I think that should do it, need to test still. (subcase doesn't check decisionResultCode)

NOTE: I didn't see the "hotfix to production" so I will have to cherry pick this comment rather than merging.
making draft PR to avoid merging

### DIFF
--- a/app/components/agenda/agenda-overview/agenda-overview-item.js
+++ b/app/components/agenda/agenda-overview/agenda-overview-item.js
@@ -105,6 +105,11 @@ export default class AgendaOverviewItem extends AgendaSidebarItem {
     }
     // no decisionResult after release
     this.documentsAreVisible = this.currentSession.may('view-documents-before-release');
+
+    // any legacy has no decisionResultCode, the document access level will determine who can view
+    if (this.args.meeting.isPreKaleidos) {
+      this.documentsAreVisible = true;
+    }
     return;
   }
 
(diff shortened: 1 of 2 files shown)


ValenberghsSven:
tested it locally.
Created an agenda with a date of 2018.
Added agendaitem with some documents.
approved, released agenda. released decisions and documents
No decision created (also no result code) > documents are visible on the 2 agenda views

Changed the meeting data to 2021.
After a cache issue (I kept seeing the 2018, had to reset cache and resource before 2021 came out) the documents were no longer showing (because no result code)

ValenberghsSven:
The jenkins link points to the wrong PR (the duplicate one PR 2317).
this one passed
https://jenkins.kanselarij.org/job/kaleidos/job/FeatureBranches/view/change-requests/job/PR-2318/

FYI, PR-2317 run failed because `digital-signing` was throwing errors on "unbound variable MODE" but that is fixed on development. So not sure why jenkins seems to have pulled the wrong image for that run (jenkins is weird sometimes, using wrong branches to test)

tdn (approved):
Verified on test with legacy and newer agendas

PR #2319, opened 2025-05-15
KAS-5052 add concept submissions, add permissions, adapt checks

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5052

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/520

- Added a concept submission
- only for first submissions (either new case or from existing case), not update submissions.
- concepts do not show in history
- concepts do not show in main datatable
- a toggle exists to only see the concepts of your own organization.
- other profiles can not see those concepts (even ministers can only see their own org) in the datatable
- Submission itself IS viewable by other orgs IF you share the link (since you can't find it naturally) and IF your ORG mandatee is on the submission aswell.
- edit only possible for the organization it belongs to (and admin, like any submission but you need the link)
- on submission we send the correct emails (first submission emails), from that point the flow is the normal flow.
- concepts can be deleted by the creator (or ORG linked), unlike normal submissions. (and admins)


ValenberghsSven:
PR reviewed by @tdn  and re-reviewed by me.

PR #2326, opened 2025-05-22
Implemented design feedback into concept submissions

tdn:
Quite a big rework of what @ValenberghsSven already had, so making this a separate PR to be merged into his.

1. Made the submissions and concepts 2 separate pages
2. Added filtering on title, but in the filter sidebar instead of in the header (we should do this for cases as well). Made the filter generic for this purpose
3. Added filtering on created date
4. Moved status pill to submissions overview-header
5. Cleaned up code in cases header that was for submissions & is no longer needed now


app/controllers/submissions/concepts.js
```diff
@@ -0,0 +1,117 @@
...
+    {
+      submissionFilter: {
+        type: 'string',
+      },
+    }
+  ];
+
+  @tracked page = 0;
+  @tracked size = PAGINATION_SIZES[2];
+  @tracked sortSubmissions = '-planned-start';
```
tdn: should be -created

app/controllers/submissions/concepts.js
```diff
@@ -0,0 +1,117 @@
...
+  @tracked dateFrom = null;
+  @tracked dateTo = null;
+  @tracked submitters = [];
```
tdn: not necessary

ValenberghsSven (approved):

PR #2328, opened 2025-05-23
KAS-5058 add reloads on possible stale relations

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5058

Just more reloads.

Noticed if you don't refresh and just keep clicking agendaitem documents we don't get new data.
This helps.
Won't do anything if there is an actual cache issue.


ValenberghsSven:
some of these reloads may be extra, but it shouldn't cause any issues

tdn (approved):
ok to merge if tests succeed

PR #2329, opened 2025-05-28
Feature/kas 5054 download documents more locations

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5054

- changed some existing translations to be more accurate with their key and added new keys
- Added a generic modal that can download any model's documents based on a path (dispatcher redirect needs to exist)
- Implemented that modal in `subcase/regular` (button), `agendaitem/documents` (button) and `subcases-overview-header` (as button and menuitem, depending on permissions)
- to avoid the button showing when there 0 files, added a `store.count` and used `documentsAreVisible` when applicable
- in `agendaitem/documents` I moved the `<Documents::LinkedDocuments>` since it was below a bunch of pop-up modals. it made more sense to have all visible elements in the top half of the file, then toggleable modals below that.
- Added generic function to construct a zip archive with optional name param (unused as of now but I suspect users might want their zips to be more uniquely named?)

- [x] https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/15
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/521


ValenberghsSven:
Reviewed with @tdn

PR #2330, opened 2025-05-28
KAS-5052 Review of changes, route change needed more updates

ValenberghsSven:
- the route "submissions" (/indieningen) was still visitable but nothing showed.
- best way would be to redirect from an actual index, and rename current index to ongoing (like search and publications)
- any transitionTo should be updated (the redirect would have fixed those, but better to be precise)
- removed the `toggleblefilters` from submission routes. Karel made a better fix after and it didn't flow to submissions it seems (use of `responsive` helper in template.
- changed the tests so we visit the correct route (redirect would have fixed it and even pass on the `?aantal=` param) still better to be precise
- removed 2 counts of unused `loadedDefaults`
- moved the shared submissions header to the index route so when switching between "ongoing" and "concepts" no longer forces the header to reload (the tabs "ongoing" and "concepts" were going away briefly)


ValenberghsSven:
https://jenkins.kanselarij.org/job/kaleidos/job/custom_frontend_backend_combination/390/

tdn (approved):

PR #2333, opened 2025-06-05
Feature/kas 5065 submission pieces access level

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5065


ValenberghsSven:
TODO: If individual document access levels change > check if notification addresses need to be updated

edit:
`DraftDocumentCard` already has that implemented. so should be ok on this branch

tdn (approved):
tested and seems to work well

PR #2334, opened 2025-06-06
KAS-4813 WIP, allow batch edit on submissions

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4813

continuation of access levels on submissions, focused on batch edits.

todo: list changes made for easier review


tdn (approved):
The functionality requested in the story works well.
 
I also tested one possible caveat I thought we didn't consider, but apparently we do: if a new version of is added as a draft=piece, and then is restricted to "vertrouwelijk", the access rights of the previous version(s) should also get updated. Indeed, they do, when the update is accepted by the admins/kanselarij. So works well!

PR #2336, opened 2025-06-10
KAS-4874 add required prop, add error class on missing doc-type

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4874

- added `documentTypeRequired` param (since there is 1 place where it is not required: agenda documents)
- added a getter that check if there is no `selectedOption`
- added a css custom class with a red border and background and use it when there is no option selected
- added an alert as well, which could be removed if not desired


tdn (approved):
Works well, only failing test is the flaky newsitem one.

PR #2337, opened 2025-06-13
Feature/kas 5057 concurrency

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5057

possible draft? don't think I will be able to fix more, since other possible issues are caused by users not communicating and combining actions that should not run together (like approving agenda and accepting submission. approving agenda and manually reordering agendaitems)

### DIFF
--- a/app/components/agenda/agendaitem/decision/digital.js
+++ b/app/components/agenda/agendaitem/decision/digital.js
@@ -181,6 +181,7 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component 
   loadReport = task(async () => {
     this.report = await this.args.decisionActivity.belongsTo('report').reload();
     if (this.report) {
+      await this.report.belongsTo('file').reload();
       await this.loadAnnotatiePiecePart.perform();
       await this.loadBetreftPiecePart.perform();
       await this.loadBeslissingPiecePart.perform();
--- a/app/controllers/agenda/agendaitems.js
+++ b/app/controllers/agenda/agendaitems.js
@@ -29,6 +29,7 @@ export default class AgendaAgendaitemsController extends Controller {
   @service intl;
   @service decisionReportGeneration;
   @service throttledLoadingService;
+  @service toaster;
 
   @lastValue('groupNotasOnGroupName') notaGroups = [];
   @tracked meeting;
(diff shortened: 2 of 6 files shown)


app/components/agenda/agendaitem/decision/digital.js
```diff
@@ -181,6 +181,7 @@ export default class AgendaAgendaitemDecisionDigitalComponent extends Component
   loadReport = task(async () => {
     this.report = await this.args.decisionActivity.belongsTo('report').reload();
     if (this.report) {
+      await this.report.belongsTo('file').reload();
```
ValenberghsSven, line 184: scenario where user 1 and user 2 both changed decision reports (also load them in ember store before)
and then switch to decision view
There was an error on `report.file.derived` since we do `file.belongsTo('derived').reload()` in the document-card. but the file was changed compared to last viewed.
Even though a report never has a derived (since source is PDF)...
just ember things with concurrency

app/utils/agendaitem-utils.js
```diff
@@ -195,6 +195,7 @@ export const reorderAgendaitemsOnAgenda = async(agenda, store, decisionReportGen
   const actualAnnouncements = [];
   const sortedAgendaitems = agendaitems.slice().sort((a1, a2) => a1.number - a2.number)
   for (const agendaitem of sortedAgendaitems) {
+    await agendaitem.belongsTo('type').reload(); // type could have changed concurrently
```
ValenberghsSven, line 198: scenario where user 1 is removing/sending back an agendaitem after user 2 changed a type of an agendaitem.

Without this:
user 1 and 2 start from agenda view
user 2 changes agendaitem 2 to announcement
user 1 removes/sends an agendaitem back.
During the process for user 1, we check every single agendaitem and assign a new number based in the index.
But since the former nota is still in the nota list, the numbering of the notas is wrong

There are more locations where we are just relying on type being the same.
Will see if there can be side effects in those locations

app/controllers/agenda/agendaitems.js
```diff
@@ -97,7 +98,25 @@ export default class AgendaAgendaitemsController extends Controller {
...
+        this.intl.t('changes-could-not-be-saved-title'),
+        {
+          timeOut: 60000,
+        }
+      );
+      // we don't want the changes anymore, but we are not going to refresh the page
+      this.notasHasChanged = false;
+      this.announcementsHasChanged = false;
+      return;
+    }
```
ValenberghsSven, line 119: high risk scenario:
User 1 and 2 have the same agenda loaded.
User 1 accepts a submission (f.e. should become agendaitem number 2)
User 2 changes the agendaitem positions without refreshing (swap positions of "stale" agendaitem 2 and 3 (in same mandatee block))
There will now be 2 agendaitems with number 2. and number 4 is missing.
Similar cases in case of deletes/send back to submitter
This prevents such issues, not allowing based the amount of items from ask query vs what we have in local stores.
Not foolproof (1 addition and 1 deletion will allow editing) but deletions are way less frequent than additions.

tdn (approved):

PR #2340, opened 2025-06-16
KAS-4575 add a check if toast is already in the shown toasts

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4575

- we can get check the AuToast service toasts list and check if your toast is included.
https://github.com/appuniversum/ember-appuniversum/blob/master/addon/services/toaster.js

- when our toast is included we can assume it is being shown and not show more of the same alert.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -7,6 +7,9 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
   @service intl;
   @service toaster;
 
+  networkToast;
+  errorToast;
+
   // Method override from RESTAdapter base class
   // See: https://api.emberjs.com/ember-data/3.28/classes/RESTAdapter/methods/ajax?anchor=ajax
   // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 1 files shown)


app/adapters/application.js
```diff
@@ -17,7 +20,10 @@ export default class ApplicationAdapter extends JSONAPIAdapter {
       }
       return await this.retryOnError(super.ajax.bind(this), arguments); // return-await of importance to be able to catch errors
     } catch (error) {
-      this.toaster.error(
+      if (this.toaster.toasts.includes(this.errorToast)) { // reduce visible toasts
```
tdn: Javascript uses [sameValueZero](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Equality_comparisons_and_sameness#same-value-zero_equality) in the `Array.includes()` function, which could be problematic here.

It probably works because it's looking for the exact same object which is set once in the next line of code.

tdn (approved):

PR #2342, opened 2025-06-24
KAS-5063 add method to manually sync case documents

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5063

- when moving subcase we need to fix case documents manually (doing it with new deltas would trigger the service too much)
- added service call. There is no json response to worry about, so just caught any errors. Worst case is that a case doesn't have the correct documents but we have no real use for them at this time.

- [x] https://github.com/kanselarij-vlaanderen/digital-signing-service/pull/58
- [ ] https://github.com/kanselarij-vlaanderen/case-documents-sync-service/pull/6
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/522


app/services/document-service.js
```diff
@@ -337,4 +337,15 @@ export default class DocumentService extends Service {
...
+
+  // case-documents-sync service
+  async syncCaseDocuments(caseId) {
+    const response = await fetch(`/case-documents-sync/cases/${caseId}/sync`, {
+      method: 'POST',
+      headers: { Accept: 'application/vnd.api+json' },
+    });
+    if (!response.ok) {
+      this.toaster.error(`Could not sync case documents`);
+    }
```
ValenberghsSven, line 349: getJsonPayloadOrThrow ?
We recieve a message back from the service on failure

PR #2343, opened 2025-06-25
KAS-5068 add preventunload and decision report generation to agendaitem panel edit

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5068

- added report generation (only when confidential changes to true, change to non confidential will require a manual change to decision report access which will trigger regeneration)
- added prevent unload to alert user before going away (they can still confirm and leave early anyway)
- changed prevent unload to remove the alert after confirming (it kept showing during any transition)
- added a `willDestroy` since you could leave without saving, leaving models dirty.

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
+++ b/app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
@@ -17,6 +17,9 @@ export default class AgendaitemCasePanelEdit extends Component {
   @service pieceAccessLevelService;
   @service agendaitemAndSubcasePropertiesSync;
   @service currentSession;
+  @service decisionReportGeneration;
+  @service store;
+  @service preventUnload;
 
   @tracked filter = Object.freeze({
     type: 'subcase-name',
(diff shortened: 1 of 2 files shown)


app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
```diff
@@ -154,4 +157,25 @@ export default class AgendaitemCasePanelEdit extends Component {
...
+      this.newsItem.rollbackAttributes();
+    }
+    if (this.internalReview?.hasDirtyAttributes) {
+      this.internalReview.rollbackAttributes();
+    }
+  }
+
+  willDestroy() {
+    super.willDestroy(...arguments);
+    this.rollbackDirtyAttributes();
```
tdn, line 179: I think we need a `this.preventUnload.disable();` here.

Right now the following doesn't work as intended:
1. Open an agendaitem and edit the subcase properties
2. Click on a different agendaitem -> no warning
3. Try to navigate to a different route -> warning

tdn (changes requested):
prevent unload doesn't work as intended (see comments)

app/components/agenda/agendaitem/agendaitem-case-panel/agendaitem-case-panel-edit.js
tdn: Resolved by making prevent-unload listen to parameter changes in the same route

tdn (approved):

PR #2344, opened 2025-06-26
Feature/kas 3044 agenda print view

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-3044

- added comments to agenda print view
- added new agendaitem marking to agenda print view
- added new document marking to agenda print view

note:
- some of the above was happening to agenda check views (by using that same components) so made sure it was fully implemented there aswell (so it wasn't halfly implemented as a side effect)
- when looking at designagenda A > no markings of "new" since everything is new

### DIFF
--- a/app/components/agenda/agenda-header/agenda-check.js
+++ b/app/components/agenda/agenda-header/agenda-check.js
@@ -14,6 +14,7 @@ import { inject as service } from '@ember/service';
 export default class AgendaHeaderAgendaCheck extends Component {
   @service toaster;
   @service intl;
+  @service agendaService;
 
   getAgendaitems = task(async () => {
     const notas = [];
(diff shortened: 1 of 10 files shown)


app/components/agenda/agenda-header/agenda-check.js
```diff
@@ -69,6 +81,28 @@ export default class AgendaHeaderAgendaCheck extends Component {
     return null;
   }
 
+  getNewPieces = task(async () => {
+    const agendaitems = await this.args.agenda.agendaitems;
+    const previousAgenda = await this.args.agenda.previousVersion;
+    const pieces = [];
+    const agendaitemNewPieces = agendaitems.map(async (agendaitem) => {
```
tdn, line 88: Was concerned whether this would be too many network calls when loading a big agenda, but I tested it on acceptance with 100+ agendaitems, and all in all it was doable. Since we perform these same calls when loading the agenda overview, it seems they are still cached by the time we load the print view or agenda-check view. 👍

tdn (approved):

PR #2345, opened 2025-07-01
KAS-5070 implement json parsing via util

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5070

- added util to parse json with all possible scenarios and outcomes (services also throw errors in different ways)
- added some error handling where it seems needed
- replace all `.json()` processing except for 1 case (where we check the response is json first)


tdn (approved):
All seems to work fine. To be retested on acceptance to be extra sure

PR #2348, opened 2025-07-03
add env property to disable session polling and add to cypress-test env

ValenberghsSven:
No ticket, cypress is less stable lately and often on sessions

Checking if disabling session polling helps


ValenberghsSven:
used as hotfix

PR #2353, opened 2025-07-17
Feature/cypress fixes

ValenberghsSven:
no ticket, trying to fix some flakyness:

- decisionActivity doesn't always have result code. needs optional param "?.uri"
- editing decision result on pill improved, only set on model when saving (also prevents unsaved model changes when not cancelling)
- subcase view might not be updated if decision result changed, added reload like we did on other locations

didn't fix:
- newsitem zebra view showing 1 item when there should be 2. I think this is an issue with sorting on dates where there are undefined dates. The best way might be to stop relying on sorting via query and sort in frontend (so we only have to load the list of agendaitems once, and keep sorting after without more queries but we need to implement in properly to support all sorting combinations
- sign flow tests, the document to sign should be there but isn't. (it is there beginning of the test, we try to start SH flow by forcing through "no email" blocking, then we expect the document to come back in the view (automated processes reset it) but it doesn't always show back up.
- after each issues. Sometimes when logging out the adapters are still attempting to load data. That causes the after each to fail and tests are skipped.

### DIFF
--- a/app/components/agenda/agendaitem/agendaitem-decision-edit.js
+++ b/app/components/agenda/agendaitem/agendaitem-decision-edit.js
@@ -4,16 +4,27 @@ import { tracked } from '@glimmer/tracking';
 import { task } from 'ember-concurrency';
 
 export default class AgendaitemDecisionEditComponent extends Component {
-  @tracked decisionActivity = this.args.decisionActivity;
+  @tracked decisionResultCode;
+
+  constructor() {
+    super(...arguments);
+    this.loadDecisionActivityResult.perform();
+  }
+
+  @task
+  *loadDecisionActivityResult() {
+    this.decisionResultCode = yield this.args.decisionActivity?.decisionResultCode;
+  }
 
   @action
   changeDecisionResultCode(resultCode) {
-    this.decisionActivity.set('decisionResultCode', resultCode);
+    this.decisionResultCode = resultCode;
   }
 
   @task
   *saveDecisionActivity() {
-    yield this.decisionActivity.save();
+    this.args.decisionActivity.decisionResultCode = this.decisionResultCode;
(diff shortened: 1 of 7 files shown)


app/components/agenda/agendaitem/agendaitem-decision-edit.js
```diff
@@ -4,27 +4,36 @@ import { tracked } from '@glimmer/tracking';
...
+  @tracked decisionResultCode;
+
+  constructor() {
+    super(...arguments);
+    this.loadDecisionActivityResult.perform();
+  }
+
+  @task
+  *loadDecisionActivityResult() {
+    this.decisionResultCode = yield this.args.decisionActivity?.decisionResultCode;
```
ValenberghsSven, line 16: `decisionActivity` shouldn't be optional here, since the save would fail if it is. I copied this from somewhere.

app/components/subcase/description-panel/edit.js
```diff
@@ -310,7 +310,7 @@ export default class SubcaseDescriptionEdit extends Component {
     } else if (
       this.agendaItemType.uri === CONSTANTS.AGENDA_ITEM_TYPES.ANNOUNCEMENT
       && (!oldDecisionResultCode
-        || oldDecisionResultCode.uri === CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD)
+        || oldDecisionResultCode?.uri === CONSTANTS.DECISION_RESULT_CODE_URIS.GOEDGEKEURD)
```
tdn, line 313: fine as is, but strictly speaking the optional `oldDecisionResultCode?` is not necessary, since `!oldDecisionResultCode` would be truthy and the OR statement would already return `true`.

tdn (approved):

PR #2356, opened 2025-08-12
KAS-5084 cancel task that yields incorrect data when exiting controller

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5084

what happened:
- we load the **agendaitems** data in the `agenda.agendaitems` **route**
- we start a **task** in the **controller** to calculate the group names in a **promise** (`groupNotasOnGroupName`) and use the `lastValue` with the current **model**
- before that promise is resolved (it takes a while) we exit the controller
- the controller is still running that task (even though we are no longer on that route)
- when opening a different agenda with 0 notas, we start a new task instance which resolves instantly
- however the other task that was still running also returns a result and that one is used instead.

How to be sure: test with slow network.
Open agenda with some notas, wait for the "0 van X agendaitems geladen" shows up, then go to agendas and open the agenda without notas.
If you add a task modifier `drop: true` you can see this is what is happening since it reproduces the issue pretty much 100%

Fix:
- cancelling the task on exit of the `agenda.agendaitems` route seems to work.
- I also cancelled loading the documents of the agendaitems, thought it might reduce loading times on switching
- since they are performed as promises, had to catch the errors thrown by the promise.
This fix alone seemed to be enough after testing a bunch

but thinking about it, there could be an alternative fix which I might want to add as a failsafe:
- limiting the task to max 1 instance and making that 1 task restartable (it should mean we drop the running task and start a new one with the new model everytime we transition into `agenda.agendaitems` route)

### DIFF
--- a/app/routes/agenda/print.js
+++ b/app/routes/agenda/print.js
@@ -69,7 +69,7 @@ export default class AgendaPrintRoute extends Route {
   *loadDocuments(agendaitems) {
     yield all(
       agendaitems.map(async (agendaitem) => {
-        await this.throttledLoadingService.loadPieces.linked().perform(agendaitem);
+        await this.throttledLoadingService.loadPieces.perform(agendaitem);
       })
     );
   }
--- a/app/controllers/agenda/agendaitems.js
+++ b/app/controllers/agenda/agendaitems.js
@@ -129,7 +129,7 @@ export default class AgendaAgendaitemsController extends Controller {
     this.router.refresh('agenda.agendaitems');
   }
 
-  @task
+  @task({ maxConcurrency: 1, restartable: true })
   *groupNotasOnGroupName() {
     const agendaitemsArray = this.model.notas.slice();
     const agendaitemGroups = [];
--- a/app/routes/agenda/agendaitems.js
+++ b/app/routes/agenda/agendaitems.js
@@ -23,6 +23,7 @@ export default class AgendaAgendaitemsRoute extends Route {
 
   @service agendaService;
   @service store;
+  @service throttledLoadingService;
 
   async model(params) {
     const {
@@ -137,6 +138,7 @@ export default class AgendaAgendaitemsRoute extends Route {
       // when rapidly switching to an agenda with 0 notas
       controller.groupNotasOnGroupName.cancelAll();
       controller.loadDocuments.cancelAll();
+      this.throttledLoadingService.loadPieces.cancelAll(); // loadDocuments task started a lot of these
     }
   }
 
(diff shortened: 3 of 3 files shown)


ValenberghsSven:
tested it on ACC with bigger agendas. could not reproduce the same issued after applying the fix
also there might be a performance increase (since we are no longer calculating a bunch of data we aren't going to show) but hard to measure it.

edit:
Checked the network tab. we are still loading pieces after exiting the route;
According to the ember-concurrency "If the Parent task is canceled, the Child task will automatically be canceled as well"
But that doesn't seem to happen? ~~But it could be because multiple places call the same task in `ThrottledLoadingService`.
So I think the tasks started from `agenda.agendaitems` controller are getting cancelled but not the ones started in `agenda-overview-item`~~

So what if I just add `this.throttledLoadingService.loadPieces.cancelAll();` in `agenda.agendaitems` controller? It works, but is not very clean since we could introduce side-effects here. We have no idea if any of those tasks need to be loaded.
~~So going to check if I can cancel the task from agenda-overview-item~~
The pieces are from the controller, the overview items only load when the components are shown

So the child tasks are not cancelling contrary to the docs.
https://v221.ember-concurrency.com/docs/child-tasks
Maybe because the child task is not in the local scope but in a service? in any case, cancelling all `loadPieces` tasks on exiting the controller seems like the best course, I don't think there would be side-effects.
3 places use the service:
- agenda.agendaitems (overview/detail of agenda main view)
- agenda-overview-item component (inside said view)
- agenda.print (same agenda, different sibling view to the above view)

Going to also add cancelling when exiting the printview, since that data would be used by nobody on exiting. The only possible benefit would be loading the cache, but we do that anyway when actually loading the data.

app/routes/agenda/print.js
```diff
@@ -69,7 +69,7 @@ export default class AgendaPrintRoute extends Route {
   *loadDocuments(agendaitems) {
     yield all(
       agendaitems.map(async (agendaitem) => {
-        await this.throttledLoadingService.loadPieces.linked().perform(agendaitem);
+        await this.throttledLoadingService.loadPieces.perform(agendaitem);
```
ValenberghsSven, line 72: this gave a warning since we are not using the result of a linked task (which is not supported).

ValenberghsSven:
I was playing around a bit in the `agenda.print` route.
I tried exiting early but we never hit the `resetController` for some reason when I do that so I couldn't cancel the pieces being loaded.
So just stopped for now. Enough time wasted on side tracking.
The current fix will remove the issue and improve performance. which is more than was asked.

tdn (approved):
works well locally

PR #2358, opened 2025-08-14
KAS-5072 refactor draft-piece edit-modal

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5072

implemented replacing logic as follows:
- refactored to just have 1 upload modal
- both old file and derived file get destroyed only if we have a replacement
- try to convert the new file to pdf

Screencaptures
- opening modal
image

- "replace" clicked
image

- file uploaded (limited type) showing alert
image

note:
- since this modal is used for draft-piece regardless of profile, admins/editors are also restricted
- no more option to manually add a PDF if automatic conversion fails (too large). this will have to be done after accepting submission (in the actual piece edit-modal)

But there is still the possibility to attempt a new conversion.
image

For this PR I kept a lot of the visual style of the existing model (using a "replace" button that opens a modal)
I also have a second suggestion for the view, where name editing is a separate card and the replacement upload modal is in it's own card.
I will branch of this PR and make that to compare.

- [ ] compare with https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2359

### DIFF
--- a/app/components/auk/file-uploader.hbs
+++ b/app/components/auk/file-uploader.hbs
@@ -38,7 +38,7 @@
           {{on "click" this.clickInput}}
           {{on "key-up" this.clickInputOnEnter}}
         >
-          {{t "upload-files" numFiles=(if @multiple "multiple" "one")}}
+          {{t "upload-files" numFiles=(if @multiple "multiple" "1")}} 
         </AuButton>
       {{else if this.uploadIsRunning}}
         {{t "uploading-files" numFiles=this.fileQueue.files.length}}
@@ -59,7 +59,7 @@
         {{#if dropzone.active}}
           {{t "drop-text"}}
         {{else if dropzone.supported}}
-          {{t "drag-or-click-for-upload" numFiles=(if @multiple "many" "one")}}
+          {{t "drag-or-click-for-upload" numFiles=(if @multiple "many" "1")}}
         {{/if}}
       </p>
     {{/if}}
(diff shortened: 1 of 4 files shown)


app/components/auk/file-uploader.hbs
```diff
@@ -38,7 +38,7 @@
           {{on "click" this.clickInput}}
           {{on "key-up" this.clickInputOnEnter}}
         >
-          {{t "upload-files" numFiles=(if @multiple "multiple" "one")}}
+          {{t "upload-files" numFiles=(if @multiple "multiple" "1")}}
```
ValenberghsSven, line 41: this used to work (difference between "bestand opladen", "bestanden opladen") but was no longer working for some reason.

app/components/documents/draft-document-card/edit-modal.js
```diff
@@ -102,104 +63,47 @@ export default class DocumentsDraftDocumentCardEditModalComponent extends Compon
...
     await this.args.documentContainer.save();
 
-    if (this.replacementSourceFile) {
-      if (this.args.piece.stamp) {
-        await this.documentService.stampDocuments([this.args.piece]);
-      }
-    } else {
-      await this.documentService.checkAndRestamp([this.args.piece]);
-    }
-
```
ValenberghsSven, line 186: draft-pieces don't even a have stamp property, so this should never do anything

tdn (approved):
We need this PR for the logic in https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2359

PR #2359, opened 2025-08-14
Feature/kas 5072 cabinet replace documents v2

ValenberghsSven:
CFR version 1:
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2358

https://kanselarij.atlassian.net/browse/KAS-5072
Starting from version 1:
- removed the toggle so the upload modal is always showing
- alert is still only shown if a file was actually uploaded (only when the previous file is about to be deleted)

Opening modal
image

Uploading a file
image


tdn (approved):
This 2nd version is nice and clean, saves the user an extra click, and works well. Much better than the current modal!

PR #2361, opened 2025-08-25
KAS-5082 change relation from belongsTo to hasMany

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5080

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/528


tdn (approved):
After handling an edge case this works fine

PR #2362, opened 2025-09-11
KAS-4425 add and implement automation to ensure signedPiece(Copy) has correct accessLevel

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4425

Not a lot of code, tried to keep it simple and clear.
Was wondering if it would have been better to do this with deltas, but that would require sending a lot of deltas where we only have to do the work a few times a day since accessLevels are not updated that frequently
I also expect some concurrency issues where we have to implement `piece.belongsTo('accessLevel').reload();` for `SignedPieces` and `SignedPieceCopy`.

Anyway, there are many pieces now where the signedPiece or the signedPieceCopy do not have the correct accessLevel.
Are we going to update those with migrations?
Ticket doesn't mention them, but we could list them with queries

Places where we save pieces but this logic is not needed:
- When we add a new version of a piece (new piece, no signed piece at that time but previous piece will be checked since it's in that piece of code)
- publication pieces (no accessLevel)
- draft-pieces (no signedPieces yet)
- places where we edit piece/file but not the accessLevel (like edit-modal)

### DIFF
--- a/app/components/documents/batch-details/batch-documents-details-modal.js
+++ b/app/components/documents/batch-details/batch-documents-details-modal.js
@@ -208,6 +208,7 @@ export default class BatchDocumentsDetailsModal extends Component {
           ) {
             // - "draft-pieces" have "pieces" as previous. we update those in a later step.
             // - some profiles are allowed to edit "draft-piece" but not "piece" so this could error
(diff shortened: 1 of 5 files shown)


tdn (approved):
pair review

PR #2364, opened 2025-09-16
KAS-4759 add check on email for reports and ratification signatures

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4759


ValenberghsSven:
tests fail on this (as they should, behaviour changed)
still need to adapt the tests to ignore the new safety mechanism (since we are faking the response anyway)

tdn (approved):

ValenberghsSven:
ah this PR seems aware of the tests from the subbranch ?
in any case, tests succeeded twice in subbranch without any changes, going to merge

image

PR #2368, opened 2025-10-06
Feature/KAS-5061 newsitems being edited by

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5061

- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/537

TODO test a bit more.
changed 1 "actions up" to process the save/cancel deeper in the tree since I needed to save normally or save + remove the user.
In print view, the `preventUnload` doesn't work when switching between "klad" en "definitief" since we are staying in the same route but change the params instead
in zebra view, opening more than 1 edit panel and closing 1 means the other opened panels now have `preventUnload` disabled. Not sure how to counter that

Yggdrasil, We WILL be propagating this relation if not ignored/removed. (only when the relation exists after releasing decisions)
Perhaps hide the alert behind a permission instead of changing yggdrasil?  --- DONE

### DIFF
--- a/app/components/news-item/edit-panel.hbs
+++ b/app/components/news-item/edit-panel.hbs
@@ -166,7 +166,7 @@
               <AuButton
                 data-test-news-item-edit-item-cancel
                 @skin="naked"
-                {{on "click" @onCancel}}
+                {{on "click" this.cancel}}
               >
                 {{t "cancel"}}
               </AuButton>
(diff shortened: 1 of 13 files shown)


tdn (approved):
just some conflicts in translations to be resolved. Otherwise looks good in pair review

PR #2370, opened 2025-10-08
remove SESSION_COOKIE_SECURE in CI override

ValenberghsSven:
Don't merge yet, going to rollback the changes done on project one at a time, checking what broke jenkins runs


ValenberghsSven:
going to close this, pushed it to acceptance branch instead (since it seems to fix the issue and acceptance also needs the fix)

PR #2371, opened 2025-10-08
KAS-5062 add filter to hide intermediate versions and ignore retracted/postponed

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5062

- add filter on "has-no" next piece. only the most recent version is shown in the list (tried locally where I saw 2 before fix, 1 after fix)
- add post query filtering of notas where agenditem is "retracted" or "postponed".
only notas are shown when the decision is "approved", "acknowledged" or undefined. (but agendaitem type has to be a nota)

Might want to test this on ACC on its own on large agendas (`decisionActivity` is not optional here since we shouldn't have a nota on an agendaitem without it except for maybe legacy, but this screen is not designed for legacy anyway...)


tdn (approved):
pair review

PR #2373, opened 2025-10-13
KAS-5060 add possibility to send only announcements, add more feedback to user

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5060

- [x] https://github.com/kanselarij-vlaanderen/newsletter-service/pull/56
- [x] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/540

- added a second check to verify if there are any announcements with a valid theme (for the agendas without nota)
- Since it was now possible to accidentally send announcements only I added some extra checks on opening the confirm windows
- show alert when no agendaitems or announcements have themes (mailchimp will fail, Belga and themis will not)
- show alert when there are only announcements with a theme (all will succeed, but could happen by mistake. will only be shown in those rare cases where an agenda has only announcements to send OR when no nota was checked "in kort bestek")
- disable sending mailchimp if 0 items can be send (based on themes). Belga also blocked (but may be ok if there are some notas/announcements, themes are irrelevant for Belga)
- No blocker on themis, just a changed button color "Toch publiceren" (with the above alert)

the alerts may be a bit much, but we barely have any feedback on these pages right now and once the mail/belga is sent it's done.


app/components/newsletter/newsletter-header-overview.js
```diff
@@ -63,6 +67,42 @@ export default class NewsletterHeaderOverviewComponent extends Component {
...
+      'filter[themes][:has:mailchimp-id]': true, // has mailchimp thema
+    });
+  }
+
+  get disableConfirm() {
+    return this.calculateTotals.isRunning ||
+    (this.notaWithThemeCount == 0 && this.announcementWithThemeCount == 0) ||
+    this.publishToAll.isRunning ||
+    this.publishToBelga.isRunning ||
+    this.publishToMail.isRunning;
```
ValenberghsSven, line 119: this is also used for Belga, but there could be newsitems that exist (without a theme) that could be sent. maybe too strict here.

app/components/newsletter/newsletter-header-overview.js
```diff
@@ -123,7 +163,7 @@ export default class NewsletterHeaderOverviewComponent extends Component {
     } catch(e) {
       console.log(e);
       this.toaster.error(
-        this.intl.t('error-send-belga'),
+        `${this.intl.t('error-send-belga')}. ${e.message}`,
```
ValenberghsSven, line 182: the backend sends a new message if we ever hit 0 newsitems to send.

tdn (approved):

PR #2376, opened 2025-10-29
Feature/kas 5097 ember v6

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5097

in progress

checking what other dependencies we can still upgrade

ember-concurrency upgrade requires all tasks to be refactored. Could do that now, but all PR's will need to be updated before or after to match it.
We could also already resolve the "inject as service" deprecation, but same issue with PR's


ValenberghsSven:
also check if we can merge the app-universum upgrade.
This will make some overrides in the package json obsolete
https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2369

ValenberghsSven:
:warning:  This branch doesn't work without cypress upgrade because of node version restriction on older cypress

edit: fixed after cypress upgrade was merged

package.json
```diff
@@ -36,129 +36,147 @@
...
+    "ember-load-initializers": "^3.0.1",
+    "ember-math-helpers": "^5.0.0",
+    "ember-modifier": "^4.2.2",
     "ember-mu-transform-helpers": "^2.1.3",
-    "ember-named-blocks-polyfill": "^0.2.5",
-    "ember-page-title": "^8.2.3",
-    "ember-pdfjs-wrapper": "git+https://github.com/MikiDi/ember-pdfjs-wrapper.git#8300c671d3293dad0c5ccf87defdec6d08f839b3",
-    "ember-plausible": "^0.1.1",
+    "ember-page-title": "^9.0.3",
+    "ember-pdfjs-wrapper": "git+https://github.com/ValenberghsSven/ember-pdfjs-wrapper.git#686f83b89e5917bb06c145bd792e280aa90bc991",
```
tdn: Let's make a fork within kanselarij-vlaanderen

tdn: https://github.com/kanselarij-vlaanderen/ember-pdfjs-wrapper/commit/686f83b89e5917bb06c145bd792e280aa90bc991

config/deprecation-workflow.js
```diff
@@ -1,15 +0,0 @@
-/* eslint-disable */
```
tdn, line 1: @ValenberghsSven check if we need this?

tdn (approved):
Preliminary approve. Still need to test some things, but I see no reason not to merge after the current release has gone to PROD

ValenberghsSven:
## :warning: 

Use https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2414 instead of this one
since that PR contains merges with ember-concurrency PR that change the resolution for `@lastValue` from a getter with `lastSuccessful.value`to a `@tracked` instead.
Manual merging of both PR 2376 (this) and then PR 2400 (concurrency) will give conflicts but can't fully resolve the differences and will result in both a getter and a tracked with the same name.

package.json
ValenberghsSven: done

PR #2381, opened 2025-11-14
KAS-5125 add warnings when newsletter has confidential items

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5125

FYI, this branch will fail on its own, it needs the backend branch of feature/KAS-5060-send-announcements-only-mail

- agenda-actions: added calculation and alert message on (re)publishing modal (note: this does not appear on the documents planning modal, where you would initially confirm the release of the documents, which I think is ok. Newsletter will already be released at that stage)
- table-row (newsletter zebra view): added confidential pill
- newsletter-header-overview: added an alert on all confirm actions (separate from the 'no newsletters have valid theme' alert) when any newsletter that is `in-newsletter` but also the subcase is `confidential`
- Had to add cache-busting since the count didn't change after updating `confidential` or `in-newsletter`

### DIFF
--- a/app/components/agenda/agenda-header/agenda-actions.hbs
+++ b/app/components/agenda/agenda-header/agenda-actions.hbs
@@ -279,6 +279,8 @@
 {{#if this.showConfirmPublishThemis}}
   <ThemisPublications::PublishConfirmationModal
     @scope={{this.themisPublicationScopes}}
+    @loading={{this.checkConfidentiality.isRunning}}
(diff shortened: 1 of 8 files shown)


ValenberghsSven:
https://jenkins.kanselarij.org/job/kaleidos/job/custom_frontend_backend_combination_2/109/ is green

tdn (approved):
pair review

PR #2386, opened 2025-11-19
upgrade to cypress v15, not really any breaking changes

ValenberghsSven:
No ticket, but hopefully resolves some issues we have with ember v6 upgrade.


ValenberghsSven:
was green once (build 3)

ValenberghsSven:
1 test is failing, unsure why, didn't have time to run it locally yet as it does need a bit of setup.
May opt to disable it and fix later (the test is mocking digital signing sending bad or successful responses)

This could be merged on its own, but it is already included in a different branch with the new ember upgrade

ValenberghsSven:
merged `feature/upgrade-cypress-with-jenkins-ci` which contained all commits

PR #2387, opened 2025-11-19
Feature/ember and cypress upgrade

ValenberghsSven:
Cypress 12.17.4 can only run on node v20.5.
In an attempt to solve the issues jenkins has I am trying an upgrade.
Cypress > 15.6.0 has little to no breaking changes for us it seems. might as well upgrade if it has no sideeffects


ValenberghsSven:
This would be the branch to merge + the concurrency deprecation branch
Test that fails is the same that fails in cypress only upgrade so not ember related

ValenberghsSven:
Since the cypress PR was merged on its own, This can get merged
It provides some fixes to tests for ember v6

PR #2388, opened 2025-11-20
first upgrade only mocklogin and acmidm

ValenberghsSven:
checking if the login issue is ember based, or mocklogin/ember-simple-auth


ValenberghsSven:
this was a test, it proved mocklogin was the issue, not ember-simple-auth

PR #2391, opened 2025-12-02
KAS-1989 add option to download only new documents from agenda

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-1989

- add option to download only "new" documents by passing a new param to backend.

- [x] https://github.com/kanselarij-vlaanderen/file-bundling-job-creation-service/pull/21
- [ ] https://github.com/kanselarij-vlaanderen/app-kaleidos/pull/548 (Don't merge, for jenkins only)

- note, this is showing on agendas where only agenda A exists. In that case both option would be the same.
Not sure if this is ok.
I guess we could only show the option if the current agenda is A and there is only 1 agenda in total.

example of the updated modal
image

### DIFF
--- a/app/components/agenda/agenda-header/agenda-actions.hbs
+++ b/app/components/agenda/agenda-header/agenda-actions.hbs
@@ -335,6 +335,25 @@
           </AuRadioGroup>
         </Fieldset.content>
       </AuFieldset>
+      <AuFieldset as |Fieldset|>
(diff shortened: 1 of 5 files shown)


tdn (approved):
pair review

ValenberghsSven:
test failed on a known flaky (newsitems zebra showing 1 item where 2 exist), running again

PR #2393, opened 2025-12-15
KAS-5121: Refactor van resterende Auk::Button & Auk::ButtonLink instanties + verwijdering componenten/styling

brenner-company:
**Tickets binnen Jira:** 
- https://kanselarij.atlassian.net/browse/KAS-5121


ValenberghsSven (approved):
I think this is fine.

On jenkins there seems to be a failing spec related to a button click in the `postpone` spec.
Going to monitor a few more runs before determining if this is a new flaky test or something related to the PR.
The same test also fails in another DES branch, where buttons are not changed.

ValenberghsSven:
postpone spec was failing, there was a use of `AuButton` with `models` and `route` which doesn't work.
It worked on `auk::button` since we implemented that but appuniversum buttons do not accept those parameters.
So needed an `on "click"` event instead after this refactor
Both in `agendaitem-retracted` and `agendaitem-postponed` component

ValenberghsSven:
jenkins is green now

PR #2399, opened 2026-02-05
KAS-4749 implement changes to when formally ok is reset

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-4749

TODO adapt tests.
Easiest start point is running the suite and starting with all the failed ones.

### DIFF
--- a/app/controllers/agenda/agendaitems/agendaitem/index.js
+++ b/app/controllers/agenda/agendaitems/agendaitem/index.js
@@ -2,7 +2,7 @@ import Controller, { inject as controller } from '@ember/controller';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
-import { reorderAgendaitemsOnAgenda, setNotYetFormallyOk } from 'frontend-kaleidos/utils/agendaitem-utils';
+import { reorderAgendaitemsOnAgenda } from 'frontend-kaleidos/utils/agendaitem-utils';
 import { isPresent } from '@ember/utils';
 import { isEnabledVlaamsParlement } from 'frontend-kaleidos/utils/feature-flag';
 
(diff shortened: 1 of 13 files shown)


app/controllers/agenda/agendaitems/agendaitem/index.js
```diff
@@ -136,7 +136,6 @@ export default class IndexAgendaitemAgendaitemsAgendaController extends Controll
     governmentAreas.length = 0;
     governmentAreas.push(...newGovernmentAreas);
     await this.subcase.save();
-    setNotYetFormallyOk(this.model);
     await this.model.save();
```
ValenberghsSven, line 139: agendaitem model save not needed

app/controllers/cases/case/subcases/subcase/index.js
```diff
@@ -79,7 +78,6 @@ export default class CasesCaseSubcasesSubcaseIndexController extends Controller
       'filter[agenda][status][:uri:]': CONSTANTS.AGENDA_STATUSSES.DESIGN,
     });
     await Promise.all(agendaitemsOnDesignAgendaToEdit.map(async (agendaitem) => {
-      setNotYetFormallyOk(agendaitem);
       return agendaitem.save();
```
ValenberghsSven, line 81: agendaitem.save no longer needed

tdn (approved):

app/controllers/agenda/agendaitems/agendaitem/index.js
ValenberghsSven: this was incorrect. we have to save agendaitem to force the concurrency.
If we don't, a save on agendaitem would be possible, but a save on subcase would error.

app/controllers/cases/case/subcases/subcase/index.js
ValenberghsSven: same, keep it

PR #2401, opened 2026-02-17
Feature/upgrade cypress with jenkins ci

ValenberghsSven:
Merged cypress upgrade with the jenkins upgrade to see if they work together.

shortlist overview fails on both, I added bigger waits on jenkins branch for that.
Checking what tests (if any) fail consistently when using both


ValenberghsSven:
image

ValenberghsSven:
flakyness exists, but we can deal with that on the go

PR #2403, opened 2026-03-02
Feature/kas 5097 kas 5171 ember plus concurrency

ValenberghsSven:
## :warning: don't merge.

this is multiple ember PR's together
+ ember-concurrency package upgrade

ember-concurrency seems to be working (prior to this, the app would fail on startup when trying to upgrade ember-concurrency if 1 task in incorrect syntax was detected (also in subdependencies like in ember-appuniversum)

issues:
`ember-power-select` is broken
Partially from our own dependency + also from ember-appununiversum (`AuFormRow` no longer works)
Both are yielding the same issue but hard to debug.
Probably related to the ember-concurrency upgrade since ember v6 branch did not have this issue. also ember-concurrency changes without package upgrade was working.

To pin this issue down I might have to try an Ember v6 project from scratch ( the superrentals demo).
Or attempt to override certain things (upgrade or downgrade) relating to ember-power-select.
We are using v7, ember-appuniversum is using v8. there is a migration guide for v7 to v8 

update: I couldn't get it to work, in the superrentals app I imported the power-select but styling seems to be missing, at a loss for a bit, might need an extra pair of eyes on it


tdn (changes requested):
to be updated with development

ValenberghsSven:
old branch, delete this without merging, make a new branch with both combined

PR #2404, opened 2026-03-06
add deprecated themes list to theme selector for legacy edits

ValenberghsSven:
No ticket yet
Made this in 15 min as a POC that editing old themes was easily changed.
Styling is in the air, I wanted it to pop out and be obvious those are old.

### DIFF
--- a/app/components/utils/themes-selector.js
+++ b/app/components/utils/themes-selector.js
@@ -12,6 +12,7 @@ export default class ThemesSelector extends Component {
   @service store;
 
   @tracked themes;
+  @tracked deprecatedThemes;
 
   constructor() {
     super(...arguments);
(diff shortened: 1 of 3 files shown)


app/components/utils/themes-selector.js
```diff
@@ -24,6 +25,12 @@ export default class ThemesSelector extends Component {
       filter: { deprecated: false },
       sort: 'label',
     });
+    if (this.args.selectedThemes?.some((theme) => theme.deprecated)) {
+      this.deprecatedThemes = yield this.store.queryAll('theme', {
+        filter: { deprecated: true },
+        sort: 'label',
+      });
+    }
```
ValenberghsSven, line 33: this will need to change to await in/after the ember-concurrency PR

tdn:
@tdn note to self: add a ticket for this

PR #2406, opened 2026-03-13
KAS-5214 fix submission inheriting old government areas

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5214

There were 2 issues going on here:

The first issue was that we added old government areas to submission from `latestSubcase`
Solved that by removing the code responsible in `app/components/cases/new-submission.js`
We don't need that piece of code. the `selectedGovernmentFields` will be populated when the selector initializes.
that populating only adds active domains and fields at the time of creation. 

Test it by:
open an older case/subcase (before 1/1/26) and make a new submission.
Even deprecated domains will get filtered out.

The second issue was after the above happened. Now we have subcases/agendaitems with old governmentAreas and the edit panel does not show any checkboxes.
Since the only way to "remove" those "hidden" fields and domains is pressing a checkbox (to change the current selection)
and then saving (the entire selection is saved) it was not possible to remove them
So we need to edit the current selection at some stage and remove the old ones.
Which I have done during construction of `app/components/utils/government-areas/edit-government-areas-modal.js`
The parent of that panel > `<Utils::GovernmentAreas::GovernmentAreasPanel` is used **almost** everywhere expect publication views AND new-submission / new-subcase-form

for the latter we use `<Cases::Subcases::GovernmentAreasPanel`
Which contains the following piece of code that will remove old fields during construction (which issue 1 relies on now)

https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/feature/KAS-5214-old-government-areas-on-submission/app/components/cases/subcases/government-areas-panel.js#L74

Todo? not sure if I checked the scenario where a deprecated domain was selected (and no fields from it) when making new submission. is this inherited or not (since we should be unable to "select" it, it shouldn't be selected. but it might still exist in the `selectedDomains` property

### DIFF
--- a/app/components/cases/new-submission.js
+++ b/app/components/cases/new-submission.js
@@ -81,14 +81,7 @@ export default class CasesNewSubmissionComponent extends Component {
       this.title = this.latestSubcase.title;
       this.subcaseName = this.latestSubcase.subcaseName;
       this.governmentAreas = await this.latestSubcase.governmentAreas;
-      for (const governmentArea of this.governmentAreas) {
-        await governmentArea.broader;
-        if (governmentArea.broader) {
-          this.selectedGovernmentFields.push(governmentArea);
-        } else {
-          this.selectedGovernmentDomains.push(governmentArea);
-        }
-      }
+      // the governmentAreas will be date filtered into selected fields and domains by the governmentAreasPanel
     }
   })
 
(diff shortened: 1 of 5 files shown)


ValenberghsSven:
This PR should have been from acceptance branch.

Made https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2407 for it with cherry picked commits

PR #2410, opened 2026-04-03
 bugfix/KAS-5143-agenda-grouping-stuck-when-tab-inactive

DavithkbY:
The groupNotasOnGroupName task uses yield animationFrame() to keep the UI responsive while iterating over agenda items. However, browsers pause requestAnimationFrame callbacks when a tab is not visible, causing the task to hang indefinitely until the user switches back to the tab.

Replaced animationFrame() with timeout(0). This still yields to the main thread each iteration to keep the UI responsive, but unlike requestAnimationFrame, setTimeout continues to fire in background tabs.

### DIFF
--- a/app/controllers/agenda/agendaitems.js
+++ b/app/controllers/agenda/agendaitems.js
@@ -3,7 +3,7 @@ import { inject as service } from '@ember/service';
 import { guidFor } from '@ember/object/internals';
 import { action, set } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
-import { task, lastValue, all, animationFrame } from 'ember-concurrency';
+import { task, lastValue, all, timeout } from 'ember-concurrency';
 import {
   setAgendaitemsNumber,
   AgendaitemGroup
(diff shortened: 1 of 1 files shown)


erikap:
I discussed this with @madnificent. The `animationFrame()` has been added in the past to prevent the browser/app freezes. The way data is loaded and prepared for display has changed a lot since then.
@DavithkbY Can you give it a try without the `yield timeout(0)`. We assume this is not necessary since there is another `yield` right after it on the next line.

DavithkbY:
> I discussed this with @madnificent. The `animationFrame()` has been added in the past to prevent the browser/app freezes. The way data is loaded and prepared for display has changed a lot since then. @DavithkbY Can you give it a try without the `yield timeout(0)`. We assume this is not necessary since there is another `yield` right after it on the next line.

I looked online and some people said to use timeout(0) to prevent browser freezes. I've added it just to be sure.
I can't test this locally because I don't have agendas that big that they need a long time to load. I need to test this on the test environment before I can conclude if the timeout is not needed.

tdn:
Replaced by https://github.com/kanselarij-vlaanderen/frontend-kaleidos/pull/2412 for woodpecker build

PR #2411, opened 2026-04-03
KAS-5131 add new save override where only accessLevelModified is set

ValenberghsSven:
https://kanselarij.atlassian.net/browse/KAS-5131

- Added save override only for accessLevel changes to `piece` without modified date, fallback to normal save if any attribute was changed
- Changed original save override to only change accessLevel if specified
- Added more clarity to update-notas tab by added specific dates of certain actions (more lines for specific changes so it's easier to decide if the nota needs an update)
- Added the save override to `draft-piece` in case it's called on this model (f.e. by bulk editing).
- Changed Bulk edit to choose either save option depending on type of changes.
- Changed `AccessLevelPill` "on save" implementations to use new save override
- Changed `PiecesAccessLevelService` calls to use new save overrides


tdn (approved):
Pair review

PR #2413, opened 2026-05-05
use id to check instead of object, use getter for plannedStart

ValenberghsSven:
fix using object compare was not going to work
fix plannedStart seemed undefined when debugging on ACC. figured I tried with `.get()` and seems to work


ValenberghsSven:
this was only for history, no review needed

PR #2416, opened 2026-05-28
Prevent duplicate organizations from being able to be created.

DavithkbY:
Prevent duplicate organizations being created by showing warning message to user.
Also add a dropdown for existing contactpersons, filtered by organization.
Redesign the flow a little bit:

image
image
image


erikap:
This one requires the Cypress tests to be updated as well @DavithkbY

DavithkbY:
> This one requires the Cypress tests to be updated as well @DavithkbY

The tests are fixed now.

DavithkbY:
Custom pipeline works: 
https://ci.kanselarij.org/job/Kaleidos/job/custom-combo/32/

app/components/publications/publication/case/contact-persons/contact-persons-panel.js
```diff
@@ -42,35 +42,40 @@ export default class PublicationsPublicationCaseContactPersonsPanelComponent ext
...
-    await contactPerson.destroyRecord();
-    await person.destroyRecord();
+  async unlink(contactPerson) {
+    const publicationFlow = this.args.publicationFlow;
+    const publicationFlows = await contactPerson.publicationFlows;
+    const idx = publicationFlows.indexOf(publicationFlow);
+    if (idx >= 0) {
+      publicationFlows.splice(idx, 1);
+    }
+    await contactPerson.save();
```
erikap: I don't think this approach will work reliably.

When fetching relationships through Ember Data, only the first page is returned (silently). In mu-cl-resources, the default page size is 20. This means that if a contact person has more than 20 publication flows, you'll only get a subset of the relationship.

If you then modify that (spliced) subset and save it back, Ember Data will overwrite the entire relationship. As a result, all publication flows that were not part of the first page will lose their link to the contact person.

Many-to-many relationships are somewhat tricky to work with in Ember Data. Whenever you save changes to such a relationship, you need to be certain that you have the complete set of related records, not just a subset.

In this case, I suspect it would be easier to persist changes from the publication-flow side, since a publication flow will likely only have a limited number of contact persons. In the opposite direction, a contact person can potentially be linked to a large number of publication flows.

To ensure you have the complete relationship, you have two options:
- Use our store extension `this.store.queryAll('contact-person')` with a `filter[publication-flows][:id:]=....`
- Include the relationship in the query. Included relationships are not paginated, so you're guaranteed to have the full set. For example:
`this.store.queryOne('publication-flow', {  include: 'contact-persons' });`

app/components/publications/publication/case/contact-persons/contact-persons-panel.js
```diff
@@ -42,35 +42,40 @@ export default class PublicationsPublicationCaseContactPersonsPanelComponent ext
...
     const publicationFlow = this.args.publicationFlow;
-    const contactPerson = this.store.createRecord('contact-person', {
-      email: email,
-      person: person,
-      publicationFlow: publicationFlow,
-    });
+    let contactPerson = contactPersonProperties.contactPerson;
+    if (contactPerson) {
+      const publicationFlows = await contactPerson.publicationFlows;
+      publicationFlows.push(publicationFlow);
```
erikap: See the other comment about many-to-many relationships in Ember Data. It's a similar problem here.

app/components/publications/publication/case/contact-persons/contact-person-add-modal.hbs
```diff
@@ -89,6 +41,112 @@
             </AuButton>
           </div>
         </AuFormRow>
+
+        {{#if this.hasOrganizationSelection}}
+          {{#unless this.isCreatingNewPerson}}
```
erikap: More readable to work with if/else since you have both cases?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -7,60 +7,124 @@ import { timeout, task } from 'ember-concurrency';
...
   @tracked isOpenOrganizationAddModal = false;
 
   @tracked organizations;
+  @tracked persons = [];
 
   @tracked firstName;
   @tracked lastName;
   @tracked email;
-  @tracked organization;
+  @tracked organizationSelection;
```
erikap: Try to use consistent naming, at least inside a single component. Use `selectedOrganization` to be in line with `selectedPerson`.

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -7,60 +7,124 @@ import { timeout, task } from 'ember-concurrency';
...
+
+  NO_ORGANIZATION = NO_ORGANIZATION;
+
+  get organization() {
+    return this.organizationSelection?.isNoOrganization ? null : this.organizationSelection;
+  }
+
+  get hasOrganizationSelection() {
+    return isPresent(this.organizationSelection);
+  }
```
erikap: Put getters below the constructor

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -90,9 +169,61 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
     }
     const organizations = await this.store.query('organization', {
       ...query,
-      'page[size]': PAGE_SIZE.SELECT,
+      'page[size]': 40,
```
erikap: Committed by mistake?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -90,9 +169,61 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
+    }
+    return [NO_ORGANIZATION, ...organizations.slice()];
+  }
+
+  async loadPersons(searchTerm) {
+    if (!this.organizationSelection) {
+      return [];
+    }
+    const baseQuery = {
+      'page[size]': 40,
```
erikap: Use page size imported from config

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -90,9 +169,61 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
+        this.store.query('person', { ...baseQuery, 'filter[first-name]': searchTerm }),
+      ]);
+      const seen = new Set();
+      persons = [];
+      for (const p of [...byLast.slice(), ...byFirst.slice()]) {
+        if (!seen.has(p.id)) {
+          seen.add(p.id);
+          persons.push(p);
+        }
+      }
```
erikap: Can't you use the costruction with `:or:` to search in 1 query?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -90,9 +169,61 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
     });
-    return organizations;
+    if (searchTerm) {
+      return organizations;
+    }
+    return [NO_ORGANIZATION, ...organizations.slice()];
+  }
+
+  async loadPersons(searchTerm) {
+    if (!this.organizationSelection) {
```
erikap: Use the get function `hasOrganizationSelection` you defined?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -90,9 +169,61 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
+        linkedNameKeys.add(personNameKey(person));
+      }
+    }
+    return persons.slice().filter((person) => {
+      const contactPersonId = person.belongsTo('contactPerson').id();
+      if (!contactPersonId) return false;
+      if (linkedContactPersonIds.has(contactPersonId)) return false;
+      if (linkedNameKeys.has(personNameKey(person))) return false;
+      return true;
+    });
```
erikap: Not sure what the intention of this code block is. I understand you're trying to filter out persons that are already linked to the publication flow? Can't you just add that as a filter in the request to the backend?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.hbs
```diff
@@ -89,6 +41,112 @@
...
+                @width="block"
+                {{on "input" (pick "target.value" (set this "email"))}}
+              />
+            </AuFormRow>
+            <AuButton
+              @skin="link"
+              @icon="cross"
+              {{on "click" this.cancelCreatingNewPerson}}
+            >
+              {{t "cancel"}}
```
erikap: This button looks a bit weird, since the modal also has a 'Annuleren' button. Maybe show something like '<-- Bestaande persoon selecteren' or open a new modal like the create organization modal

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -74,8 +138,23 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
+      organization.rollbackAttributes();
+      this.toaster.error(
+        this.intl.t('organization-with-same-name-exists'),
+        this.intl.t('warning-title')
+      );
+      return;
+    }
+    await organization.save();
+    this.setOrganization(organization);
     this.isOpenOrganizationAddModal = false;
```
erikap: Using an `else` block would avoid the preliminary `return;` and make the code more logical readable.

erikap (changes requested):

app/components/publications/publication/case/contact-persons/contact-person-add-modal.hbs
```diff
@@ -89,6 +41,107 @@
             </AuButton>
           </div>
         </AuFormRow>
+        {{#if this.isCreatingNewPerson}}
```
erikap: I know I suggested the option of '<-- Bestaande contactpersoon selecteren' before, but the UI/modal feels weird with this nested contact person form. I would be consistent with the creation of a new organization and open a new modal on top. 

It's best to show the selected organization as read-only in the create new contact person modal such that the user is aware for which organization he's creating a contact person.

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -74,9 +133,24 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
   @action
-  addOrganization(organization) {
-    this.organization = organization;
-    this.isOpenOrganizationAddModal = false;
+  async addOrganization(organization) {
+    const name = organization.name;
+    const existingOrganizations = await this.loadOrganizations(name);
+    const normalized = name?.trim().toLowerCase();
+    const duplicate = existingOrganizations.find(
+      (o) => o.id !== organization.id && o.name?.trim().toLowerCase() === normalized
```
erikap: Since organization is an unsaved Ember Data record it will never have an `id`. The first check is therefore redundant.

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -74,9 +133,24 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
   }
 
   @action
-  addOrganization(organization) {
-    this.organization = organization;
-    this.isOpenOrganizationAddModal = false;
+  async addOrganization(organization) {
+    const name = organization.name;
+    const existingOrganizations = await this.loadOrganizations(name);
```
erikap: You cannot reuse `this.loadOrganizations` here since it returns paginated search results. You should do a `this.store.queryAll()` with a filter on the name.

Maybe it's more clean to move the entire duplication checking to the organization-add-modal component?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
```diff
@@ -88,11 +162,54 @@ export default class PublicationsPublicationCaseContactPersonAddModalComponent e
...
+      const person = await cp.person;
+      if (person) {
+        linkedNameKeys.add(personNameKey(person));
+      }
+    }
+    return persons.filter((person) => {
+      const contactPersonId = person.belongsTo('contactPerson').id();
+      if (linkedContactPersonIds.has(contactPersonId)) return false;
+      if (linkedNameKeys.has(personNameKey(person))) return false;
+      return true;
```
erikap: Question still holds what you're trying to achieve here? Can't it be expressed in the query to the backend?

Some bad smells:
- `this.args.publicationFlow.contactPersons` will not give you the full list of contact persons if it doesn't fit on one page. You probably need them all.
- `person.belongsTo('contactPerson').id()`: avoid fetching async relationships this way

erikap (changes requested):

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
DavithkbY: We're filtering out contact persons already linked to the publication.
Since it's a many to many relation, I don't think there is a way to filter it in the backend query, or is there?

app/components/publications/publication/case/contact-persons/contact-person-add-modal.hbs
DavithkbY: I moved it to a new modal now.

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
DavithkbY: Removed the id check

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
DavithkbY: Fixed and I also moved the duplication check to the org add modal

app/components/publications/publication/case/contact-persons/contact-person-add-modal.js
DavithkbY: I did fix the other bad smells in the meanwhile.

DavithkbY:
Custom pipeline running:
https://ci.kanselarij.org/job/Kaleidos/job/custom-combo/lastBuild/

app/components/publications/publication/case/contact-persons/contact-person-add-modal.hbs
```diff
@@ -7,75 +7,27 @@
...
-            value={{this.email}}
-            @width="block"
-            {{on "input" (pick "target.value" (set this "email"))}}
-          />
-        </AuFormRow>
         <AuFormRow>
           <AuLabel for="organizationInput">
             {{t "organization"}}
           </AuLabel>
           <div data-test-contact-person-select-organization class="auk-o-flex">
```
erikap: Can you increase the width of the dropdown fields in this form such that they span the entire width of the modal (like the input fields in the organization/person modal)

erikap (changes requested):
When adding or removing a contact person from a publication flow in the contact-persons-panel, the list is not updated. It requires a refresh before the result is visible.

PR #2417, opened 2026-06-11
Use trimtext to trim newlines/empty whitespaces before sending a title to the backend

DavithkbY:
Use trimtext to trim newlines/empty whitespaces before sending a title to the backend


erikap (changes requested):
There are multiple places in the code base where a  `case`  record is created. 
We should also add trimming when the title is edited after creation.

erikap (changes requested):
Also trim on places where title can be edited.

erikap (approved):

PR #2419, opened 2026-06-22
Fix "New" badge on agenda check by loading compared pieces with findRecord insteaf of queryRecord

DavithkbY:
queryRecord expected an object as the second argument, but in some cases it was being handed a string.
I've added a check where we check if the second argument is a string, if it is we will use findRecord.
If not, we will use queryRecord.

This fixes the bug where new items weren't being labeled with 'new'. 
https://kanselarij.atlassian.net/browse/KAS-5235

### DIFF
--- a/app/services/agenda-service.js
+++ b/app/services/agenda-service.js
@@ -53,10 +53,7 @@ export default class AgendaService extends Service {
         item.id
       );
       if (!itemFromStore) {
-        const modelName = singularize(item.type);
-        itemFromStore = typeof item.id === 'string'
-          ? await this.store.findRecord(modelName, item.id)
-          : await this.store.queryRecord(modelName, item.id);
+        itemFromStore = await this.store.findRecord(singularize(item.type), item.id);
       }
       itemsFromStore.push(itemFromStore);
     }
@@ -77,10 +74,7 @@ export default class AgendaService extends Service {
         item.id
       );
       if (!itemFromStore) {
-        const modelName = singularize(item.type);
-        itemFromStore = typeof item.id === 'string'
-          ? await this.store.findRecord(modelName, item.id)
-          : await this.store.queryRecord(modelName, item.id);
+        itemFromStore = await this.store.findRecord(singularize(item.type), item.id);
       }
       itemsFromStore.push(itemFromStore);
     }
@@ -101,10 +95,7 @@ export default class AgendaService extends Service {
         piece.id
       );
       if (!pieceFromStore) {
-        const modelName = singularize(piece.type);
-        pieceFromStore = typeof piece.id === 'string'
-          ? await this.store.findRecord(modelName, piece.id)
-          : await this.store.queryRecord(modelName, piece.id);
+        pieceFromStore = await this.store.findRecord(singularize(piece.type), piece.id);
       }
       piecesFromStore.push(pieceFromStore);
     }
--- a/app/services/agenda-service.js
+++ b/app/services/agenda-service.js
@@ -53,10 +53,10 @@ export default class AgendaService extends Service {
         item.id
       );
       if (!itemFromStore) {
-        itemFromStore = await this.store.queryRecord(
-          singularize(item.type),
-          item.id
-        );
+        const modelName = singularize(item.type);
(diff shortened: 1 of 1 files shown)


erikap:
@DavithkbY I would expect `item.id` to be always a string. Can you reproduce cases where it isn't?  If not, I assume we can always use `findRecord` instead of `queryRecord`.

DavithkbY:
> @DavithkbY I would expect `item.id` to be always a string. Can you reproduce cases where it isn't? If not, I assume we can always use `findRecord` instead of `queryRecord`.

I wasn't able to reproduce cases where non objects were being passed on. But I thought because the code is 5 years old, queryRecord must've been used for some reason.

That's why I added a check for both, if it's a string we'll use findRecord, if not queryRecord. Just to be on the safe side.
I don't think the check can hurt. But if you're positive only strings will be passed, I can remove the check for string and just always use findRecord.

erikap:
If you cannot reproduce the case where non-objects are passed, I don't see how this PR will solve the issue of the 'new' badge? Am I missing something? :thinking: 

> > @DavithkbY I would expect `item.id` to be always a string. Can you reproduce cases where it isn't? If not, I assume we can always use `findRecord` instead of `queryRecord`.
> 
> I wasn't able to reproduce cases where non objects were being passed on. But I thought because the code is 5 years old, queryRecord must've been used for some reason.
> 
> That's why I added a check for both, if it's a string we'll use findRecord, if not queryRecord. Just to be on the safe side. I don't think the check can hurt. But if you're positive only strings will be passed, I can remove the check for string and just always use findRecord.

DavithkbY:
> If you cannot reproduce the case where non-objects are passed, I don't see how this PR will solve the issue of the 'new' badge? Am I missing something? 🤔
> 
> > > @DavithkbY I would expect `item.id` to be always a string. Can you reproduce cases where it isn't? If not, I assume we can always use `findRecord` instead of `queryRecord`.
> > 
> > 
> > I wasn't able to reproduce cases where non objects were being passed on. But I thought because the code is 5 years old, queryRecord must've been used for some reason.
> > That's why I added a check for both, if it's a string we'll use findRecord, if not queryRecord. Just to be on the safe side. I don't think the check can hurt. But if you're positive only strings will be passed, I can remove the check for string and just always use findRecord.

Oh no, I meant it the other way around. I couldn't reproduce the case where objects are passed. It would always just pass strings. And that's why the queryRecord would fail.

I too would think it would be logical that item.id is a string and not an object. So no Idea why queryRecord would have been used.

erikap:
Makes more sense :sweat_smile: 

Looking at the broader context of the `agenda-service` each of the `this.store.findRecord/queryRecord` is preceded by `this.store.peekRecord`

```
      let itemFromStore = this.store.peekRecord(
        singularize(item.type),
        item.id
      );
 ```

Therefore I think it's safe to assume `item.id` will always be a string, since that's also what `peekRecord()` expects.

It seems the intention of the code was to verify first if the item has already  been loaded in the Ember Data store using `peekRecord`. If not, it's loaded from the backend. I assume someone has made a mistake and wanted to use `queryRecord` to force the request to the backend, but failed to realise that this function expects different arguments to `peekRecord`/`findRecord`. Just using `findRecord` will do the job imo :+1:

DavithkbY:
> Makes more sense 😅
> 
> Looking at the broader context of the `agenda-service` each of the `this.store.findRecord/queryRecord` is preceded by `this.store.peekRecord`
> 
> ```
>       let itemFromStore = this.store.peekRecord(
>         singularize(item.type),
>         item.id
>       );
> ```
> 
> Therefore I think it's safe to assume `item.id` will always be a string, since that's also what `peekRecord()` expects.
> 
> It seems the intention of the code was to verify first if the item has already been loaded in the Ember Data store using `peekRecord`. If not, it's loaded from the backend. I assume someone has made a mistake and wanted to use `queryRecord` to force the request to the backend, but failed to realise that this function expects different arguments to `peekRecord`/`findRecord`. Just using `findRecord` will do the job imo 👍

I removed the extra check.

erikap (approved):

PR #2420, opened 2026-07-09
Prevent signflow from being started when decisions isnt released, prevent new document versions when sign flow is already started

DavithkbY:
Stop displaying the signing workflows for a postponed agenda item that has been re-agendized.
We do this by looking at the decision inside the treatment's internal-decision-publication-activity.
The treatment always has the decisions of the latest agenda item.

Also prevent new document versions from being able to be created when a signflow already has started.
Also accepting new updates isn't allowed when a signflow already has started.

### DIFF
--- a/app/components/documents/add-draft-document-card.js
+++ b/app/components/documents/add-draft-document-card.js
@@ -29,9 +29,12 @@ export default class DocumentsAddDraftDocumentCardComponent extends Component {
   @service intl;
   @service pieceAccessLevelService;
   @service draftSubmissionService;
+  @service signatureService;
 
   @tracked piece;
   @tracked documentContainer;
+
+  @tracked hasStartedSignFlow = false;
 
   @tracked isOpenUploadModal = false;
   @tracked isOpenVerifyDeleteModal = false;
@@ -113,6 +116,11 @@ export default class DocumentsAddDraftDocumentCardComponent extends Component {
         `You should provide @piece or @documentContainer as an argument to ${this.constructor.modelName}`
       );
     }
+    if (this.piece?.constructor.modelName === 'piece') {
+      this.hasStartedSignFlow = await this.signatureService.hasStartedSignFlow(
+        this.piece
+      );
+    }
   });
 
   loadFiles = task(async () => {
@@ -142,7 +150,9 @@ export default class DocumentsAddDraftDocumentCardComponent extends Component {
   });
 
   get mayShowAddNewVersion() {
-    return this.piece.constructor.modelName === 'piece';
+    return (
+      this.piece.constructor.modelName === 'piece' && !this.hasStartedSignFlow
+    );
   }
 
   get mayShowDeleteDraftPiece() {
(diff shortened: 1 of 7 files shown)


app/routes/signatures/index.js
```diff
@@ -61,19 +61,6 @@ export default class SignaturesIndexRoute extends Route {
     const filter = {
       ':has-no:meeting': true,
       'sign-subcase': {
```
erikap, line 63: When testing on production-like data it seems we still need the following filters to prevent "old" sign flows to appear in the list (and make the route error in some cases):
```
      'sign-subcase': {
        'sign-marking-activity': {
          piece: {
            ':has:agendaitems': 'yes',
            ':has-no:next-piece': 'yes'
          },
        },
        ':has-no:sign-preparation-activity': 'yes',
      }
```

Unwanted sign flows that appear in the list without these filters:
- sign-flows started on a document for which a new version has been uploaded afterwards
- sign-flows started on a document that is no longer attached to an agendaitem
- sign-flows with a sign-marking-activity that no longer have a document attached


Other option is to cleanup the data in the triplestore. But even then I would add the filtering because we're not sure these scenarios won't happen in the future. What do you think @tdn ?

app/components/submission/header.js
```diff
@@ -39,11 +39,35 @@ export default class SubmissionHeaderComponent extends Component {
...
+  }
+
+  loadStartedSignFlows = task(async () => {
+    if (!this.isUpdate) {
+      return;
+    }
+    const draftPieces = await this.args.submission?.pieces;
+    const found = [];
+    for (const draftPiece of draftPieces?.slice() ?? []) {
+      const previousPiece = await draftPiece.previousPiece;
```
erikap, line 58: I think to be correct we should walk the entire chain of previous document versions and check if none of them has a sign-flow.
I hoped it would be possible to express this in one query, but I'm afraid it's not.

app/components/documents/add-draft-document-card.js
```diff
@@ -142,7 +150,9 @@ export default class DocumentsAddDraftDocumentCardComponent extends Component {
   });
 
   get mayShowAddNewVersion() {
-    return this.piece.constructor.modelName === 'piece';
+    return (
+      this.piece.constructor.modelName === 'piece' && !this.hasStartedSignFlow
```
erikap: Maybe it's more clear to the user to run the check when they click the button and yield a warning then?

erikap (commented):
Not sure what the best pattern is in case there is a sign flow on a piece:
- not showing the 'Upload new version' button (but user may wonder why it's not visible)
- showing the 'Upload new version' button and yield a warning once the user clicks it (but user may wonder why we show it)

The second option has the advantage the check is executed closer to the moment the user actually uploads a new version.

Should the check for sign-flows should also be added on the regular (non-draft) documents?

app/routes/signatures/index.js
erikap: We will do both.

DavithkbY: Implemented the filter

DavithkbY:
@erikap I removed the code changes for the draft documents

PR #2423, opened 2026-07-23
Add resource attribute to submission history, add checkbox for admins to see concept submissions of all organizations

DavithkbY:
Add resource attribute to submission history, add checkbox for admins to see concept submissions of all organizations

### DIFF
--- a/app/components/cases/submissions/history-panel.hbs
+++ b/app/components/cases/submissions/history-panel.hbs
@@ -17,7 +17,11 @@
       {{else}}
         {{#if @multiple}}
           {{#each @statusChangeActivities as |statusChangeActivityArray|}}
-            <div class={{if (has-next statusChangeActivityArray @statusChangeActivities) "auk-u-mb-4"}}>
+            {{! resource attribute for debugging with SPARQL }}
+            <div
+              resource={{get (object-at 0 statusChangeActivityArray) "submission.uri"}}
+              class={{if (has-next statusChangeActivityArray @statusChangeActivities) "auk-u-mb-4"}}
+            >
               {{#let (object-at (add statusChangeActivityArray.length -1) statusChangeActivityArray) as |creationActivity|}}
                 <Submission::StatusChangeActivity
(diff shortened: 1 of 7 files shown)


app/components/cases/submissions/history-panel.hbs
```diff
@@ -17,7 +17,11 @@
       {{else}}
         {{#if @multiple}}
           {{#each @statusChangeActivities as |statusChangeActivityArray|}}
-            <div class={{if (has-next statusChangeActivityArray @statusChangeActivities) "auk-u-mb-4"}}>
+            {{! resource attribute for debugging with SPARQL }}
+            <div
+              resource={{get (object-at 0 statusChangeActivityArray) "submission.uri"}}
```
erikap: I don't see the annotation. Maybe because of the async relation? You can mabye move it down and set in in the `Submission::StatusChangeActivity` component?

translations/nl-be.json
```diff
@@ -358,6 +358,7 @@
   "show-archived": "Toon verwijderde dossiers",
   "show-archived-only": "Toon enkel verwijderde dossiers",
   "show-as-remark": "Dit is een mededeling",
+  "show-other-organizations": "Toon andere organisaties",
```
erikap: Better to show 'Toon alle organisaties' (and `show-all-organizations`), since your own organization will also be included

erikap (changes requested):
Good solution to add the checkbox on the submission concepts route. I would just change the label to 'Toon alle organisaties' (or maybe even 'Toon alle kabinetten').

`resource` RDFa annotation doesn't seem to work.

app/components/cases/submissions/history-panel.hbs
DavithkbY: Weird, for me it works:
image

translations/nl-be.json
DavithkbY: Fixed

PR #2425, opened 2026-08-06
Sort by creation date instead of name for document versions

DavithkbY:
Sort by creation date instead of name for document versions. I also added a fallback when creation date is not set, to still filter by name.

### DIFF
--- a/app/utils/documents.js
+++ b/app/utils/documents.js
@@ -70,9 +70,18 @@ export const sortDocumentContainers = async (
 };
 
 export const sortPieceVersions = (pieces) => {
-  // This function wraps sortPieces just so that the intent of calling it
-  // is more obvious, i.e. when calling this function you're passing in all
-  // the pieces that belong to a single document container.
+  // Sort by created date, if it's missing fallback to sorting by name
+  const createdAvailable = pieces.every((piece) => piece.created);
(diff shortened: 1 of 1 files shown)


app/utils/documents.js
```diff
@@ -70,9 +70,18 @@ export const sortDocumentContainers = async (
...
+  // Sort by created date, if it's missing fallback to sorting by name
+  const createdAvailable = pieces.every((piece) => piece.created);
+  if (createdAvailable) {
+    return pieces.slice().sort(
+      (pieceA, pieceB) =>
+        pieceB.created - pieceA.created ||
+        compareFunction(
+          new VRDocumentName(pieceA.name),
+          new VRDocumentName(pieceB.name)
+        )
```
erikap: Why this fallback if you checked whether every piece has a `created` attribute?

DavithkbY: I changed it to sort by previous/nextpiece attribute. I asked Tom if it was possible that there would be old records that didn't have a creation date. That's why I added the fallback.

I've still added the fallback now with the new sorting, also just in case we don't have a clean chain.

erikap:
`unit/access-level.cy.js` test fails in the test suite. Can you have a look @DavithkbY ?

